Issue #6, opened 2017-02-02
Audio support for CEL/CELX scripts


CelestiaProject:
**Celestia-ED-1.6.1** contains the scripts with audio playback. Used **OpenAL** library. 
I believe that to add audio playback for scripts in Celestia 1.7.0 is a good idea. But OpenAL library is outdated and need to use another.

Alexell:
Maybe **FMOD Ex** library?

Alexell:
FMOD Ex will be added.

Alexell:
Original patch created by Victor (vhpgomes), modified by Vincent,
modified and adapted for **FMOD library** by Alexell.
Done:  [37ed932](https://github.com/CelestiaProject/Celestia/commit/37ed932d37efe5c4d4d6650cf8ed37ad6a251122)

CelestiaProject:
Need additional testing and code analysis, because celestia.exe sometimes crash, and source code compilation is not always successful.

Alexell:
Added lost classes in command.cpp
CELX method work correctly.
CEL method error (need to fix).

CelestiaProject:
Need additional testing, because celestia.exe and celestia qt.exe sometimes crashes with **fmod.dll exception**

CelestiaProject:
Need help to fix FMOD support to work properly on Win32, Win64, QT, Linux, MacOS.

CelestiaProject:
@dbrant, we need help to fix FMOD support to work properly on Win32, Win64, QT, Linux, MacOS.

dbrant:
I have a couple of concerns about the choice to use FMOD instead of OpenAL, specifically when building for Linux:
- I'm not familiar with FMOD, but it looks like it's a commercial (closed-source) library. Because of this, it's not available in any package manager of any Linux distribution. Therefore, if a volunteer (like me) wants to start contributing, they would need to manually go to the fmod website, download the libraries (which requires creating an account!), and manually copy the libraries and headers to /usr/local/. That's quite a few steps, and easily enough to turn someone away from contributing.
- More importantly, since FMOD is closed-source, will it prevent Celestia from being included in the package managers of Linux distros (i.e. allow users to install it with `apt-get install celestia`)? Again, I'm not familiar with the licensing and distribution agreements of FMOD, but I doubt that Ubuntu or Fedora would agree to distribute Celestia if it starts linking to a closed-source library.

CelestiaProject:
[...]

Issue #1, opened 2017-02-02
Eigen library update


CelestiaProject:
In the future, it would be nice to update the Eigen library because there was a lot of improvements in the mathematical part of it.

CelestiaProject:
Done in [27223e3](https://github.com/CelestiaProject/Celestia/commit/27223e31ef8d1ac524c451b3c676319b71a49434) by @dbrant

PR #12, opened 2017-04-08
Upgrade project to VS 2017, and update for 64-bit support.

dbrant:
This set of commits allows us to build the project using Visual Studio 2017, as well as updates the code for 64-bit Windows (and adds an x64 build target). Other notes:

- Upgrade to Eigen version 3.3, and update the code accordingly.
- Upgrade to libpng 1.6.29.


dbrant:
This would take care of #1 as well.

CelestiaProject:
Hey. Thanks for the work you've done. It is really big and important for us. Your Celestia version compiles for Win32 and Win64 correctly? Celestia works without errors?

CelestiaProject:
Also interested in QT version, since it will be the basis for Celestia 1.7.0

Alexell:
FMOD was taken from [here](http://www.fmod.org/download) (**FMOD Studio for Unreal Engine 4 (4.15)** section. I think that I did not use it correctly, because Celestia sometimes crashes with fmod.dll exception. (described in #6)

If you can help to properly use FMOD, that it works correctly on x86 and x64, and also supported on Mac and Linux (Celestia is cross-platform) - I will be very grateful to you.

dbrant:
To answer your questions:
- Yes, with these patches, Celestia compiles successfully for Win32 and 64, and runs without errors. The only issue seems to be with FMOD, as you've described.
- I haven't tried building it with Qt, but I can't imagine that these patches break the existing Qt frontend, since none of my changes really touch the Qt side of things. I'll give it a try later.  I'll also verify that it still builds and runs correctly in Linux and Mac.

I did also have a question of my own:
- How do you generate the installer package (for Windows)? I'm not too familiar with making installers; I've used NSIS a long time ago, but it doesn't seem like that's being used here...

Alexell:
> Yes, with these patches, Celestia compiles successfully for Win32 and 64, and runs without errors. The only issue seems to be with FMOD, as you've described.
If you can properly connect FMOD, so that it works on win32, win64, qt, mac and linux - we will be very grateful, like many users who will then use the program.

> ... none of my changes really touch the Qt side of things ...
If we update the project, I would like to ask you to update the QT version. Now I use QT4 (I compile with QT 4.8.0 and QT Creator 2.5.2). I would like to use the new QT and QT Creator, respectively.

> How do you generate the installer package (for Windows)?
Previously Celestia installer was collected through Inno Setup (file celestia.iss). Now before the release we are still far away, so I did not think what program we will use for this.

PR #24, opened 2017-11-11
Build on linux

rikardfalkeborn:
This is what I needed to change to build on Linux.


rikardfalkeborn:
With the latest restructuring I can see why this was closed (a comment would have been nice though). Are you interested in making configure work on Linux on the main branch (the last commit + one more change makes it work for me) or is it the QT version that is the supported one now (which builds fine on Linux, module a path change for cspice in celestia.pro, but that might depend on how the system is set up  I guess)?

Issue #4, opened 2017-02-02
QT version localization


CelestiaProject:
Dialogs in the experimental **QT version** of Celestia available in English only. In the future, we need to organize a localization.

CelestiaProject:
#33

CelestiaProject:
#33

Alexell:
Everything is working.

PR #33, opened 2018-03-09
Add QTranslator support and generate initial .ts files

iglosiggio:
Hi! I'm working to add Celestia to a debian-based distro and while packaging the software i saw the oportunity to give it some love on the translations.

On our package we generate the .qm files and install them manually, but i think that this can be incorporated into celestia.pro easily.

The initial .ts files were generated with the command `lupdate celestia.pro`, and i've translated the spanish UI.

It's worth noting that this only affects the Qt dialogs that are created via resource files (.ui) and the Qt translatable string (the `tr` mactro). I've also updated in our package de .pot files but those changes are still messy.


375gnu:
i'd rather add new strings to existing *.po & *.pot files instead of introducing new files just for interface

iglosiggio:
I did thought that initially but quickly realized that [QT cannot use .mo](https://bugreports.qt.io/browse/QTBUG-2404) files for the translation of it's ui. I think that the [alternative](https://askubuntu.com/a/143153) isn't much better.

I'm open for suggestions on how to make this better anyway :smile: using __both__ gettext and QTranslator is an ugly hack (that i personally found better that the proposed alternatives)

375gnu:
`*.po` files can be converted to `*.qm` files using `lconvert`. But for for me it's better to use gettext as for the rest of the application.

iglosiggio:
I did already knew that, but given the way that the beta QT frontend builds it's dialogs i really don't know how can gettext be used to translate the resource files.

I originally used extractrc+gettext but it seems that the only way for those translations to take effect was hacking into QTranslator.

If you could guide me with some documentation on this i will try to make an alternative patch.

Thanks for the comments!

375gnu:
This [manual](https://doc.qt.io/qt-5/uic.html) contains some interesting things:

> -tr <func> Use <func> for translating strings instead of tr().
> --include <file> Add #include <file> to the output.

But I don't know how to pass uic options from a project file.

iglosiggio:
I will investigate on that, thanks!

375gnu:
I've managed to enable gettext usage with qt5 but with workarounds:
to celestia.pro add the following lines:
```
load(uic)
uic.commands +=  "--tr qt5gettext --include qt5gettext.h"
```
create qt5gettext.h with
```
#include <libintl.h>
#define qt5gettext(s,i) gettext(s)
```

We need this ugly workaround because uic generates function calls with 2 arguments:
```
label_3->setText(qt5gettext("Time source:", Q_NULLPTR));
```

iglosiggio:
That's genius! I was stuck trying to get [QMAKE_UIC_FLAGS](http://doc.qt.io/qt-5/qmake-variable-reference.html#qmake-uic-flags) working.

375gnu:
I was struggling against this variable as well until google said me that it's removed in 5.x.

CelestiaProject:
All QT dialog boxes can be automatically translated by the content of .mo files.
This can be seen right away, as soon as you start Celestia QT in a language other than English - some of the dialog boxes have already been translated.

CelestiaProject:
#4

Alexell:
I checked the localization work without using the code given in this request.
QT dialogs are not really translated by .po / .mo files.

Alexell:
@iglosiggio @375gnu Localization with compiled .qm files works.
But how do I make the compiled .qm files in the **locale** folder and not in the Celestia root folder?

iglosiggio:
http://doc.qt.io/qt-5/qtranslator.html#load

I think that you can do `celestiaTranslator.load("locale/celestia_" + functionThatReturnsSystemLocale());`

And in the install of qmake install the .qm files on that folder.

PD: I think that @375gnu solution for a unified translation system was a lot more elegant than my solution of using both gettext and QTranslator.

Alexell:
@iglosiggio Thanks. It work.
P.S. Yes you are right. But it takes more time to implement it. In any case, I do not see anything wrong: in Celstia 1.6.1 there were .mo files with translation gettext and dll files with translations of dialog boxes. In Celestia 1.7.0 there will be .mo files and .qm files instead of dll files.

PR #38, opened 2018-03-12
fix whitespaces and indents

375gnu:
trailing spaces are removed
tabs are replaced with 4 spaces


CelestiaProject:
Good job. But now there is a merger conflict, so the merger will be postponed. We need to study the subject of the conflict.

375gnu:
i've rebased against current main branch and rechecked everything manually.

Issue #34, opened 2018-03-12
panning is broken


375gnu:
When one starts panning (horizontal or vertical movement of a camera) the camera "jumps" (usually up and right) and then continues to move properly.

This issue exist only with a Qt version, not with a glut one.

CelestiaProject:
Do you have any ideas how to solve the problem?

gstark31897:
Hey, do you guys mind if I work on this? I made a pull request that should address this.

CelestiaProject:
@gstark31897 Of course, do not mind. The main thing that everything worked!

PR #41, opened 2018-04-24
Fix panning issue #34

gstark31897:
Hey everyone, I made a patch that should address issue #34 
I reworked a majority of the qt mouse movement code to reduce the complexity.
I removed lastX and lastY from the code as well since they weren't needed to track mouse position anymore.
I'm happy to make this change look more like the original file if you want it to reflect some of the gtk code (if applicable, I didn't look for the gtk code).

Let me know what you think.


CelestiaProject:
GTK is no longer needed. We must go completely to QT.
Thank you very much for the work done.

gstark31897:
Happy to help :)

375gnu:
Thank you @gstark31897! This release fixed the issue.

PR #42, opened 2018-05-02
Replace charEntered call in qt CelestiaAppWindow with proper CelestiaCore API

pirogronian:
Hi All
I decidet to rewrite internal communitation between CelestiaCore and UI (in this case, Qt). My presumption is any input event handling should by done by UI, not by CelestiaCore. So, in this PR I replaced any charEntered call in qtappwin with proper CelestiaCore function, creating few new functions in CelestiaCore as needed, as well as one or two new actions in qtappwin.
I checked every change, however this is my first ever PR and I'm not very familiar with git yet. If something is messed, sorry.


375gnu:
While the direction taken is to have only QT5 based backend it would be good to leave charEntered() code working for GLUT users. At least on some video drivers and hardware CelestiaQT causes lockups.

pirogronian:
Will this code interferring with Qt actions shortcuts? If so, what about moving this code to separate component, that could be enabled on demand?

375gnu:
Need to check. Maybe it worth to use `#ifdef ... #endif` instead of comments.

pirogronian:
Done as You suggested, but time travel delay activation seems to be always ignored. I'm not sure if it's a regression or I just don't know what it really is... Eventually I will revert changes done for that action.

pirogronian:
As QActions truly grab keyboard shortcuts, there is no need to additionally prevent charEntered implementation processing. However, this behaviour brake name searching (TextEnterMode). So I'm working on it.

pirogronian:
Now I want to switch to different branch and other changes bring in separate pull requests. I'm not sure, what are quality standards here and if my PR will be merged eventually. However, until my motivation wont vapourise, I will continue to commit to my fork. Maybe it will prove to be worth it.
I see some problems with ql widget class, as it pass only capital letters for some keys. Not sure, why it's written so.

celestia.pro
```diff
@@ -468,6 +468,10 @@ INCLUDEPATH += thirdparty/glew/include
 INCLUDEPATH += thirdparty/Eigen
 INCLUDEPATH += thirdparty/curveplot/include
 
+PRECOMPILED_HEADER =
```
375gnu: Please leave these changes in your local branch.

From here:

celestia.pro
```diff
@@ -609,8 +613,8 @@ unix {
     isEmpty (LUAPC) {error("No shared Lua library found!")}
 
     PKGCONFIG += glu $$LUAPC libpng theora
-    INCLUDEPATH += /usr/local/cspice/include
-    LIBS += -ljpeg /usr/local/cspice/lib/cspice.a
+#    INCLUDEPATH += /usr/include
+    LIBS += -ljpeg /usr/lib/cspice.a
```
375gnu: ... to there.

src/celestia/CelestiaCoreApplication.h
```diff
@@ -0,0 +1,159 @@
+#ifndef __CELESTIA_CORE_APPLICATION__H__ 
+#define __CELESTIA_CORE_APPLICATION__H__
+
+//#include <cstring>
```
375gnu: Please remove this useless  comment

src/celestia/celestiacore.cpp
```diff
@@ -1485,11 +1488,11 @@ void CelestiaCore::charEntered(const char *c_p, int modifiers)
         else
             flash(_("Alt-azimuth mode disabled"));
         break;
-
+//#ifndef QT_GUI_LIB
```
375gnu: It would be better to remove these commented guards.

src/celestia/celestiacore.h
```diff
@@ -294,7 +294,7 @@ class CelestiaCore // : public Watchable<CelestiaCore>
     void resize(GLsizei w, GLsizei h);
     void draw();
     void tick();
-
+
```
375gnu: Please remove extra spaces.

src/celestia/qt/qtappwin.cpp
```diff
@@ -1406,6 +1449,19 @@ void CelestiaAppWindow::populateBookmarkMenu()
     m_bookmarkManager->populateBookmarkMenu(bookmarkMenu);
 }
 
+/*
```
375gnu: Please remove this commented code.

src/celestia/CelestiaCoreApplication.h
```diff
@@ -0,0 +1,159 @@
...
+    void toggleAltAzimuthMode() {
+        addToHistory();
+        setAltAzimuthMode(!getAltAzimuthMode());
+        if (altAzimuthMode)
+        {
+            aflash("Alt-azimuth mode enabled");
+        }
+        else
+            aflash("Alt-azimuth mode disabled");
+    }
```
375gnu: Closing braked is missing

src/celestia/CelestiaCoreApplication.h
```diff
@@ -0,0 +1,159 @@
...
+    template <int F> class CelestiaRenderFlag {
+        int flag;
+        CelestiaCoreApplication *core;
+        const char *onMsg;
+        const char *offMsg;
+    
+    public:
+        CelestiaRenderFlag(
+            CelestiaCoreApplication *c,
+            const char *on = 0,
```
375gnu: Please use NULL instead of 0.

src/celestia/CelestiaCoreApplication.h
```diff
@@ -0,0 +1,159 @@
...
+    void gotoSelection(double gotoTime,
+                       const Eigen::Vector3f& up,
+                       ObserverFrame::CoordinateSystem upFrame);
+    
+    void gotoSelection(double gotoTime) {
+        gotoSelection(gotoTime, Eigen::Vector3f::UnitY(), ObserverFrame::ObserverLocal);
+    }
+    
+    void gotoSurface(double time = 5.0) {
+        aflash("Goto surface");
```
375gnu: Mark string for gettext (`N_()` afair)

src/celestia/CelestiaCoreApplication.h
```diff
@@ -0,0 +1,159 @@
...
+class CelestiaCoreApplication : public CelestiaCore {
+    bool autoMessages;
+    
+public:
+    
+    CelestiaCoreApplication();
+    
+    void setAutoMessagesOn() { autoMessages = true; }
+    void setAutoMessagesOff() { autoMessages = false; }
+    bool getAutoMessages();
```
375gnu: `getAutoMessages()` is left unimplemented.

src/celestia/CelestiaCoreApplication.cpp
```diff
@@ -0,0 +1,78 @@
...
+    if (!showActiveViewFrame)  flashFrameStart = currentTime;
+}
+
+bool CelestiaCoreApplication::setLightTravelDelay(bool on) {
+    if (sim->getSelection().body() &&
+        (sim->getTargetSpeed() < 0.99 * astro::speedOfLight))
+    {
+        Vector3d v = sim->getSelection().getPosition(sim->getTime()).offsetFromKm(sim->getObserver().getPosition());
+        lightTravelFlag = !lightTravelFlag;
+        if (on)
```
375gnu: I'd rather rewrite this using ternary operator:
`setLightTravelDelay(on ? v.norm() : -v.norm());

src/celestia/CelestiaCoreApplication.cpp
```diff
@@ -0,0 +1,78 @@
...
+            setLightTravelDelay(v.norm());
+        }
+        else
+        {
+            setLightTravelDelay(-v.norm());
+            
+        }
+        return true;
+    }
+    else
```
375gnu: `else` is useless here

src/celestia/CelestiaCoreApplication.h
```diff
@@ -0,0 +1,159 @@
...
+            core->notifyWatchers(RenderFlagsChanged);
+        }
+    
+        void setOff() {
+            core->getRenderer()->setRenderFlags(core->getRenderer()->getRenderFlags() & ~F);
+            core->notifyWatchers(RenderFlagsChanged);
+        }
+    
+        void set(bool on) {
+            if (on) setOn();
```
375gnu:
```if (on)
       setOm();
   else
       setOff();```

File: src/celestia/CelestiaCoreApplication.h [hunk 14]
```diff
@@ -0,0 +1,159 @@
...
+    void setAutoMessagesOff() { autoMessages = false; }
+    bool getAutoMessages();
+    void aflash(const char *n) {
+        if (getAutoMessages())  flash(_(n));
+    }
+    
+    template <int F> class CelestiaRenderFlag {
+        int flag;
+        CelestiaCoreApplication *core;
+        const char *onMsg;
```
375gnu: It seems that you don't use `onMsg` & `offMsg` actually. `flashOn` & `flashOff`are never used.

src/celestia/CelestiaCoreApplication.h
pirogronian: aflash() implementation do it already, however I can change it, if it's a bad idea.

pirogronian: All strings I used are from legacy code, however I know little about used here translation method. I will take a look and check, how it works and what is needed to add. Thanks for help.

375gnu:
Hi @pirogronian, if you has finished with this PR could you please squash your commits? I suggest to have 2 commits: 1st one which introduces CelestiaCoreApplication and 2nd one which actually uses it.

pirogronian:
Of course, whenever I will have learned, how to squash commits. In case I will fail to do it fast enough, there is simple rule: CelestiaCoreApplication.* files introduces new subclass, changes in qtappwin.* and qtcelestiaactions.* contains usage of this class.

375gnu:
This is very trivial: `git rebase -i HEAD~9', where 9 is a number of your commits in this branch.

pirogronian:
Done without errors, but I see that nothing happened (<code>git push</code> said 

> Everything up-to-date

). I will read a bit more.

375gnu:
hmm, if you did correctly you would see an error that non fast-forward pushes are can't be done.

pirogronian:
At last I do something comparable to squashing, however without <code>git rebase</code>, but <code>git reset HEAD~9</code> (which doesn't modify working directory) and adding files in two commits. Not sure, if after apply first one project will compile (I saw <code>private:</code> changed to <code>protected:</code> in <code>celestiacore.h</code>, changed with second commit). However I'm now much more familiar with git itself...

CelestiaProject:
What is the stage of this pull request? Have you checked your changes in compiled Celestia?

pirogronian:
Of course. I didn't commit until checked  by recompile and run Celestia, trying features I worked on.

Issue #3, opened 2017-02-02
Need to fix obs:gotolonglat error


CelestiaProject:
In Celestia 1.7.0 source code (even in the 5229 revision from SF), the error occurred: In CELX scripts where used obs:gotolonglat command, not processed **distance** parameter. The reason for this is unknown to me, but I think that error associated with porting the source code on a new mathematical engine **Eigen**.

Alexell:
**Write in celx:** `obs:gotolonglat(Earth, math.rad(0), math.rad(30), 3.5004702E14, 5)`
Celestia must go to Earth at a distance 3.5004702E14, but Celestia go to Earth center!
**File:** celx_observer.cpp
**String 363:** distance = distance / KM_PER_LY;
If comment this string, the Celestia go to Earth the correct distance.

Alexell:
[c81bd5f](https://github.com/CelestiaProject/Celestia/commit/c81bd5fed991c810a30d35be3fa6adfdf39c0730)

PR #47, opened 2018-05-08
Lua 5.2 and 5.3 fixes

375gnu:
Currently on unix platforms we can build Celestia only with Lua 5.2 support, but its support isn't complete.
This PR tries to fix issues we have:
1. It restores usage of older Lua versions (5.1 and 5.0).
1. Adds fixes to enable script usage (loads coroutines library used internally).
1. Adds a new method `gotoobject` behaving exactly as `goto` method. When compiled against Lua < 5.2 both methods exist, with Lua >= 5.2 only `gotoobject`. This change is required because Lua 5.2 introduced `goto` statement so we can't use `goto` as function name anymore.

Windows and OSX build are not affected because they have Lua 5.1 hardcoded.

I don't insist on `gotoobject` name we can find a better one.

My suggestion is support Lua 5.1 as a main Lua version and announce experimental support support for Lua 5.2 and 5.3 so extension authors have time to fix their scripts.

### DIFF
--- a/scripts/tour-system.celx
+++ b/scripts/tour-system.celx
@@ -1,7 +1,7 @@
-function goto(o, t)
+function gotoobject(o, t)
     local obs = celestia:getobserver()
     obs:follow(o)
-    obs:goto(o, t)
+    obs:gotoobject(o, t)
         while (obs:travelling()) do
         wait(0)
     end
@@ -11,7 +11,7 @@ function visit(o)
     local i, v
     celestia:select(o)
     celestia:flash(o:type() .. " - " .. o:name())
-    goto(o, 3)
+    gotoobject(o, 3)
     wait(0.5)
     local children = o:getchildren()
     for i, v in ipairs(children) do
--- a/scripts/tour-system.celx
+++ b/scripts/tour-system.celx
@@ -1,3 +1,5 @@
+-- Title: Tour around Solar system objects
+
 function gotoobject(o, t)
     local obs = celestia:getobserver()
     obs:follow(o)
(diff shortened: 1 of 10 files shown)


scripts/tour-system.celx
```diff
@@ -1,7 +1,9 @@
-function goto(o, t)
+-- Title: Tour around Solar system objects
+
+function gotoobject(o, t)
```
375gnu: goto is reserver keyword in Lua >= 5.2

scripts/tour-system.celx
```diff
@@ -1,7 +1,9 @@
...
     local obs = celestia:getobserver()
     obs:follow(o)
-    obs:goto(o, t)
+    obs:gotoobject(o, t)
```
375gnu: goto is reserver keyword in Lua >= 5.2

scripts/z-dist.celx
```diff
@@ -20,7 +20,7 @@ end
 
 function rgb2hex(r,g,b)
    -- Converts color code from RGB to Hex
-   local hex = string.format("#%.2x%.2x%.2x", r,g,b)
+   local hex = string.format("#%.2x%.2x%.2x", math.floor(r), math.floor(g), math.floor(b))
```
375gnu: Lua 5.3 is stricter in number conversions.

`Error: [string "scripts/z-dist.celx"]:23: bad argument #3 to 'format' (number has no integer representation)`

src/celestia/celx.cpp
```diff
@@ -515,7 +515,11 @@ LuaState::LuaState() :
     ioMode(NoIO),
     eventHandlerEnabled(false)
 {
+#if LUA_VER >= 0x050100
     state = luaL_newstate();
+#else
+    state = lua_open();
```
375gnu: Lua 5.0 doesn't have `luaL_newstate()`

CelestiaProject:
Excellent work. Thank you!

PR #49, opened 2018-05-20
Play sound

pirogronian:
Added frontend-backend interface for simple playback. Added basic Qt backend. Reverted list of lua packages in .pro to fix too low lua version detection (which has blown up any compilation try). Confused by github due to strange branch history (too many commits as it were no squashed). Probably I should rebase intead of pull... Hoping project maintainers will help with it.
But, of course, everything is tested. Almost. scripts/audio/sound_demo.cel is failing with empty error mesage. No real chance to imagine, what is wrong. Probably will fixed in separate pull request.

### DIFF
--- a/celestia.pro
+++ b/celestia.pro
@@ -8,6 +8,7 @@ RCC_DIR = rcc
 
 QT += opengl
 QT += xml
+QT += multimedia
 
 unix {
     !exists(config.h):system(touch config.h)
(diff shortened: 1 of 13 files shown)


pirogronian:
Sorry for messing with history of this pull request, but I found one forgotten, uncommitted file and, as usually, started messing with git, to avoid several commits in this PR. Should be fine now.

PR #57, opened 2018-06-06
unix: install splash.png into a proper location

375gnu:


Alexell:
What is this for? Splash - one picture that does not contain any inscriptions, except the name of the program. It does not need to be moved.

But very necessary to implement **display of loading files** in QT splash as is done in the windows splash (not QT).

375gnu:
Current code loads `splash.png` from the current directory. So in my case it doesn't load it at all. Celestia should load it from its data directory, in case of GNU/Linux this is `$prefix/share/celestia`, in case of Windows `c:/Program Files/Celestia/` etc.

So my PR just installs it there and tells Celestia to load it from the proper location (but only on Unix-like systems, on other OSes celesta.pro should changed appropriately.)

375gnu:
> At you for Linux systems it turns out splash will be loaded not from share/celestia/splash.png, but from share/celestia/splash/splash.png

Yes, i just haven't checked the path while was writing the previous comment.

> In addition, you in the UNIX section declared SPLASH_DIR and used it in qtmain.cpp. On Windows, this is likely to cause a compilation error (I can not verify right now).

No because of the following:
```#ifndef SPLASH_DIR
#define SPLASH_DIR
#endif```
So on Windows and OS `SPLASH_DIR` is empty and `pixmap(SPLASH_DIR "splash.png");` is effective equal to `pixmap("splash.png");`

Issue #55, opened 2018-05-30
Undefined Q_NAMESPACE


375gnu:
Hi @pirogronian,

I'm compiling on Debian 9 based system, Qt version is 5.7.1.

```
g++ -c -m64 -pipe -O2 -Wall -W -D_REENTRANT -fPIC -DGLEW_STATIC -DEIGEN_NO_DEBUG -DEIGEN_NO_DEBUG -DNDEBUG -DNO_DEBUG -DTHEORA -DLUA_VER=0x050300 -DCELX -DCONFIG_DATA_DIR=\"/opt/celestia/share/Celestia_QT\" -DQT_NO_DEBUG -DQT_OPENGL_LIB -DQT_WIDGETS_LIB -DQT_MULTIMEDIA_LIB -DQT_GUI_LIB -DQT_XML_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -I../../Celestia -I. -I../../Celestia -I../src -I../thirdparty/glew/include -I../thirdparty/Eigen -I../thirdparty/curveplot/include -I/usr/local/cspice/include -isystem /usr/include/libdrm -isystem /usr/include/lua5.3 -isystem /usr/include/libpng16 -isystem /usr/include/x86_64-linux-gnu/qt5 -isystem /usr/include/x86_64-linux-gnu/qt5/QtOpenGL -isystem /usr/include/x86_64-linux-gnu/qt5/QtWidgets -isystem /usr/include/x86_64-linux-gnu/qt5/QtMultimedia -isystem /usr/include/x86_64-linux-gnu/qt5/QtGui -isystem /usr/include/x86_64-linux-gnu/qt5/QtXml -isystem /usr/include/x86_64-linux-gnu/qt5/QtNetwork -isystem /usr/include/x86_64-linux-gnu/qt5/QtCore -Imoc -I. -I/usr/lib/x86_64-linux-gnu/qt5/mkspecs/linux-g++-64 -o obj/qtglwidget.o ../src/celestia/qt/qtglwidget.cpp
In file included from ../src/celestia/qt/qtappwin.h:16:0,
                 from ../src/celestia/qt/qtmain.cpp:29:
../src/celestia/qt/QtCelestiaCoreApplication.h:8:5: error: ‘Q_NAMESPACE’ does not name a type
     Q_NAMESPACE
     ^~~~~~~~~~~
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qglobal.h:83:0,
[...]

PR #58, opened 2018-06-06
Fix Qt compatibility and some other fixes

375gnu:
1.    Fix compatibility with older Qt versions (Closes: #55)
2.    QtAudioManager fixes
    
      1) Better check for Unix-like systems
      2) Use std::cerr to report errors instead of std::cout
      3) Move QMediaPlayer creation to prevent memory leak

### DIFF
--- a/src/celestia/qt/QtAudioManager.cpp
+++ b/src/celestia/qt/QtAudioManager.cpp
@@ -1,4 +1,3 @@
-
 #include <iostream>
 #include <QtCore/QDir>
 #include <QtCore/QUrl>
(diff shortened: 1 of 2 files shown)


src/celestia/qt/QtAudioManager.cpp
```diff
@@ -76,7 +75,9 @@ void QtAudioManager::playAll() {
 }
 
 void QtAudioManager::pauseAll() {
-    cout << "pauseAll()\n";
+#ifndef NO_DEBUG
+    cerr << "pauseAll()\n";
```
pirogronian: It's not an error message

375gnu: that's just a matter of preferences, i usually prefer to have debug messages on stderr, but i'll change if you wish.

375gnu:
fixed

PR #63, opened 2018-06-25
prevent to include unused windows header files [DO NOT MERGE]

GottfriedSp:


Alexell:
Does it have conflicts with QT? The old WIN32 interface will not be used.

375gnu:
Might be good to release a maintenance release 1.6.3 with fixes. But for it need to create a new branch from 1.6.2 code, I hope we can find a proper commit for it.

But there is no reason to merge code for to-be-removed backends to the `main` branch.

And regarding the code quality I'd like to see broader explanation in the commit message (what files won't be included) and this macros should be defined in `config.h` or project files (`Makefile.am`).

GottfriedSp:
This does not conflict with QT.
But I like the native window, as it is fast and the Visual Studio is the best debugger and editor.
I like Visual Studio 2015, Visual Studio 2017 I don't know if I can install it at Windows 7.
For 32-bit Windows I use platform toolset: Visual Studio 2015 - Windows XP (v140_xp) - this works with Windows XP SP3.
The #define WIN32_LEAN_AND_MEAN before including windows.h does not includes hundreds of unused headers with windows.h (direct draw, shellapi, wincrypt, common dialogs, winsock, ...)
Inside Qt the WIN32_LEAN_AND_MEAN is mostly defined, as it is in conflict with winsock2.h when not defined.

Qt5 version 5.6.3 is the last one which work with Windows XP SP3
Qt Creator I don't like, as it is slow and does not work with remote desktop or xrdp, and debugging with it is some special.
I compiled with Qt creator in windows with Qt5.6.3 MSVC2015 64-bit but it crashes at starting the celestia. the same with Qt5.6.3 MSVC2015 32-bit
With Visual Studio 2015 and Qt Visual Studio Tools extension it works with Qt5.6.3

GottfriedSp:
On windows it work both (32-bit and 64-bit)
32-bit windows XP SP3 +
64-bit Vista SP1 +

GottfriedSp:
I have other performance improvements.
I can create patch files if you prefer this, so you can test it before insert into this branch.
For the thirdparty I have included the source files, what compiled as static libs.
Lua 5.1.5

GottfriedSp:
You can discard my pull request!
It doesn't crash, when compiled with visual studio 2008.
And the WIN32_LEAN_AND_MEAN speedup the compile time only for some seconds.

GottfriedSp:
I saw in the coding-standars.html:

> In function calls or definitions, there should be no space between the parentheses and either the arguments or the function name. When there are multiple arguments, a single space should follow each comma separating arguments. 

Maybe this must adapted.
The last entry could be removed, I think nobody uses  Microsoft Visual C 6.0 anymore: 

> The Scope of for

375gnu:
`if` is not a function.

Issue #64, opened 2018-07-04
celx file-io does not work with qt version


GottfriedSp:
the file-io does not work with qt, but it works with native windows (1.7.0)
This example script demonstrate it.
[scripts.zip](https://github.com/CelestiaProject/Celestia/files/2163702/scripts.zip)

First command is for allowing io access: (this ask the first time for permission to file access: y = yes, ESC = cancel, any other key = no)
`celestia:requestsystemaccess()`
I have tried y, Y, and z (german keyboard changed the key 'y' and 'z' from english layout)
The next statement causes a "Fatal Error" [string "scripts/Test.celx"]:11: attempt to index global 'io' (a nil value)
`f = assert(io.open(pathname .. "/data.txt", "r"))`
See link to lua manual 1.5.0 and later [https://en.wikibooks.org/wiki/Celestia/Celx_Scripting/CELX_Lua_Methods](url)

The celestia:getobserver():goto does not go to destination with newer lua interpreter (goto has no label)

375gnu:
What Lua version do you use?

`goto` is a keyword in Lua 5.2 so the method was renamed to `gotoobject`, but it's still available as `goto` on Lua 5.1 and earlier.

So that I assume that you use Lua 5.2 or 5.3 that might dropped automatic loading of `io` library as it did for `coroutines`.

375gnu:
@GottfriedSp please test #65

GottfriedSp:
Ok, I try the gotoobject.

The other thing I found in celx.cpp line 744, this i changed to:
`        if(strcmp(c_p, "y") == 0 || strcmp(c_p, "Y") == 0)`
The qt sends from CelestiaGlWidget::keyPressEvent line 535 the key 'Y':
`            appCore->charEntered('Y');`
Now it works perfectly.
From Windows native the key 'y' comes from MainWindowProc() line 4291:
`            appCore->charEntered(utf8CharCode, modifiers);`

I added the commit into my Pull request. (Each new Push from my local repository added to the Pull request to the open Pull request here, and I don't know why)

Current Lua: 5.1.5
I try to update to Lua 5.3.4 but I think this works also.

375gnu:
If you use windows, then the only supported Lua so far is 5.1. Only Unix builds support Lua 5.2 & 5.3.

That's quite strange that you have io==nil with 5.1, my Celestia 1.6.2 linked against Lua 5.1 works with your script.

GottfriedSp:
[...]

PR #65, opened 2018-07-04
[DON'T MERGE!] LUA: load luaopen_io library (Closes: #64)

375gnu:


375gnu:
This is not required. `void LuaState::requestIO()` loads required libraries.

PR #67, opened 2018-07-04
fixed file-io with qt, check key 'y' and 'Y'.

GottfriedSp:
Comming from CelestiaGlWidget::keyPressEvent() case VK_Y:
Checking 'y' and 'Y' if user has caps lock enabled or pressed the 'Y' key

### DIFF
--- a/src/celestia/celx.cpp
+++ b/src/celestia/celx.cpp
@@ -741,7 +741,7 @@ bool LuaState::charEntered(const char* c_p)
     if (ioMode == Asking && getTime() > timeout)
     {
         int stackTop = lua_gettop(costate);
-        if (strcmp(c_p, "y") == 0)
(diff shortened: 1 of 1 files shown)


375gnu:
Actually your change is a workaround, the proper fix should be applied to celestiacore.cpp around line 1396. But i don't oppose merging of this until the proper fix is made.

375gnu:
Line 1247 is even a better location. `static bool getKeyName(const char* c, int modifiers, char* keyName, unsigned int keyNameLength)`

GottfriedSp:
Do we have no problem with lua scripts with "celestia_keyboard_callback" function, when someone check for upper and lower case letters?
Ah, ok, this should not happens, but must be tested I think [custom key functions](https://en.wikibooks.org/wiki/Celestia/Celx_Scripting/Q%26A/Assigning_Custom_Key_Functions_with_CELX)

PR #69, opened 2018-07-05
common latin keystroke out of switch

GottfriedSp:
I removed the simple latin keys out of switch, the upper and lower case works now for all keys with on screen search.
This should be compared with Cyrillic keyboard.
Reduces code size by 230 lines, binary size and complexity.

The `#if 0` to `#endif` can be removed if Cyrillic keyboard works.


src/celestia/qt/qtglwidget.cpp
```diff
@@ -445,14 +445,25 @@ void CelestiaGlWidget::keyPressEvent( QKeyEvent* e )
     if (e->modifiers() & ControlModifier)
         modifiers |= CelestiaCore::ControlKey;
 
-    switch (e->key())
+    int ch = e->key();
```
375gnu: do we really need this part?

GottfriedSp: We can remove this and change the if to:
```
-    int ch = e->key();
+    if ((e->key() >= 'B' && e->key() <= 'Y'))
+    {
+        if (e->modifiers().testFlag(Qt::ControlModifier))
+            appCore->charEntered(e->text().toUtf8().data(), modifiers);   // [Ctrl+ch]
+        else if (e->modifiers().testFlag(Qt::ShiftModifier))
+            appCore->charEntered(static_cast<char>(e->key()));        // toupper(ch)
+        else
+            appCore->charEntered(static_cast<char>(tolower(e->key())));

```
But the compiler should create the same binary code.

375gnu: I meant not the single line `int ch = e->key();` but the whole chunk
```+    if ((e->key() >= 'B' && e->key() <= 'Y'))
+    {
+        if (e->modifiers().testFlag(Qt::ControlModifier))
+            appCore->charEntered(e->text().toUtf8().data(), modifiers);   // [Ctrl+ch]
+        else if (e->modifiers().testFlag(Qt::ShiftModifier))
+            appCore->charEntered(static_cast<char>(e->key()));        // toupper(ch)
+        else
+            appCore->charEntered(static_cast<char>(tolower(e->key())));```

What doesn't work with it and how your code fixes it?

Other issues: why don't you check A & Z(*)? How this code behaves if both CapsLock & Shift are pressed?

If your additions are required then you have added them into the wrong location: they should be in `default` section of `switch` so you don't need to check for 'B' < key 'Y'.

* - I know the answer.

And question for the whole team: should we distinguish lower- & uppercase letters in this case?

GottfriedSp: The VK_B (0x42) is a uppercase 'B' and there is no VK_A or VK_Z, so I check for 'B' to 'Y'

To check previous with if before switch, is much faster than in default. for the default, the CPU must check first all cases before go to default mark.
The default handle first the special keys with function handleSpecialKey(), there is another switch case for detecting cursor and other keys.
switch blocks have high complexity when you check with a code analyser.

The lower case is checked the lines before
    if (e->modifiers() & ShiftModifier)
        modifiers |= CelestiaCore::ShiftKey;

Ah, ok, I see in the handleSpecialKey() the key 'A' and 'Z' is also checked in case and call later the function:   appCore->charEntered(e->text().toUtf8().data(), modifiers);

I think the lower/uppercase letters calculated from e->text().toUtf8().data() is much overhead to convert each latin letter to UTF-8
I saw, the key 'A' and 'Z' must be handled in handleSpecialKey(), when I change from 'A' to 'Z' the speed can not be changed - 'A' increase speed, 'Z' decreased speed.

375gnu: You haven't answered the main question: What doesn't work without your addition and how it fixes that?

GottfriedSp:
For me, all works fine with lower ans uppercase letters.
I read the comment: `// Support for Cyrillic keyboard`
I have no Cyrillic keyboard, so I can't test this.
When this works with Cyrillic keyboard too, the disabled code block with `#if 0` can be removed

375gnu:
Either for me. Works without any additional code.

That "Support for Cyrillic keyboard" will not work properly if your english layout is dvorak or another non-default one. While the original idea is ok the implementation is terrible, it should be more generic.

I also not happy with usage of toUtf8(), because, for example, chinese & japan users prefer their own encodings.

So my opinion is that we should just revert those 2 commits from @Alexell.

GottfriedSp:
I have a QWERTZ (german) and a QWERTY (english) keyboard
The keys a-z handled as latin letters this is default ASCII table 7-bit, all other as UTF-8 keys.

I tested unused key 'ä' (0xc4) - unicode = U+00E4 -> UTF-8 = 0xC3 0xA4.
The textEnterMode encodes the UTF-8 and displays 'ä' and use letter 'a' when searching for objects.
This is correct functionality for me.

Other languages can be handled in switch case as  switch can be added with other languages if needed.
To the cases 1102, 1073, ... I didn't come, I think theese are the special keys of Cyrillic keyboard.
I removed the case with latin letters, now the code is more readable. a possible comment to each number would be helpfull for later reading the code.

375gnu:
I use cyrillic keyboard and everything works without that ugly hack. We should remove it. I won't be surprised if Qt5 itself handles different keyboards internally.

375gnu:
#79 is a better version

PR #73, opened 2018-07-08
fixed linux QT build

TheBigW:
fixes the Linux QT build on Ubuntu 18.04


375gnu:
Hi @TheBigW, autotools are not supported any more as the oly supported target is a Qt5 based one with qmake.

TheBigW:
Thanks, I didn't know that and just was looking for an easy way to get it running in Linux again. I will definitely give qmake a try (so far compilation fails after calling qmake and make....).

If qmake is the way to go would it make sense to dump all the auto-tools remnants? If auto-tools are not supported anymore anyway and it is not removed would it hurt to include the few changes to make it work?

PR #75, opened 2018-07-12
add missing model files for mars mission

GottfriedSp:
The model files was missing from the mars mission "Mars Odyssey"


375gnu:
Could you provide information about source of that models and their license?

GottfriedSp:
The sources for this models i copied from 1.6.1 from the models directory, I think this was in 1.6.1-src in the models-src directory and taken from there when creating an installer.
In earlier versions 1.6.0 and lower, these models also the same.

375gnu:
Just a thought: maybe worth to convert them to cmod?

Alexell:
@375gnu Yes, cmod should load faster.

Issue #70, opened 2018-07-08
Full screen does not work


GottfriedSp:
Full screen does not work with qt.
It resizes the window to max size but does not change to full screen.

375gnu:
Works for me. What OS & Qt version do you use?

GottfriedSp:
The menuBar, the toolBars and the solarSystemBrowser is visible in full screen.
With the native windows interface all of this is hidden.
This should be hidden like in other programs (vlc media player, stellarium, media player, presentation programms, ...)

Windows 7 64bit
Qt 5.6.3

375gnu:
So, I suppose it's just not implemented. Do you wish to?

GottfriedSp:
I can try this, but my vacation is over and could be taken some time.

GottfriedSp:
Pull request is done with #81

375gnu:
this was fixed some time ago

PR #81, opened 2018-07-15
Fullscreen

GottfriedSp:
This is the fixes for issue #70
I added additionally a visual studio 2005 project with Qt Visual Studio Plugin

Switching to full screen saves the current tool bars in registry and disable them.
Switching back to window, the settings was restored to the tool bars

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -907,7 +907,10 @@ void CelestiaAppWindow::slotToggleFullScreen()
         // Switch to window
         menuBar()->setFixedHeight(menuBar()->sizeHint().height());
         setWindowState(Qt::WindowNoState);
-        readSettings();
+        QSettings settings;
+        settings.beginGroup("MainWindow");
+        settings.setValue("Fullscreen", false);
+        readSettings();     // restore last windowed settings
     }
     else
     {


375gnu:
Hi @GottfriedSp!

Some trivial issues:
* no need for 'removed unused header' as native win32 backend is deprecated
* please use just `(float)` instead of `static_cast<float>()` to be consistent across th whole codebase
* please don't change the shortcut (Shift-F11 -> F11)

Much more important issue: you want to change application behavior you don't like but i believe there are people who prefer the current one. What I'd rather like to see is a configuration option so users can choose a fullscreen window kind they want (with or without widgets).

GottfriedSp:
Ok, I reverted to Shift-F11 to be compatible with user guide. I changed because other programms (Firefox, Opera, ...) use this button. Celestia with native windows uses the key "Alt+Enter"
I saw some static_cast in code. next time i use C-style casts.

I save current views and disable the menu with size 0.
The native windows Celestia was like Firefox, when moving to top of fullscreen, the menubar scrolled down and the menu can be selected. But I think this is not easy with Qt.
But when we add shortcuts to the widgets, these can be activated with fullscreen.
It is possible to save the widgets in fullscreen to a other registry keyword, so the widgets in fullscreen can also be stored if the user want to use widgets.

I check if I can enable the menuBar when move with mouse to top of window, to be able to aktivate the other widgets.

GottfriedSp:
In the user guide i found in page 12:

> Celestia will immediately switch to Full Screen Mode and draw the screen in that pixel size, filling up the entire screen.  All of the toolbars, sliders and side borders will disappear.

> 31)	[Alt+Enter] is a useful keyboard shortcut that toggles between Windowed Mode and Full screen. 

Was the Alt+Enter also in gtk and kde?

GottfriedSp:
"Alt+Enter" does not work wit Qt.
Key F11 is used for movie capture

Alexell:
Is the work completed or waiting for new commits in this request?

PR #86, opened 2018-07-29
port left unported cel commands to eigen

375gnu:
Celestia has been moving to Eigen usage but there still some code pieces using old vector math.

This PR changes CEL script engine to use Eigen. Another PR will address more complicated case of CELX.

As far as I tested it there is no differences between the old an new code except for `setorientation {angle axis}` command, but I believe that my code is correct because it explicitly normalizes `axis` while the old code does not, so actually the old code incorrectly processed not normalized vectors.

By default the new, Eigen based, code is used, to enable the old one define preprocessor variable `__CELVEC__`.


375gnu:
Don't merge as I pushed some incomplete code by mistake.

375gnu:
i've remove unfinished commits, so these 3 can be merged

375gnu:
Finished completely.

375gnu:
I have found a bug in this code chunk and  while I have a workaround for it I don't gonna push it, so postpone merging for a while until I found a proper fix.

375gnu:
The bug is fixed, can be merged.

PR #88, opened 2018-08-06
Modernize to C++ 11, part I

375gnu:
DONE:
1. run clang-tidy and then cleaned up the result
1. replaced own implementation of Timer with the one from C++ library
1. replaced own typedefs with standard types
1. replaced some astro:: methods with templates
1. some cleanups

TODO:
1. replace raw pointers with shared_ptr and unique_ptr where appropriate, I have some commits done locally but might need to review them
1. replace own directory class with std::experimental::filesystem if the latter is supported by all required compilers
1. continue to convert to modern syntax if there any C++98/03 pieces left


375gnu:
https://github.com/CelestiaProject/Celestia/pull/88/commits/84d69c1491d2ea49848c1c67ce984c3b8dd4de47 should be changed as it uses C++14 features.

pirogronian:
Isn't eliminating all raw pointers too radical change for now? Or, mayby it's better to do changes gradually, not all at once?

375gnu:
I'm not going to replace all raw pointers at least because there are places like
```
static Foo defaultFoo;

class Bar{
Foo* foo;
void set_foo(Foo* f){ foo = f==nullptr ? & defaultFoo : f; }
}
```

such code requires refactoring.

Actually I have more interesting tasks for now :)

PR #96, opened 2018-09-29
Relocation of LUA interface implementation code to improve code organisation.

pirogronian:
Hereby is try to overcome madness of lua related code. Overbloated celx.cpp is divided to LuaState, CelxLua and lua celestia object related code files. Moreover, I included nice lua ++ wrapper for future use, as I plan to, using it, completely rewrite lua interface implementation, make it much more simple, elegant and understandable.
Code was tested with randomly selected scripts.


375gnu:
i haven't checked your code changes carefully yet, but i've noticed that you name files incorrectly (e.g. LuaState.cpp or CelxLua.cpp). Could you name them like other celestia files (e.g. lua_state.cpp or celx_lua.cpp)?

I'll do a more complete review later when i have time.

pirogronian:
I can rename these files, although I arleady have created many new files "incorectly" named and I believed it's ok. Personally I dont like Celestia filenames convention.

375gnu:
Personal preferences are not a good reason to make ad-hoc changes.

I have a question regarding that wrapper, what lua version does it support?

And please make proper commits: 1st for celx restructure and 2nd for the wrapper.

pirogronian:
Celestia filenames convention was already ad-hoc in my opinion... But never mind. I changed file names as You suggested. Such as commits.
Quessing from cmake files, wrapper supports Lua 5.2 and LuaJIT 5.1. I think it can be easly prepared for different Lua versions using macros.
I agree with postponing this PR. I also thinked about replacing native celestia objects with C++11 standard libs...

pirogronian:
Celestia lua object, defined  now in  celx_celestia.cpp

PR #100, opened 2018-10-07
Build upon Qt interface control over FPS rate.

pirogronian:
FPS rate value can be set to default, 15, 30, 60, 120 or custom (1-2048), via View menu or Shift+F shortcut. Intended  to limit  FPS rate with high CPU load or on weak machines.

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -39,6 +39,7 @@
 #include <QDesktopServices>
 #include <QDesktopWidget>
 #include <QUrl>
+#include <QInputDialog>
 #include <vector>
 #include <string>
 #include <cassert>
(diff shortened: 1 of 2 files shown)


375gnu:
But, frankly speaking, I doubt that limiting FPS will help weak hardware. And for powerful one we have vsync.

pirogronian:
Done as You suggested. Maybe it's little help with performance, but on my machine I fell difference.

src/celestia/qt/qtappwin.cpp
```diff
@@ -401,12 +405,26 @@ void CelestiaAppWindow::init(const QString& qConfigFileName,
     connect(fullScreenAction, SIGNAL(triggered()), this, SLOT(slotToggleFullScreen()));
     viewMenu->addAction(fullScreenAction);
 
+    QMenu *fpsMenu = new QMenu(_("FPS rate"), this);
+    
+    std::vector<int> fps_array = { 0, 15, 30, 60, 120 };
```
375gnu: why vector instead of array? array is faster in this case as it's a tiny wrapper around c-style arrays.

pirogronian:
Perhaps due to my limited knowledge of C++11 standard. I never used std::array before and overlooked it in Your example. Thanks for review. Edited to use array.

src/celestia/qt/qtappwin.cpp
pirogronian: Perhaps due to my limited knowledge of C++11 standard. I never used std::array before and overlooked it in Your example. Thanks for review. Edited to use array.

pirogronian:
About vsync, I recently searched the code and found only one place, that could be related to it: implementation of `paintGL()` method in `CelestiaGlWidget` class (which call `celestiacore->draw()`). Simulation and widget update id done in every "tick", which is driven by timer with arbitrate frequency (which is reasonable, as Celestia may have running script, or, from very recently, playing audio).

PR #101, opened 2018-10-12
Replace deprecated QGLWidget with QOpenGLWidget

pirogronian:
It appears, that GLEW is incompatibile with Qt even on source code level. However, relocation of GLEW initialization was caused by delayed OpenGL context initalization in the new class.

### DIFF
--- a/src/celestia/qt/qtglwidget.cpp
+++ b/src/celestia/qt/qtglwidget.cpp
@@ -23,6 +23,7 @@
 #include <QPaintDevice>
 #include <QMouseEvent>
 #include <QSettings>
+#include <QMessageBox>
 
 #ifndef DEBUG
 #  define G_DISABLE_ASSERT
@@ -82,7 +83,7 @@ const unsigned int DEFAULT_TEXTURE_RESOLUTION = medres;
 
 
 CelestiaGlWidget::CelestiaGlWidget(QWidget* parent, const char* /* name */, CelestiaCoreApplication* core) :
-    QGLWidget(parent)
+    QOpenGLWidget(parent)
 {
     setFocusPolicy(Qt::ClickFocus);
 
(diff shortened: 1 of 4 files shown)


375gnu:
What Qt version QGLWidget was deprecated?

src/celestia/qt/qtglwidget.cpp
```diff
@@ -147,6 +148,24 @@ static GLContext::GLRenderPath getBestAvailableRenderPath(const GLContext& glc)
 
 void CelestiaGlWidget::initializeGL()
 {
+    GLenum glewErr = glewInit();
+    if (glewErr != GLEW_OK)
+    {
+        QString rn;
+        switch(glewErr) {
```
375gnu: According to https://codeyarns.com/2015/09/14/how-to-check-error-in-glew/ GLEW uses only 4 errors (I'm too lazy to recheck myself.)

pirogronian: That's right indeed.

375gnu:
inWith this change I have a lot of error messages every single second:
```QXcbConnection: XCB error: 3 (BadWindow), sequence: 609, resource id: 119537691, major code: 148 (Unknown), minor code: 2```

src/celestia/qt/qtglwidget.cpp
375gnu: `switch` doesn't use Celestia's formatting.

pirogronian:
I didn't remeber such a behaviour and just recompiling this branch to check it. However, it appears to be qt debug output and can be ignored and hide: [QXcbConnection: XCB error: 3 (BadWindow) is driving me crazy !](https://www.reddit.com/r/archlinux/comments/7si1jv/qxcbconnection_xcb_error_3_badwindow_is_driving/dt5d3aq/)

PR #103, opened 2018-10-23
[NOT FOR MERGE] Use Sol2 Lua wrapper to implement celx_celestia and celx_object.

pirogronian:
This is preparation step for easy implementation of Lua interface for Audio3D branch features. Also, it can be a step toward better abstraction layer between UI, Lua and core/engine.

It's build upon #96 and contains it.


375gnu:
Sol2 requires C++14.

pirogronian:
I can't confirm that. Both documentation suggest older standards support and I also tested it by placing `CXXFLAGS += -std=c++11`in `celestia.pro` and make some sources which includes sol2.

PR #115, opened 2018-11-14
FPS control menu polishing

### LINKED ISSUE #111: FPS control menu should use checkboxes
FPS control menu should use checkboxes instead of plain menu entries. Also please move 'Limit Frame Rate' entry into it (trnaslating it into "Enable Vsync").

pirogronian:
Make FPS control menu checkable, move VSync option to it and rename properly. Fixes #111 .

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -92,8 +92,8 @@ static const int CELESTIA_MAIN_WINDOW_VERSION = 12;
 static CelestiaAppWindow* MainWindowInstance = nullptr;
 static void ContextMenu(float x, float y, Selection sel);
 
-static int fps_to_ms(int fps) { return fps ? 1000 / fps : 0; }
-static int ms_to_fps(int ms) { return ms ? 1000 / ms : 0; }
+static int fps_to_ms(int fps) { return fps > 0 ? 1000 / fps : 0; }
+static int ms_to_fps(int ms) { return ms > 0? 1000 / ms : 0; }
 
 // Progress notifier class receives update messages from CelestiaCore
 // at startup. This simple implementation just forwards messages on
@@ -135,7 +135,6 @@ public:
 private:
     QWidget* parent;
 };
-
 
 CelestiaAppWindow::CelestiaAppWindow(QWidget* parent) :
     QMainWindow(parent)
@@ -1228,25 +1227,31 @@ void CelestiaAppWindow::createMenus()
     textureResolutionMenu->addAction(actions->mediumResAction);
     textureResolutionMenu->addAction(actions->highResAction);
 
-#ifdef VIDEO_SYNC
-    displayMenu->addSeparator();
-    displayMenu->addAction(actions->toggleVSyncAction);
-#endif
-
-    QMenu *fpsMenu = new QMenu(_("FPS rate"), this);
+    QMenu *fpsMenu = new QMenu(_("&FPS control"), this);
 
     std::array<int, 5> fps_array = { 0, 15, 30, 60, 120 };
-
-    for (auto ifps : fps_array) {
-        QAction *fps = new QAction(ifps == 0 ? _("Auto") : QString::number(ifps), this);
+    QActionGroup *fpsg = new QActionGroup(this);
+    QAction *fps;
+    for (auto ifps : fps_array)
+    {
+        fps = new QAction(ifps == 0 ? _("Auto") : QString::number(ifps), this);
+        fps->setCheckable(true);
         fps->setData(ifps);
-        connect(fps, &QAction::triggered, this, &CelestiaAppWindow::setFPS);
+        fpsg->addAction(fps);
+        QObject::connect(fps, &QAction::triggered, this, &CelestiaAppWindow::setCheckedFPS);
         fpsMenu->addAction(fps);
     }
-    QAction *fps = new QAction(_("Custom"), this);
+    fps = new QAction(_("Custom"), this);
+    fps->setCheckable(true);
     fps->setShortcut(QString("Shift+F"));
-    connect(fps, &QAction::triggered, this, &CelestiaAppWindow::setCustomFPS);
+    fpsg->addAction(fps);
+    fpsg->setExclusive(true);
+    QObject::connect(fps, &QAction::triggered, this, &CelestiaAppWindow::setCustomFPS);
     fpsMenu->addAction(fps);
+#ifdef VIDEO_SYNC
+    fpsMenu->addSeparator();
+    fpsMenu->addAction(actions->toggleVSyncAction);
+#endif
     displayMenu->addMenu(fpsMenu);
 
     /****** Bookmark menu ******/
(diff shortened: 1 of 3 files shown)


src/celestia/qt/qtappwin.cpp
```diff
@@ -92,8 +92,8 @@ static const int CELESTIA_MAIN_WINDOW_VERSION = 12;
 static CelestiaAppWindow* MainWindowInstance = nullptr;
 static void ContextMenu(float x, float y, Selection sel);
 
-static int fps_to_ms(int fps) { return fps ? 1000 / fps : 0; }
-static int ms_to_fps(int ms) { return ms ? 1000 / ms : 0; }
+static int fps_to_ms(int fps) { return fps > 0 ? 1000 / fps : 0; }
```
375gnu: Is it possible to use mks instead of ms? Then you'll have a more correct result.

pirogronian: It's impossible with qt timers. If c++ std library offer more accurate timer, then maybe...

pirogronian:
I can implement menu netries as checkboxes, however it will be misunderstanding of checkboxes. In my entries are radio buttons, because only one option have to be checked at a time. Checkboxes are independend from each other. If other exclusive actions groups are implemented using checkboxes, this is, in my opinion, wrong.

375gnu:
You can check "Texture resolution" and "Star style", they use checkboxes. They are mutually exclusive. Technically you are right, but in that case you'll have to rewrite other menus. And radio buttons look ugly :)

375gnu:
One more question. It seems that on application startup no menu entries are selected, aren't they?

pirogronian:
Yes, texture resolution and star style are examples of incorrect use of checkboxes :-) In this case I'm really unsure I should follow your suggestions... I', afraid of going against library creators intentions and against general GUI designation standards. Moreover, this can mislead some users. I'd rather remake other exclusion menus with radio buttons. Radio buttons look can be dependend on used qt style. Are you shure, that checkboxes are absolutely necessary?
And yes, at start there is no selection. I started with mechanism selecting items based on actual timer settings, but after strange behaviour mess I decided to radically simplify everything.

375gnu:
I've found a bug. Open "Custom FPS", press Cancel, check "FPS Control" -> Custom is selected instead of a previous one.

375gnu:
s/enaugh/enough

src/celestia/qt/qtappwin.cpp
```diff
@@ -136,6 +136,47 @@ class AppAlerter : public CelestiaCore::Alerter
...
+
+        m_group = new QActionGroup(p);
+        for (auto ifps : fps_array)
+        {
+            fps = new QAction(ifps == 0 ? _("Auto") : QString::number(ifps), p);
+            fps->setCheckable(true);
+            fps->setData(ifps);
+            m_map[ifps] = fps;
+            m_group->addAction(fps);
+    }
```
375gnu: incorrect indentation

src/celestia/qt/qtappwin.cpp
```diff
@@ -136,6 +136,47 @@ class AppAlerter : public CelestiaCore::Alerter
     QWidget* parent;
 };
 
+class FPSActionGroup
+{
+    QActionGroup *m_group;
+    std::map<int, QAction*> m_map;
```
375gnu: better `m_actions`, as m_map is meaningless name
the same for m_custom and m_last.

The should be self documenting, e.g. customFPSAction, lastFPS or smth similar.

src/celestia/qt/qtappwin.cpp
```diff
@@ -136,6 +136,47 @@ class AppAlerter : public CelestiaCore::Alerter
...
+        m_custom->setCheckable(true);
+        m_custom->setShortcut(QString("Shift+F"));
+        m_group->addAction(m_custom);
+        m_group->setExclusive(true);
+    }
+
+    std::map<int, QAction*> &actions() { return m_map; }
+    QAction *custom() { return m_custom; }
+    int last() { return m_last; }
+    void update(int fps)
```
375gnu: This method's implementation shouldn't be inside the class definition because it will be inlined.

src/celestia/qt/qtappwin.cpp
```diff
@@ -1463,23 +1499,28 @@ void CelestiaAppWindow::closeEvent(QCloseEvent* event)
...
     timer->setInterval(fps_to_ms(fps));
+    fpsActions->update(fps);
 }
 
 void CelestiaAppWindow::setCustomFPS()
 {
     bool ok;
     int fps = QInputDialog::getInt(this, _("Set custom FPS"), _("FPS value"), ms_to_fps(timer->interval()), 0, 2048, 1, &ok);
     if (ok) setFPS(fps);
+    else fpsActions->update(fpsActions->last());
```
375gnu: New lines?

src/celestia/qt/qtappwin.cpp
pirogronian: Indeed, however custom() and last() should be rather inline anyway.

src/celestia/qt/qtappwin.cpp
pirogronian: Indeed. I use short names to avoid lot's of typing... ;-)

src/celestia/qt/qtappwin.cpp
```diff
@@ -1474,20 +1509,26 @@ void CelestiaAppWindow::setCheckedFPS()
...
 {
     bool ok;
-    int fps = QInputDialog::getInt(this, _("Set custom FPS"), _("FPS value"), ms_to_fps(timer->interval()), 0, 2048, 1, &ok);
-    if (ok)  setFPS(fps);
+    int fps = QInputDialog::getInt(this,
+                                   _("Set custom FPS"),
+                                   _("FPS value"),
+                                   ms_to_fps(timer->interval()),
+                                   0, 2048, 1, &ok);
+    if (ok) setFPS(fps);
```
375gnu: please move `setFPS` to a new linw, the same for else branch

src/celestia/qt/qtappwin.cpp
```diff
@@ -136,6 +136,52 @@ class AppAlerter : public CelestiaCore::Alerter
...
+    m_customAction = new QAction(_("Custom"), p);
+    m_customAction->setCheckable(true);
+    m_customAction->setShortcut(QString("Shift+F"));
+    m_actionGroup->addAction(m_customAction);
+    m_actionGroup->setExclusive(true);
+}
+
+void FPSActionGroup::updateFPS(int fps)
+{
+    if (fps < 0 ) fps = 0;
```
375gnu: please move `fps = 0` to a new line

src/celestia/qt/qtappwin.cpp
```diff
@@ -1228,25 +1273,20 @@ void CelestiaAppWindow::createMenus()
...
 
-    std::array<int, 5> fps_array = { 0, 15, 30, 60, 120 };
-
-    for (auto ifps : fps_array) {
-        QAction *fps = new QAction(ifps == 0 ? _("Auto") : QString::number(ifps), this);
-        fps->setData(ifps);
-        connect(fps, &QAction::triggered, this, &CelestiaAppWindow::setFPS);
-        fpsMenu->addAction(fps);
+    fpsActions = new FPSActionGroup(this);
+    for (const auto a : fpsActions->actions())
```
375gnu: you have forgotten `&` after `auto`. you don't need a new intermediate object here.

PR #116, opened 2018-11-14
[DON'T MERGE] New rendering flags to show/hide solar system objects

375gnu:


375gnu:
TODO: win32, labels for all new objects

375gnu:
@pirogronian this is mostly question of style, but in C functions you should use it with 64 bit types or compilers will warn you about casting int64 to int and so on.

PR #117, opened 2018-11-15
Remove size limit of eclipsing satelites.

### LINKED ISSUE #114: Remove size limit for eclipsing satelites
A user's request

pirogronian:
I believe it fixes #114 , but to be sure I need exact bodies configuration to look at and compare.


375gnu:
I can't tell if it really fixes #114 and doesn't introduces new bugs without testing. But what you can change now: s/bottom/size/, update the commentary to reflect your changes.

375gnu:
Let's postpone this pull request. I'm unsure that we should really change something here.

PR #118, opened 2018-11-15
Fix: In QtEclipseFinder "All eclipses" button acted as "Lunar eclipses".

pirogronian:


pirogronian:
It would be easier to me to make findEclipses flag agnostic, than imlement proper flag interpretation inside it.

375gnu:
You can do it flag agnostic, as it checks flags inside. The problem that findEclipses accepts only solar or lunar eclipses, not both of them. If you unable to implement a proper fix, let's merge this PR as I'm not going to make a proper fix now or in the nearest future.

PR #125, opened 2018-11-19
Relocations celestia, celscript, font, image and texture objects code…

pirogronian:
… from overbloated celx.cpp to appriopriate files. Second try.


src/celestia/celx.cpp
```diff
@@ -96,15 +97,15 @@ bool CelxLua::mapsInitialized = false;
 static const double MaxTimeslice = 5.0;
 
 // names of callback-functions in Lua:
-static const char* KbdCallback = "celestia_keyboard_callback";
-static const char* CleanupCallback = "celestia_cleanup_callback";
+const char* KbdCallback = "celestia_keyboard_callback";
```
375gnu: is it possible to move this and other ex-static things to celx_celestia.cpp? I just don't like that they are visible to other modules now.

Anyway, even if it's impossible, I approve the change. But please change commit title to something more simple like 'Move lua methods to celx_celestia and celx_misc', and rebase on current master, it's conflicting now.

pirogronian:
@375gnu It compile and run fine, but please take a look again. There were strange conflicts in celx.cpp.

375gnu:
Looks good for me.

PR #127, opened 2018-11-20
Amend location types

375gnu:
large -> largeringed (less confusing name)
trench -> ring (we already have a trench-like feature)


375gnu:
LargeRinged are definitely a planetary feature, it's a "Cryptic ringed features" (see https://planetarynames.wr.usgs.gov/DescriptorTerms)

With Rings I understand your confusion, but currently rings are named using locations but using "Other" type. Until we create something more suitable for this purpose I favor location features usage for rings. On the other hand is it worth to invent something just to add names to rings?

Issue #126, opened 2018-11-20
Extend from_ssc to support dsc and stc fragments


375gnu:
1. Rename `celestia:from_ssc` to `celestia:loadfragment`;
1. Make it accepting 2 arguments: `type` = `set{"dsc", "ssc", "stc"}`, `fragment` = `string`;
1. The same for legacy scrips.

PR #129, opened 2018-11-21
Add loadfragment script function instead of from_ssc (Closes: #126)

375gnu:


375gnu:
You can add implementation of `SolarSystemCatalog::load` to your TODO list :)

PR #132, opened 2018-11-22
User-defined generic category support (core + lua).

pirogronian:


375gnu:
Why to add additional ancestor to all object? Why not just add a categories as a private members to them?

pirogronian:
Well... I always thought that good programmer should minimize code duplication.
What means "categories as a private members"?

375gnu:
You have `class Foo : Object`, instead of this I suggested `class Foo { private: Object* o; }`

But now think that your approach is better as you need some public functions to work with categories. Maybe you just need a better name like `UserCategoryHelper` instead of Object.

pirogronian:
> Maybe you just need a better name like UserCategoryHelper instead of Object.
Maybe. However, I also tend to minimize names length :-) Moreover, I've already thought a lot about to unify these three classess, because it looks really similar in many ways.

pirogronian:
These methods works, but for full comfort I want to fix one FIXME ;-)

src/celengine/object.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <string>
+#include <set>
+
+class UserCategory;
+
+class Object
+{
+    typedef std::set<UserCategory*> cat_set;
```
375gnu: According to Celestia's style guide it should be CategorySet.

src/celengine/object.h
```diff
@@ -0,0 +1,29 @@
...
+
+    cat_set *m_cats;
```
375gnu: I'd make it not a pointer but just `CategorySet m_cats`. Of course if you have tests that prove that it's worth to have such optimization here, then I'm okay with pointer.

src/celengine/object.h
```diff
@@ -0,0 +1,29 @@
...
+    bool _removeFromCategory(UserCategory*);
+public:
+    bool addToCategory(UserCategory*);
+    bool removeFromCategory(UserCategory*);
+    bool inCategory(UserCategory*) const;
+
+    bool addToCategory(const std::string&);
+    bool removeFromCategory(const std::string&);
+    bool inCategory(const std::string&) const;
+    int inCategories() const { return m_cats == nullptr ? 0 : m_cats->size(); }
```
375gnu: Misleading method name

src/celestia/celx_category.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "celx_internal.h"
+#include "celx_category.h" 
+#include <celengine/category.h>
+
+static int category_tostring(lua_State *l)
+{
+    CelxLua celx(l);
+
+    UserCategory *c = *celx.getThis<UserCategory*>();
+    std::string s = fmt::format("[UserCategory:{0}]", c->name());
```
375gnu: Could you replace it with `fmt::printf`? Just for consistence with other code. If `fmt::format` has advantages over `fmt::printf` then we'll allow its usage, but currently let's stick with a single API.

src/celestia/celx_internal.h
```diff
@@ -94,7 +95,160 @@ class CelxLua
     CelxLua(lua_State* l);
     ~CelxLua() = default;
 
+    static int localIndex(int n) { return lua_upvalueindex(n); }
+
+    /**** push-on-stack methods ****/
+
+    int push()
```
375gnu: It looks like these method should return `void`.

src/celestia/celx_internal.h
```diff
@@ -94,7 +95,160 @@ class CelxLua
...
+    }
+    template <typename T> T *newUserDataArray(int n)
+    {
+        return reinterpret_cast<T*>(lua_newuserdata(m_lua, sizeof(T) * n));
+    }
+    template <typename T> T *newUserDataArray(T *a, int n)
+    {
+        T *p = reinterpret_cast<T*>(lua_newuserdata(m_lua, sizeof(T) * n));
+        for (int i = 0; i < n; i++)
+            p[i] = a[i];
```
375gnu: Either `memcpy`  or `std::copy` are better choice.

src/celestia/celx_internal.h
```diff
@@ -94,7 +95,160 @@ class CelxLua
...
+        int n = a.size();
+        int i = 0;
+        T *array = celx.newUserDataArray<T>(n);
+        for (auto &it : a)
+        {
+            array[i] = it.second;
+            i++;
+        }
+        celx.push((n - 1));
+        auto lambda = [](lua_State *l)
```
375gnu: The function is too big to be lambda.

src/celestia/celx_internal.h
```diff
@@ -94,7 +95,160 @@ class CelxLua
...
+    bool isInteger(int n = 0) const { return lua_isinteger(m_lua, n); }
+    bool isNumber(int n = 0) const { return lua_isnumber(m_lua, n); }
+    bool isBoolean(int n = 0) const { return lua_isboolean(m_lua, n); }
+    bool isString(int n = 0) const { return lua_isstring(m_lua, n); }
+    bool isUserData(int n = 0) const { return lua_isuserdata(m_lua, n); }
+    bool isValid(int) const;
+
+    /**** get methods ****/
+
+    int getInt(int n = 0) const { return lua_tointeger(m_lua, n); }
```
375gnu: We already have `Celx::getSafe` family of funtions.

src/celestia/celx_internal.h
```diff
@@ -149,8 +308,41 @@ class CelxLua
                         const char* errorMessage = "Boolean argument expected",
                         bool defaultValue = false);
 
-    LuaState* getLuaStateObject();
+    template<typename T> T *safeGetUserData(int index = 0,
+                                            FatalErrors errors = AllErrors,
+                                            const char *errorMessage = "User data expected")
+    {
+        if (!safeIsValid(index))  return nullptr;
```
375gnu: Please don't use such style. There are of places in your code where it can be misunderstood due to this style.

src/celestia/celx_internal.h
```diff
@@ -149,8 +308,41 @@ class CelxLua
...
+        if (!isUserData(index))
+        {
+            if (errors & WrongType)
+                doError(errorMessage);
+            return nullptr;
+        }
+        return reinterpret_cast<T*>(lua_touserdata(m_lua, index));
```
375gnu: There are places with wrong style. Better:
```
if (isUserData(index))
    return reinterpret_cast<T*>(lua_touserdata(m_lua, index));

if (errors & WrongType)
    doError(errorMessage);

return nullptr;
```
Such code is much easier to understand.

src/celengine/object.h
pirogronian: I just realized, that this class will be instantiated by thausend or more times, so its size should be minimized. However, I didn't test how much memory take an empty set. This could depend on compiler or library implementation.

src/celestia/celx_internal.h
pirogronian: I did it for shorter code. Every push(...) function returns number of pushed elements (I know, for now it is 1 in every case), so I can write
`return push(...);`
instead of
`push();
return 1;`

src/celestia/celx_internal.h
pirogronian: So, here are unsafe (but faster) functions... Perhaps I wrote them for symmetry.

src/celestia/celx_internal.h
pirogronian: Do you mean `if - then` in one line?

375gnu: Yes

src/celestia/celx_internal.h
375gnu: so more proper to write as return push(...)

src/celestia/celx_internal.h
```diff
@@ -94,7 +95,137 @@ class CelxLua
...
+    {
+        CelxLua celx(m_lua);
+        int n = a.size();
+        T *array = celx.newUserDataArray<T>(n);
+        for (auto &it : a)
+        {
+            *array = value(it);
+            array++;
+        }
+        celx.push((n - 1));
```
375gnu: extra braces

src/celengine/object.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    UserCategory *c = UserCategory::find(s);
+    if (c == nullptr)
+        return false;
+    return removeFromCategory(c);
+}
+
+bool Object::isInCategory(UserCategory *c) const
+{
+    if (m_cats == nullptr)  return false;
```
375gnu: wrong style

src/celengine/object.cpp
```diff
@@ -0,0 +1,66 @@
...
+    return m_cats->count(c);
```
375gnu: != 0 missed

pirogronian: Rather new line before `return false;`...

375gnu: I mean `m_cats->count(c) > 0`

pirogronian:
category_getcategories is for now unimplemented due to Object <-> Selection incompatibilities.

src/celestia/celx_internal.h
```diff
@@ -149,8 +291,42 @@ class CelxLua
...
+        if (!isType(i, celxClassId(*a)))
+        {
+            if (fatalErrors & WrongType)
+                doError(msg);
+            return nullptr;
+        }
+        return a;
+    }
 
+    template<typename T> T *getThis(FatalErrors fatalErrors = AllErrors,
```
375gnu: Please replace `getThis` with more meaningful name.

src/celengine/object.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    if (m_cats == nullptr)
+        m_cats = new CategorySet;
+    m_cats->insert(c);
+    return true;
+}
+
+bool Object::addToCategory(UserCategory *c)
+{
+    if (!_addToCategory(c))  return false;
```
375gnu: still wrong style

src/celengine/object.cpp
```diff
@@ -0,0 +1,67 @@
...
+    {
+        delete m_cats;
+        m_cats = nullptr;
+    }
+    return true;
+}
+
+bool Object::removeFromCategory(UserCategory *c)
+{
+    if (!_removeFromCategory(c))  return false;
```
375gnu: and here

src/celestia/celx_category.cpp
```diff
@@ -0,0 +1,99 @@
...
+    const char *n = c->name().c_str();
+    return celx.push(n);
+}
+
+static int category_createchild(lua_State *l)
+{
+    CelxLua celx(l);
+
+    UserCategory *c = *celx.getThis<UserCategory*>();
+    const char *n = celx.safeGetString(2);
```
375gnu: check for nullptr

src/celestia/celx_object.cpp
```diff
@@ -1408,21 +1409,69 @@ static int object_setatmosphere(lua_State* l)
...
+    Selection *s = celx.getThis<Selection>();
+    bool ret;
+    if (celx.isUserData(2))
+    {
+        UserCategory *c = *celx.getUserData<UserCategory*>(2);
+        ret = s->object()->addToCategory(c);
+    }
+    else
+    {
+        const char *n = celx.safeGetString(2, AllErrors, "Argument to object:addtocategory() must be string or userdata");
```
375gnu: nullptr

src/celestia/celx_object.cpp
```diff
@@ -1408,21 +1409,69 @@ static int object_setatmosphere(lua_State* l)
...
+    Selection *s = celx.getThis<Selection>();
+    bool ret;
+    if (celx.isUserData(2))
+    {
+        UserCategory *c = *celx.getUserData<UserCategory*>(2);
+        ret = s->object()->removeFromCategory(c);
+    }
+    else
+    {
+        const char *n = celx.safeGetString(2, AllErrors, "Argument to object:addtocategory() must be string or userdata");
```
375gnu: nullptr

src/celestia/celx_category.cpp
pirogronian: I added additional check inside Celx_safeGetString.

src/celestia/celx_internal.h
pirogronian: Hmm, I have no idea. Legacy <classname>_this was more meaningful?

src/celestia/celx_category.cpp
375gnu: I did not understood it, I'll create a new issue for tis case. I just need to make some checks before this.

src/celestia/celx_internal.h
375gnu: Nope :)

375gnu: But I  don't have any good proposals so if you either don't have any, let's keep getThis.

src/celengine/category.cpp
```diff
@@ -0,0 +1,114 @@
...
+    for (auto it : m_catlist)
+        deleteCategory(it);
+}
+
+UserCategory::CategoryMap UserCategory::m_allcats;
+UserCategory::CategorySet UserCategory::m_roots;
+
+UserCategory *UserCategory::newCategory(const std::string &s, UserCategory *p)
+{
+    if (m_allcats.count(s))  return nullptr;
```
375gnu: count() > 0 and new line

src/celengine/category.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+UserCategory *UserCategory::createRoot(const std::string &n)
+{
+    return newCategory(n, nullptr);
+}
+
+UserCategory *UserCategory::find(const std::string &s)
+{
+    if (!m_allcats.count(s)) return nullptr;
```
375gnu: count > 0 and  new line

src/celengine/category.cpp
```diff
@@ -0,0 +1,114 @@
...
+//    std::cout << "UserCategory::find(" << s << "): exists\n";
+    return (*m_allcats.find(s)).second;
```
375gnu: m_allcats.find(s)->second

src/celengine/category.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+bool UserCategory::deleteCategory(UserCategory *c)
+{
+    if (!find(c->name()))  return false;
```
375gnu: new line

src/celengine/category.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include <set>
+#include <map>
+#include "object.h"
+
+class UserCategory
+{
+public:
+    typedef std::set<Object*>    ObjectSet;
```
375gnu: extra spaces

src/celengine/category.h
```diff
@@ -0,0 +1,55 @@
...
+    typedef std::set<UserCategory*> CategorySet;
+    typedef std::map<const std::string, UserCategory*> CategoryMap;
+
+private:
+    std::string   m_name;
+    ObjectSet      m_objlist;
+    CategorySet      m_catlist;
```
375gnu: extra spaces

src/celengine/category.h
```diff
@@ -0,0 +1,55 @@
...
+
+    bool addObject(Object*);
+    bool removeObject(Object*);
+    const ObjectSet &objects() const { return m_objlist; }
+    const CategorySet &children() const { return m_catlist; }
+
+private:
+    bool _addObject(Object*);
+    bool _removeObject(Object*);
+    void _insertChild(UserCategory *);
```
375gnu: extra space

src/celengine/object.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <string>
+#include <set>
+
+class UserCategory;
+
+class Object
```
375gnu: As I said previously `Object` is too generic and doesn't reflect its purpose.

src/celestia/celx_internal.h
```diff
@@ -149,8 +291,42 @@ class CelxLua
...
+                                         const char *msg = "Celx class expected")
+    {
+        T *a = safeGetUserData<T>(i, fatalErrors, msg);
+        if (!isType(i, celxClassId(*a)))
+        {
+            if (fatalErrors & WrongType)
+                doError(msg);
+            return nullptr;
+        }
+        return a;
```
375gnu:
```
    if (isType(i, celxClassId(*a)))
        return a;

    if (fatalErrors & WrongType)
        doError(msg);
    return nullptr;
```

375gnu: cause first we care about returned value

src/celestia/celx.cpp
```diff
@@ -1233,13 +1235,15 @@ const char* Celx_SafeGetString(lua_State* l,
...
+    const char *s;
+    if (!lua_isstring(l, index) || 
+        ((s = lua_tostring(l, index)) == nullptr))
     {
         if (fatalErrors & WrongType)
             Celx_DoError(l, errorMsg);
         return nullptr;
     }
-    return lua_tostring(l, index);
+    return s;
```
375gnu: please revert this change. if the object is `lua_isstring` then `lua_tostring` won't return nullptr. And you still return nullptr to the caller.

let me show a simple example:
```
#include <iostream>
#include <string>

const char* fn() {return nullptr;}

int main()
{
  std::string s = fn();
  std::cout << s << '\n';
}
```

Now compile and run it: `g++ t.cc && ./a.out`

You get:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted
```

And that's why I'm telling you that Celx_SafeGetString is not so safe and you should check it's return.

src/celestia/celx_category.cpp
375gnu:
```
category:createchild()
-- or
category:createchild(1)
```
and celestia is aborted

because you did not check for nullptr

pirogronian:
Understood.

pirogronian:
This was really long day... However, I know there is at least one bug.

src/celengine/selection.h
```diff
@@ -87,4 +96,9 @@ inline bool operator!=(const Selection& s0, const Selection& s1)
     return s0.type != s1.type || s0.obj != s1.obj;
 }
 
+inline bool operator<(const Selection& s0, const Selection& s1)
+{
+    return s0.type < s1.type || s0.obj < s1.obj;
```
375gnu: I'm not sure that you do what you actually want.
1. I suppose you should compare objects only if `s0.type == s1.type`
1. obj is a pointer, so pointer comparison is nonsense, do you meant `*s0.obj < *s1.obj`?

src/celengine/category.cpp
```diff
@@ -0,0 +1,114 @@
+#include <iostream>
+#include "category.h"
+
+UserCategory::UserCategory(const char *n, UserCategory *p) : m_name(n), m_parent(p) {}
+UserCategory::UserCategory(std::string n, UserCategory *p) : m_name(n), m_parent(p) {}
+UserCategory::~UserCategory() { cleanup(); }
```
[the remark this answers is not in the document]
375gnu: one more: if you use `std::string n`, then use `m_name(std::move(n))` to avoid copy creation

src/celengine/selection.h
375gnu: Pointer comparison must not be used here. It's valid for `!=` to ensure that s0 and s1 doesn't point to the same object. But not in this case.

```
if (s0.type != s1.type)
    return s0.type < s1.type;

// here goes more complicated code to work with s0.type == s1.type
// not an actual code, but pseudo code
switch (s0.type)
{
case Type_Nil: return false;
case Type_Body: return *static_cast<Body*>s0.obj < *static_cast<Body*>s1.obj;
...
}
```

pirogronian: Hmm, I see no problem. Pointers comparisions is standard procedure in c++ , as pointer is just unsigned integer. It is fast and reliable. I see also no ease workaround. (Casting to typed pointer changes nothing.) I used it mamy times with many kinds of containers with any problem.

pirogronian: Hmm, I see no problem. Pointers comparisions is standard procedure in c++ , as pointer is just unsigned integer. It is fast and reliable. I see also no ease workaround. (Casting to typed pointer changes nothing.) I used it mamy times with many kinds of containers with any problem.

375gnu: Okay, let me explain. You have two planets, Earth and Saturn. Both are Type_Body, but Earth is located in memory at 0x0100 and Saturn is located at 0x0010. Your code says that Saturn < Earth. That's absurd.

pirogronian: And this absurd is needed to store it in std::set. I hope nobody will try to compare two select objects, expecting anything else.

375gnu: Hi @pirogronian, I've just realized that you won't need this operator if you use `unordered_set` instead of `set`. It seams that you actually don't need ordering.

pirogronian: Segfaults eliminated, code simplified.

src/celengine/category.h
```diff
@@ -0,0 +1,65 @@
...
+    const ObjectSet &objects() const { return m_objlist; }
+    const CategorySet &children() const { return m_catlist; }
+
+private:
+    bool _addObject(Selection);
+    bool _removeObject(Selection);
+    void _insertChild(UserCategory *);
+    bool removeChild(UserCategory *);
+    void setParent(UserCategory*);
+    static bool _deleteCategory(UserCategory *);
```
375gnu: extra space

src/celengine/category.h
```diff
@@ -0,0 +1,65 @@
...
+    void cleanup();
+
+    static CategoryMap m_allcats;
+    static CategorySet m_roots;
+public:
+    static UserCategory *newCategory(const std::string&, UserCategory* parent = nullptr);
+    static UserCategory *find(const std::string&);
+
+    static const CategoryMap &getAll() { return m_allcats; }
+    static const CategorySet &getRoots() { return m_roots; }
```
375gnu: `getAll() const` and `getRoots() const`

src/celengine/category.h
```diff
@@ -0,0 +1,65 @@
...
+
+#include <set>
+#include <unordered_set>
+#include <map>
+#include "selection.h"
+
+class UserCategory
+{
+public:
+    struct Hasher
```
375gnu: i'm ok with the custom hasher, just a question - did it worked with a default one?

src/celengine/category.h
pirogronian: `In file included from src/celengine/category.cpp:3:

src/celengine/category.h:59:40: error: static member function ‘static const CategoryMap& UserCategory::getAll()’ cannot have cv-qualifier

     static const CategoryMap &getAll() const { return m_allcats; }
                                        ^~~~~
src/celengine/category.h:60:42: error: static member function ‘static const CategorySet& UserCategory::getRoots()’ cannot have cv-qualifier

     static const CategorySet &getRoots() const { return m_roots; }
                                          ^~~~~`
Internet says, that static members may not be const (nor virtual nor volatile etc.)

src/celengine/category.h
pirogronian: Default didn't exist. That's why I had to write own.

src/celengine/category.h
375gnu: i missed `static`, sorry

pirogronian:
1. Done.
2. The only information about dprintf I found here: [](http://man.freetechsecrets.com/dprintf.3.html). If this is it, I don't know how to use it.
3. Do you mean issues mentioned by you? I hope all are resolved. But I will do some tests yet.

375gnu:
>     2. The only information about dprintf I found here: http://man.freetechsecrets.com/dprintf.3.html. If this is it, I don't know how to use it.

I mean this macro: https://github.com/CelestiaProject/Celestia/blob/master/src/celutil/debug.h#L15

>     3. Do you mean issues mentioned by you? I hope all are resolved. But I will do some tests yet.

Incorrect style and **incorrect std::string usage in a constructor**
https://github.com/CelestiaProject/Celestia/pull/132/files#r237907787

Here you do extra operations. You added explicit return with value, but you don't use it.
https://github.com/CelestiaProject/Celestia/pull/132/files#r236565735

https://github.com/CelestiaProject/Celestia/pull/132/files#r237910377
https://github.com/CelestiaProject/Celestia/pull/132/files#r237910656

Optional. But proposed code is easier to understand.
https://github.com/CelestiaProject/Celestia/pull/132/files#r236568921

src/celestia/celx_internal.h
pirogronian: I dont understand. This is as I write.

pirogronian:
How to choose debug level?

375gnu:
> How to choose debug level?

compile with `-DDEBUG`. actuall https://github.com/CelestiaProject/Celestia/blob/master/src/celutil/debug.cpp is very simple

src/celestia/celx_internal.h
375gnu: Ahhh, you want to return values of **your** push-es. okay.

PR #138, opened 2018-11-26
Git submodules

375gnu:
Currently to play with only.


375gnu:
Glew should be reverted

pirogronian:
What's wrong with glew?

375gnu:
Ping @pirogronian . Are you ready to use git submodules?

pirogronian:
Shame to admit, I didn't pay a lot of attention to it (in particular I've never tested it on real example). But in theory it looks quiet clear. It shouldn't be much more difficult than switching to cmake.

PR #144, opened 2018-11-28
Initial cmake implementation

375gnu:
I've figured out how to workaround segfault on exit. It was caused by libcelestia.so so now I create a static library instead. Anyway for Windoze and Visual Studio we must build static library.

It's not perfect, and it's still a lot to do, but at least this is a good start point.

Usage example:
```
mkdir build-cmake
cd build-cmake
cmake -DENABLE_GTK=ON -DCMAKE_INSTALL_PREFIX=/opt/celestia -DCMAKE_CXX_FLAGS="-O2" ..
make -j2
```

If you want to test Gtk2 build then #143 must be merged.

Known issues: ugly :), with Visual Studio CMake <= 3.11 should be used due to cotire module.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,250 @@
+cmake_minimum_required(VERSION 3.1.0)
+
+if(POLICY CMP0072)
+  cmake_policy(SET CMP0072 NEW)
+endif()
+
+project(celestia)
+
+set(VERSION "1.7.0")
+#
+#
+#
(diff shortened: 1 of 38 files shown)


375gnu:
@pirogronian 

It's in ~ShaderManager() src/celengine/shadermanager.cpp line 303. Looks like double free.

I had it on a system with GCC 6.4, so I used valgrind to find problems and find some and after addition of some explicit `delete` the issue did not emerge anymore on that system. But now it emerges on a system with GCC 8.

Issue #150, opened 2018-12-02
Segfaults on exit with libcelestia.so


375gnu:
I suppose this [link](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables) explains the issue.

375gnu:
unreproducible anymore

PR #151, opened 2018-12-02
Remove some global vars

375gnu:
At least now I don't have double free on exit, see #150 .

Instead of 
```
static ShaderManager g_ShaderManager;
return g_ShaderManager;
```
I initially made
```
static ShaderManager* g_ShaderManager = nullptr;
if(g_ShaderManager == nullptr)
    g_ShaderManager = new ShaderManager;
return *g_ShaderManager;
```

Both variants don't have any issues, but the 1st is faster.

### DIFF
--- a/src/celengine/dsodb.cpp
+++ b/src/celengine/dsodb.cpp
@@ -41,13 +41,13 @@ using namespace std;
 
 
 // 100 Gly - on the order of the current size of the universe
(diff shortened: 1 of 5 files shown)


375gnu:
Oh lol, I did it wrong, the 2nd variant is more correct.

PR #164, opened 2018-12-08
[Do not merge] Translatable text object.

pirogronian:
After some tries, I decided to meet simplicity.


pirogronian:
As I already merged code with another approach, this PR is at least redundant.

PR #172, opened 2018-12-13
Categories loaded from catalog files.

pirogronian:


extras-standard/iss/iss.ssc
```diff
@@ -39,6 +39,9 @@
     Orientation [ 90 0 0 1 ]
 
     Albedo        0.10
+
+    Category "Spacecrafts"
+    Categories [ "NASA" "Space Stations" "LEO objects" ]
```
375gnu: Do we really need 2 similar entries?

src/celengine/catentry.cpp
```diff
@@ -23,11 +23,16 @@ bool CatEntry::addToCategory(UserCategory *c)
     return c->_addObject(toSelection());
 }
 
-bool CatEntry::addToCategory(const std::string &s)
+bool CatEntry::addToCategory(const std::string &s, bool f)
```
375gnu: Please rename `f` into something more meaningful like `createNew`

src/celengine/catentry.cpp
```diff
@@ -73,3 +78,22 @@ bool CatEntry::isInCategory(const std::string &s) const
...
+    Value *a = hash->getValue("Categories");
+    if (a == nullptr)
+        return;
+    ValueArray *v = a->getArray();
+    if (v == nullptr)
+        return;
+    for (auto it : *v)
+    {
+        cn = it->getString();
+        if (cn.size() > 0)
```
375gnu: `if (!cn.empty())`

src/celengine/catentry.cpp
```diff
@@ -73,3 +78,22 @@ bool CatEntry::isInCategory(const std::string &s) const
         return false;
     return isInCategory(c);
 }
+
+void CatEntry::loadCategories(Hash *hash)
```
375gnu: `bool` instead of `void`?

375gnu: yes

src/celengine/catentry.cpp
```diff
@@ -73,3 +77,26 @@ bool CatEntry::isInCategory(const std::string &s) const
...
 }
+
+bool CatEntry::loadCategories(Hash *hash)
+{
+    std::string cn;
+    if (hash->getString("Category", cn) && cn.size())
+    {
+        addToCategory(cn, true);
+        return true;
+    }
```
375gnu:
```
if (hash->getString("Category", cn)
{
    if (cn.empty())
        return false;
    addToCategory(cn, true);
    return true;
}
```

src/celengine/catentry.cpp
375gnu: `std::string::size() > 0` is considered a bad style in almost every C++ guide

src/celengine/catentry.cpp
```diff
@@ -73,3 +77,30 @@ bool CatEntry::isInCategory(const std::string &s) const
         return false;
     return isInCategory(c);
 }
+
+bool CatEntry::loadCategories(Hash *hash)
+{
+    std::string cn;
+    if (hash->getString("Category", cn) && cn.size())
```
375gnu: Remove `&& cn.size()`

375gnu:
Please don't push changes ssc files, add new put to `test` subdir instead.

PR #175, opened 2018-12-15
Celx_misc modernisation & some internal fixes.

pirogronian:


src/celestia/celx_misc.cpp
```diff
@@ -261,132 +241,109 @@ static int image_new(lua_State* l, Image* i)
...
 static int texture_tostring(lua_State* l)
 {
-    // TODO: print out the actual information about the texture
-    lua_pushstring(l, "[Texture]");
+    CelxLua celx(l);
 
-    return 1;
+    auto tex = *celx.this_tex;
+    std::string s = fmt::sprintf("[Texture:%dx%d]", tex->getWidth(), tex->getHeight());
+    return celx.push(s.data());
```
375gnu: s.c_str()

src/celestia/celx_misc.cpp
```diff
@@ -261,132 +241,109 @@ static int image_new(lua_State* l, Image* i)
...
-static Texture* this_texture(lua_State* l)
-{
-    Texture* texture = to_texture(l,1);
-    if (texture == nullptr)
-        Celx_DoError(l, "Bad texture object!");
+// ==================== Texture ============================================
 
-    return texture;
-}
+#define this_tex getThis<Texture*>()
```
375gnu: J U S T  N O  W A Y.

Preprocessor must not be used this way.

src/celestia/celx_misc.cpp
```diff
@@ -261,132 +241,109 @@ static int image_new(lua_State* l, Image* i)
 }
 #endif
 
-static Image* to_image(lua_State* l, int index)
-{
-    Image** image = static_cast<Image**>(lua_touserdata(l, index));
-
-    return image != nullptr ? *image : nullptr;
-}
+#define this_image getThis<Image*>()
```
375gnu: See below

src/celestia/celx_misc.cpp
pirogronian: What's wrong with `data()`? According to documentation, it's identical to `c_str()`.

src/celestia/celx_misc.cpp
pirogronian: Yes, I knoe, wrong style. But what is the best way of preprocessor use?

src/celestia/celx_misc.cpp
375gnu: when you see c_str() you understand that receiver wants char*, what does it want if you see data() - void*?
and Celestia uses c_str() exclusively. let's stick to one style

src/celestia/celx_misc.cpp
375gnu: in C++ it's better to avoid preprocessor. and definitely it shouldn't be used just to save some key strokes.

pirogronian: Qt programmers certainly didn't use this rule... :-)

PR #179, opened 2018-12-19
Lua method to bind custom translation domain.

pirogronian:


src/celestia/celx_celestia.cpp
```diff
@@ -2475,6 +2475,18 @@ static int celestia_getrootcategories(lua_State *l)
     return celx.pushIterable<UserCategory*>(set);
 }
 
+static int celestia_bindtextdomain(lua_State *l)
+{
+    CelxLua celx(l);
+    
+    const char *domain = celx.safeGetString(2, AllErrors, "First argument of celestia:bindtranslationdomain must be domain name string.");
```
375gnu: Please check domainname: "domainname must be a non-empty string."

and plz rename your commit to some concrete "add celestia:bindtextdomain method" or similar, as you don't have support for ssc & friends.

src/celestia/celx_celestia.cpp
```diff
@@ -2475,6 +2475,28 @@ static int celestia_getrootcategories(lua_State *l)
     return celx.pushIterable<UserCategory*>(set);
 }
 
+static int celestia_bindtextdomain(lua_State *l)
+{
+    CelxLua celx(l);
+
+    const char *domain = celx.safeGetString(2, AllErrors, "First argument of celestia:bindtranslationdomain must be domain name string.");
+    if (domain == nullptr)
```
375gnu: I suppose `domain == nullptr` is not enough, you should also check that domain != ""
`if (domain == nullptr || *domain == '\0')`

375gnu:
@pirogronian why did you merge *incorrect* code which even doesn't compile????

PR #181, opened 2018-12-21
Some celx module fixes and new method.

pirogronian:


pirogronian:
1. Ok, will change.
2. What are you talking about? safeGetNonEmptyString? I use it in somep laces, but not in this PR.
3. Method name, exposed to lua, is "bindtranslationdomain". I can change C++ function name as well, if you wish.

src/celestia/celx_celestia.cpp
```diff
@@ -2475,22 +2475,12 @@ static int celestia_getrootcategories(lua_State *l)
...
 {
     CelxLua celx(l);
 
-    const char *domain = celx.safeGetString(2, AllErrors, "First argument of celestia:bindtranslationdomain must be domain name string.");
-    if (domain == nullptr || *dir == '\0')
-    {
-        celx.doError("Domain name string must not be empty.");
-        return 0;
-    }
+    const char *domain = celx.safeGetNonEmptyString(2, AllErrors, "First argument of celestia:bindtranslationdomain must be domain name string.");
```
375gnu: you still don't check returned value

375gnu:
>     2. What are you talking about? safeGetNonEmptyString? I use it in somep laces, but not in this PR.

If you use than it's ok

>     3. Method name, exposed to lua, is "bindtranslationdomain". I can change C++ function name as well, if you wish.

Let's keep names in sync to avoid confusion.

src/celestia/celx_celestia.cpp
375gnu: if it really abort execution, then it's ok

PR #184, opened 2018-12-27
I18n support for user categories.

pirogronian:
Due to on-time usage of translation procedure, there is no need for storing any string. Support for catalogue files should be better. However, its structure is very inconvenient. Ideally, there should be option for global (one-per-file) translation domain definition.


src/celengine/category.cpp
```diff
@@ -1,11 +1,15 @@
 #include <iostream>
+#include <libintl.h>
```
375gnu: Please use `#include <celutil/util.h>` instead. It has some macro required for windows.

src/celengine/category.cpp
```diff
@@ -1,11 +1,15 @@
...
 #include <celutil/debug.h>
 #include <celengine/catentry.h>
 #include "category.h"
 
-UserCategory::UserCategory(const std::string &n, UserCategory *p) :
+UserCategory::UserCategory(const std::string &n, UserCategory *p, const char *domain) :
```
375gnu: std::string &domain

src/celengine/category.cpp
```diff
@@ -47,9 +51,9 @@ bool UserCategory::_removeObject(Selection s)
     return true;
 }
 
-UserCategory *UserCategory::createChild(const std::string &s)
+UserCategory *UserCategory::createChild(const std::string &s, const char *domain)
```
375gnu: std::string &domain

src/celengine/category.cpp
```diff
@@ -109,11 +113,11 @@ void UserCategory::cleanup()
 UserCategory::CategoryMap UserCategory::m_allcats;
 UserCategory::CategorySet UserCategory::m_roots;
 
-UserCategory *UserCategory::newCategory(const std::string &s, UserCategory *p)
+UserCategory *UserCategory::newCategory(const std::string &s, UserCategory *p, const char *d)
```
375gnu: std::string &domain (and NOT just *d* - d? what does it mean???)

src/celengine/category.cpp
```diff
@@ -122,9 +126,9 @@ UserCategory *UserCategory::newCategory(const std::string &s, UserCategory *p)
     return c;
 }
 
-UserCategory *UserCategory::createRoot(const std::string &n)
+UserCategory *UserCategory::createRoot(const std::string &n, const char *domain)
```
375gnu: and here

src/celengine/category.h
```diff
@@ -21,17 +21,18 @@ class UserCategory
 
 private:
     std::string m_name;
+    const char *m_i18n;
```
375gnu: std::string m_i18nname;

we won't have 100500 categories, so no need to optimize memory usage *here*

src/celengine/category.h
```diff
@@ -21,17 +21,18 @@ class UserCategory
...
     ObjectSet m_objlist;
     CategorySet m_catlist;
     UserCategory *m_parent;
 
-    UserCategory(const std::string&, UserCategory *parent = nullptr);
+    UserCategory(const std::string&, UserCategory *parent = nullptr, const char *domain = nullptr);
```
375gnu: string

src/celengine/category.h
```diff
@@ -21,17 +21,18 @@ class UserCategory
...
     ~UserCategory();
 public:
-    const std::string &name() const { return m_name; }
+    const char *name(bool i18n = false) const { return i18n ? m_i18n : m_name.c_str(); }
```
375gnu: strings

src/celengine/catentry.cpp
```diff
@@ -80,12 +80,16 @@ bool CatEntry::isInCategory(const std::string &s) const
 
 bool CatEntry::loadCategories(Hash *hash)
 {
+    const char *d = nullptr;
+    std::string td;
+    if (hash->getString("TrDomain", td) && !td.empty())
```
375gnu: TranslationFile ?

pirogronian: Rather not, as `gettext` doesn't provide ability to choose particular translation file - at most we can provide catalog. So do you mean `TranslationCatalog`, avoiding domain as proxy?

src/celengine/category.cpp
pirogronian: Because while earlier, single work I rarely named function parameters and there was nobody to tell me this is bad. Sorry, just forgot.

src/celengine/catentry.cpp
375gnu: I just mean that we need a more meaningful name than TrDomain. TranslationDomain or TranslationCatalog are both ok.

src/celengine/category.cpp
```diff
@@ -109,11 +113,11 @@ void UserCategory::cleanup()
 UserCategory::CategoryMap UserCategory::m_allcats;
 UserCategory::CategorySet UserCategory::m_roots;
 
-UserCategory *UserCategory::newCategory(const std::string &s, UserCategory *p)
+UserCategory *UserCategory::newCategory(const std::string &s, UserCategory *p, const std::string &d)
```
375gnu: still just d

src/celengine/category.h
```diff
@@ -21,17 +21,18 @@ class UserCategory
 
 private:
     std::string m_name;
+    std::string m_i18n;
     ObjectSet m_objlist;
     CategorySet m_catlist;
     UserCategory *m_parent;
 
-    UserCategory(const std::string&, UserCategory *parent = nullptr);
+    UserCategory(const std::string&, UserCategory *parent = nullptr, const std::string &domain = std::string());
```
375gnu: just `domain = ""`?

src/celengine/catentry.cpp
```diff
@@ -80,12 +81,23 @@ bool CatEntry::isInCategory(const std::string &s) const
 
 bool CatEntry::loadCategories(Hash *hash)
 {
+    const char *d = nullptr;
+    std::string td;
+    std::string tdir;
+    hash->getString("TranslationDirectory", tdir);
```
375gnu: how can you define a directory if you don't know where your addon will be put? you should use current directory instead.

src/celengine/category.cpp
pirogronian: Ah, here... I assumed it was in a header file. But if so, why `s` and `p` are ok?

src/celengine/catentry.cpp
pirogronian: I assumed it can be relative directory. Do you mean at every `TranslationDomain` new `bindtextdomain` should be call with current directory? But if so, why new domain should be defined at all? Why just not `UseLocalTranslation [boolean]` or so?

src/celengine/category.cpp
375gnu: not ok actually :) we should avoid cryptic names, but just to keep the pr as small as possible i said only about the new argument

src/celengine/catentry.cpp
375gnu: yes we don't need to define it, we can just assume that every foo_bar.ssc may have foo_bar.mo so just do `bindtextdomain("foo_bar", current_dir)`

pirogronian:
As there is no way to determine filename (or file path) from `istream` object, I've used additional directory parameter.

src/celestia/celx_category.cpp
```diff
@@ -25,11 +25,16 @@ static int category_getname(lua_State *l)
...
+    bool i18n = false;
     UserCategory *c = *celx.getThis<UserCategory*>();
-    const char *n = c->name().c_str();
+    if (celx.isBoolean(2))
+        i18n = celx.getBoolean(2);
+    const char *n = c->name(i18n).c_str();
     return celx.push(n);
 }
 
+#define D cout << __LINE__ << endl
```
375gnu: Remove this definition. Other things are ok,

PR #186, opened 2018-12-30
Add legacy OpenGL linking option

pirogronian:
Edition in root `CMakeLists.txt` fixed my issue with OpenGL detection.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,13 @@
 cmake_minimum_required(VERSION 3.1.0)
 
-if(POLICY CMP0072)
-  cmake_policy(SET CMP0072 NEW)
+set(LEGACY_OPENGL_LIBS OFF CACHE BOOL "Use legacy OpenGL libraries instead of glvnd library (Default: off)")
+
+if (POLICY CMP0072)
+  if (LEGACY_OPENGL_LIBS)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ if(ENABLE_THEORA)
   add_definitions(-DTHEORA)
 endif()
 
+set(OpenGL_GL_PREFERENCE LEGACY)
```
375gnu: CMake policy CMP0072 should be set properly instead. See first lines of the file. And it's better to make it configurable instead.

pirogronian: I'm newbie to cmake. How to make it configurable?

375gnu:
```
cmake_minimum_required(VERSION 3.1.0)

set(LEGACY_OPENGL_LIBS OFF CACHE BOOL "Use legacy OpenGL libraries instead of glvnd library (Default: off)")

if (POLICY CMP0072)
  if (LEGACY_OPENGL_LIBS)
    cmake_policy(SET CMP0072 OLD)
  else()
    cmake_policy(SET CMP0072 NEW)
  endif()
endif()

...
```

Then you can run `cmake -DLEGACY_OPENGL_LIBS=ON ..`

PR #188, opened 2019-01-03
Include data disposition while loading categories from catalog file.

pirogronian:


src/celengine/catentry.cpp
```diff
@@ -64,6 +64,19 @@ bool CatEntry::removeFromCategory(const std::string &s)
     return removeFromCategory(c);
 }
 
+bool CatEntry::clearFromCategories()
```
375gnu: is this a good name? maybe just `clearCategories`?

src/celengine/catentry.h
```diff
@@ -10,6 +10,13 @@ class UserCategory;
 class CatEntry
 {
 public:
+    enum DataDisposition
```
375gnu: `enum class DataDisposition` to enforce additional compiler's type safety checks

src/celengine/solarsys.cpp
```diff
@@ -39,15 +39,6 @@
 using namespace Eigen;
 using namespace std;
 
-
-enum Disposition
```
375gnu: Definitely no.

either:
```
location->loadCategories(objectData, static_cast<CatEntry::Dispositon>disposition, directory);
```
or
add a general `celengine/disposition.h` with `enum Disposition` and use it instead (instead of CatEntry::Dispositon too). In this case make 1st commit the change above and another commit with this chnage.

pirogronian: Even if we assume that CatEntry may state for both Category Entry and Catalog Entry (or we'll just rename it to something more general)?

375gnu: I don't get your question :) Nie rozumiem :)

pirogronian: Casting looks extremely ugly, while creating another header file looks a bit excessively. So, why not `CatEntry::DataDisposition`, if not because of its name and suggested exclusive function?

375gnu: Because CatEntry add unneeded and not fully correct dependency between different modules. If you don't like to add an new .h, you can check if it's possible to add the definition to some existing headers.

src/celengine/catentry.cpp
```diff
@@ -64,6 +65,19 @@ bool CatEntry::removeFromCategory(const std::string &s)
...
+{
+    bool ret = true;
+    while(m_cats != nullptr)
+    {
+        UserCategory *c = *(m_cats->begin());
+        bool ret2 = removeFromCategory(c);
+        if (!ret2)
+            ret = false;
```
375gnu:
```suggestion
        if (!removeFromCategory(c))
             ret = false;
```

src/celengine/catentry.cpp
```diff
@@ -64,6 +65,19 @@ bool CatEntry::removeFromCategory(const std::string &s)
     return removeFromCategory(c);
 }
 
+bool CatEntry::clearFromCategories()
+{
+    bool ret = true;
+    while(m_cats != nullptr)
+    {
+        UserCategory *c = *(m_cats->begin());
+        bool ret2 = removeFromCategory(c);
```
375gnu:
```suggestion
        if (!removeFromCategory(c))
             ret = false;
```

PR #207, opened 2019-01-18
Untemplatization of NameDatabase

pirogronian:


375gnu:
@pirogronian What's the problem with templates?

Issue #169, opened 2018-12-11
Bayer designations for globular clusters


LukeCEL:
Star names allow Bayer designations: e.g. "ALF Cen" becomes displayed as "α Cen". However, we should also have this feature for globular clusters, specifically for two globular clusters (xi Tucanae and omega Centauri) with Bayer designations.

pirogronian:
Fixed in #213 .

375gnu:
Not fixed yet

Issue #170, opened 2018-12-11
Differentiate Bayer and variable designations for MU, NU


LukeCEL:
Currently stars labeled "MU xxx" or "NU xxx" are displayed as Bayer designations, but it doesn't allow variable star designations. For example, HIP 33345 = μ Canis Majoris but HIP 35355 = MU Canis Majoris.

Based on the [Wikibook](https://en.wikibooks.org/wiki/Celestia/STC_File#Bayer_Star_Names), changing the code to only allow "MU." or "NU." is one option, however that wouldn't be backwards compatible.

LukeCEL:
Currently I don't have access to my Celestia installation so I can't provide any URLs.

Also, I should note that the default Celestia package doesn't include variable star designations. It's just that if you edit starnames.dat or add a new star with the name "MU CMa" or something, Celestia will assume that it's "μ CMa".

PR #213, opened 2019-01-19
Generalization of NameDatabase

pirogronian:
Reason: as Celestia has have ability to handle multiple data catalogs of many types of objects, there is a need for code generalization where possible. This should help keeping these code relatively simple and hence maintainable.


pirogronian:
However, separating DSO stars and bodies on a database level was always artificial and inconvenient to me. So, it had to be step towards eventual unification. I want tyo avoid situation like
`if (messing with stars)
   StarNameDatabase db = new StarNameDatabase;
else if (messing with dsos)
   DSONameDatabase db = new DSONameDatabase;`
The real difference is in fact a naming convention, physically unrelated to factual object nature. I wanted to go ahead situations with multiple object classification (gas giant planet/brown dwarf star or so). Compared to eventual code simplification, little loss in OOP seems to be a negligible disadvantage.
Maybe only method `readStars/loadStarNames` should be moved somewhere else, for example to StarDatabase, at leas for now.

375gnu:
Unfortunately, stars and DSOs naming nomenclatures are too different to unify them without additional `if (...) {...} else if (...) {...} else {...}` you want to avoid. Even for stars, we don't have nomenclature for variable stars which will add additional `if`s.

And in your change you still have `if (messing with stars) StarNameDatabase db = new StarNameDatabase; else if (messing with dsos) DSONameDatabase db = new DSONameDatabase;`
see https://github.com/CelestiaProject/Celestia/pull/213/files#diff-1669c4969693bad396c6ad4704051f88

pirogronian:
Even if there is no way to avoid `if else` blocks, I want to group them in few places or even one place rather, than let the code to be spread among many files and classess.
Secondary reason was elimination of some small files and additional classes. I find it desirable, if there is no method name and semantic collisions.
I wandered about use single instance of NameDatabase for both stars and dsos database, but I'm afraid of names and (specially) indexes collisions.

375gnu:
> Even if there is no way to avoid `if else` blocks, I want to group them in few places or even one place rather, than let the code to be spread among many files and classess.

We have only 1 such place in celestiacore.cpp in any case.

> BTW, #169 is example we may need star name convention also for some DSOs.

Good opportunity to move some helper code to the parent.

> Secondary reason was elimination of some small files and additional classes. I find it desirable, if there is no method name and semantic collisions.

Have you tried to recompile render.cpp? On my 2.1GHz 11 yo AMD it was a nightmare. Because it's terribly large. The smaller files the better.

> I wandered about use single instance of NameDatabase for both stars and dsos database, but I'm afraid of names and (specially) indexes collisions.

Soooo, no reason to implement them in a single class.

We have a lot of things to fix and this case is definitely not among them. Don't fix what isn't broken. I can share a list of Celestia's bugs (~50 items) with you :)

pirogronian:
Ok, technically both helpers are good for both stars and dsos. Also, I tried to fix #169 , but didn't test it yet.

> I can share a list of Celestia's bugs (~50 items) with you :)

Yes, please :)

375gnu:
On 1/21/19, pirogronian <notifications@github.com> wrote:

> Ok, technically both helpers are good for both stars and dsos. Also, I tried
> to fix #169 , but didn't test it yet.

Could take over a related issue #170?

>> I can share a list of Celestia's bugs (~50 items) with you :)
>
> Yes, please :)

Sent by email.

pirogronian:
#169 fixed.

375gnu:
That's a terrible workaround. Users won't be happy with it.

I suppose the better is to check a star type and not replace MU/NU if it's a variable one. @LukeCEL what do you think?

But this will work for display only. An additional fix is required for search console.

375gnu:
On 1/22/19, pirogronian <notifications@github.com> wrote:

> This time I am unsure. On one hand it would be more elegant from user point
> of view. However it would bind name semabtick to physical properties of
> object, something I want to avoid. Moreover, as letters replacement take
> place just before name displaying, displaying code would have to take
> object's type into account. We also don't know if there will be need for
> another changes in the future, when another naming convention arrive.
> On the other hand, I tried to "format string" approach, which maybe looks
> hacky but is naming-convention agnostick and faster for users.

Celestia exists not for itself. Unfortunately existing nomenclatures
are "tied" to physical properties. So such code will be always
"hackish".

>> But this will work for display only. An additional fix is required for
>> search console.
>
> Certainly not. I tested it on the search console. However, there is still a
> problem with differentiate both cases while searching on console. Maybe just
> let user to type extra '_', as in catalog file.

 I was talking about my proposed approach.

pirogronian:
So... As there is no way in Celestia to distinguish in `find` or `findFromPath` function between Greek letter notation and variable star notation, I propose to 
1. Store in database already formatted names and
2. Store in every console search complection entry both matching name and unique id (Selection or catalog number).
There are still some questions:
1. How to indiucate variable star with Bayer notation also (maybe name order in string?) and
2. How to indicate variable star itself (`SpectralType "Variable"` or `SpectralType { ... }`?)

375gnu:
https://en.wikibooks.org/wiki/Celestia/STC_File#Bayer_Star_Names - this link gives some useful info. So, if you want greek "μ" you should use "mu." in the console and stc files.

Basically I support the idea to keep names "formatted".

I'm not ready to answer 3 & 4, we should have conversation with people more aware in astronomy. @LukeCEL, @Askaniy

LukeCEL:
> To prevent replacement on Greek letter, just add '_' at the beginning of name. 

To be honest, I'm fine with the '_' approach. Currently to get Greek letters you capitalize them. This already feels pretty "hackish" to me. 

> I suppose the better is to check a star type and not replace MU/NU if it's a variable one. @LukeCEL what do you think?

How would Celestia check if a star is variable? Maybe we could add a new parameter called `VariableType`, but that would be its own separate issue.

Also, variable stars and designations do not have a one to one correspondence, i.e. not all variable stars are given variable star designations. For example, μ Gem ([HIP 30343](http://simbad.u-strasbg.fr/simbad/sim-id?Ident=%40885951&Name=*%20mu.%20Gem&submit=submit)) is a long-period variable star. If the name formatting was intrinsically tied to the variable type, then this star would be formatted as "MU Gem".

As for 3 and 4, I'm also not ready to answer these. I think we should discuss this on the Celestia Forums.

LukeCEL:
> https://en.wikibooks.org/wiki/Celestia/STC_File#Bayer_Star_Names - this link gives some useful info. So, if you want greek "μ" you should use "mu." in the console and stc files.

I basically agree. I should point out that most people in the past have used "MU" for "μ" in the past.

> Basically I support the idea to keep names "formatted".

Can you clarify what this means?

375gnu:
On 1/23/19, LukeCEL <notifications@github.com> wrote:

>> Basically I support the idea to keep names "formatted".
> Can you clarify what this means?

So, currently we keep in memory names with abbreviations and now we
think that we should keep them with the abbreviations replaced with
greek letters.

pirogronian:
Latest force-pusged commit enable formatted names in database. However, it broke console search complection. `UTF8StringCompare` seems to not work with strings with Greek letters. Maybe it's time to use more complex solution for unicode support?

375gnu:
I'd start with https://en.wikibooks.org/wiki/Celestia/STC_File#Bayer_Star_Names for 1.7.

If you want something more complex, then start with textual description and unit tests. I always thought about necessity of unit test in Celestia, and that's a good place to start with.

pirogronian:
Well, my recent changes are very provisional and experimental. Tests seem to pass, but latest commit brings very unexpected results. Any ideas about possible bug location or more useful tests?

375gnu:
You have provided very basic tests. Unit testing is a slightly different thing. So, if you have `Foo` class you make tests to ensure that its methods behave as expected and return expected values. Then you add tests to ensure that things which use your `Foo` work as expected. So in this case I'd add test to simulate search and completion.

pirogronian:
Changes are still a bit chaotic, but complection for DSO & stars seems to work well.

pirogronian:
@375gnu what about this pr? Some serious changes needed or only polishing?

pirogronian:
Tinyutf8 was thought to provide more comprehensive UTF8 support. I also worried that UTF8StringCompare gives bad results (although I'm still not sure about that).

375gnu:
I don't think that tinyutf8 is required just for comparisons. We might want to use it or similar library later when we have a use case for it.

And tinyutf8 is not C++11 compatible.

pirogronian:
@375gnu I removed `tinyutf8` dependency.

375gnu:
Hi @pirogronian, could you split the last 4 commits into their own branch? I suppose that after testing and fixing some remarks we can merge them. But I still not happy with the first 3 of them. So let's not merge them (at least for now).

pirogronian:
I'm not sure how to do that (still not git master) but I'll try.

src/celutil/utf8.cpp
```diff
@@ -577,9 +577,9 @@ int UTF8StringCompare(const std::string& s0, const std::string& s1, size_t n)
         ch0 = UTF8Normalize(ch0);
         ch1 = UTF8Normalize(ch1);
 
-        if (ch0 < ch1)
+        if (ignoreCase ? std::tolower(ch0) < std::tolower(ch1) : ch0 < ch1)
             return -1;
-        if (ch0 > ch1)
+        if (ignoreCase ? std::tolower(ch0) > std::tolower(ch1) : ch0 > ch1)
```
375gnu: you call tolower twice here

it's better to 
```
if (ignoreCase)
{
    ch0 = std::tolower(ch0);
    ch1 = std::tolower(ch1);
}
```

src/celutil/utf8.cpp
```diff
@@ -897,3 +897,93 @@ ReplaceGreekLetterAbbr(char *dst, unsigned int dstSize, const char* src, unsigne
...
+        int i = 0;
+        while(i >= 0)
+        {
+            std::string rets;
+            i = findGreekNameIndexBySubstr(s, i);
+            if (i >= 0)
+            {
+                rets = Greek::getInstance()->abbrevs[i];
+                rets += " ";
+                ret.push_back(ReplaceGreekLetterAbbr(rets));
```
375gnu: .emplace_back should be better here

src/celutil/utf8.cpp
```diff
@@ -897,3 +897,93 @@ ReplaceGreekLetterAbbr(char *dst, unsigned int dstSize, const char* src, unsigne
...
+    else
+    {
+        std::string prefix = s.substr(0, sp);
+        std::string suffix = s.substr(sp + 1);
+        std::string rets = Greek::canonicalAbbreviation(prefix);
+        if (!rets.empty())
+        {
+            rets += " ";
+            rets += suffix;
+            ret.push_back(ReplaceGreekLetterAbbr(rets));
```
375gnu: and here

PR #215, opened 2019-01-20
Celx LUA value to Celestia parser Value convertion helper.

pirogronian:


src/celestia/celx.cpp
```diff
@@ -1798,6 +1798,61 @@ bool CelxLua::isType(int index, int type) const
...
+                {
+                    delete hash;
+                    hash = nullptr;
+                }
+                if (array != nullptr)
+                {
+                    array->push_back(getValue(-1));
+                }
+            }
+            if (isString(-2))
```
375gnu: `else`

src/celestia/celx.cpp
```diff
@@ -1798,6 +1798,61 @@ bool CelxLua::isType(int index, int type) const
...
+                    array = nullptr;
+                }
+                if (hash != nullptr)
+                {
+                    hash->addValue(getString(-2), *getValue(-1));
+                }
+            }
+            pop(1);
+            if (array == nullptr &&
+                hash == nullptr)
```
375gnu: don't break line

src/celestia/celx.cpp
```diff
@@ -1798,6 +1798,61 @@ bool CelxLua::isType(int index, int type) const
...
+                break;
+        }
+        pop(1);
+        if (hash != nullptr)
+            v = new Value(hash);
+        if (array != nullptr)
```
375gnu: `else`

pirogronian:
As LUA don't separate arrays and hash tables, I check keys integrity. If table has only integer keys, then become an Array. Ih has only string keys, become an Hash. Otherwise table is invalid and function returns null pointer.

Issue #219, opened 2019-01-29
Can't compile in Ubuntu 18.10- Cannot find file: celestia.pro.


wrapperband:
Can't compile in Ubuntu 18.10 - Cannot find file: celestia.pro.

I am following the instructions in the README as instructed, but there is no celestia.pro file

qmake -qt=5 PREFIX=/opt/celestia -o Makefile celestia.pro

Cannot find file: celestia.pro.

pirogronian:
Qmake support was removed just days ago. Please use cmake instead.

wrapperband:
Cheers, what is the new instructions?

wrapperband:
Ok :  it's in commit. I'll try this  : 

after cd build

cmake ..
- make -j $(nproc || echo 4)

375gnu:
I'll update README.txt & INSTALL.txt, thanks for pointing out

pirogronian:
@wrapperband You probably want `cmake .`, with one dot. Moreover, if Celestia wont find opengl, then try add `-DLEGACY_OPENGL_LIBS`.

wrapperband:
It seems to have worked so far : I'll test the build now.

wrapperband:
Ok, it lied - and no celestia-qt was created - I'll try cmake with one dot.

wrapperband:
Didn't compile this time, with cmake .

-- Could NOT find Gettext (missing: GETTEXT_MSGMERGE_EXECUTABLE GETTEXT_MSGFMT_EXECUTABLE) 
CMake Warning at CMakeLists.txt:81 (message):
  Gettext is not found, traslations won't be created.

Found Eigen3 3.3.4
Using bundled libfmt
-- Could NOT find GLUT (missing: GLUT_glut_LIBRARY GLUT_INCLUDE_DIR) 
CMake Warning at src/celestia/CMakeLists.txt:89 (message):
  GLUT library isn't found, not building GLUT fronend.

Either not Unix or Gtk frontend is disabled.
Either not Windows or Windows frontend is disabled.

375gnu:
did you run `make` after `cmake`? Cmake is like qmake, it only generates makefiles, but doesn't build anything.

wrapperband:
Ok :   Yes I did Make : trying to find where it put the file supposedly 

Celestia/build/src/celestia/qt

pirogronian:
I forgot - after cmake run `make`. These messages are warnings only. Btw could you install gettext and glut on your system?

[...]

PR #220, opened 2019-01-29
Update README and INSTALL (Closes: #219)

375gnu:
Update of README file is too large as it had DOS EOL.


375gnu:
I have removed all dead links

PR #222, opened 2019-02-02
fix extra-standard to copy directory hierarchy

joequant:
The cmakefiles install everything into the base directory which means that the models don't appear.


375gnu:
#227 is a better implementation

Issue #154, opened 2018-12-06
Support for additional star catalogues


375gnu:
For example Kepler & Gaia

pirogronian:
I think about exposing unified data manipulation Lua api, so incomporating any type of data would be a matter of writing proper lua script(s). I'm impressed by game [Pioneer](https://pioneerspacesim.net). It has nearly all game mechanics implemented in lua. Core implements only data model, rendering and some other important utilities. This approach seems to be the best choice for me. In Celestia, there could be table [Catalogue name] => object name/nr, however, it can also be implemented as user defined properties.

375gnu:
Currently a lot of thing are implemented in Celestia core. See https://github.com/CelestiaProject/Celestia/blob/master/src/celengine/stardb.cpp#L33. Unfortunately different star catalogs have different naming scheme, so it's no trivial to move this into config file.

pirogronian:
So, doing shortly, we can just implement additional `parse<CatalogName>CatalogNumber` function to map particular catalog identifier to internal Celestia index and extend `catalogNumberToString`. However, code of database is so ugly, that I would be happy making it cleaner. However, I'm not sure how exactly this code work. Mayby we should start to make inline documentation of code first.

PR #247, opened 2019-03-24
[WIP] Unified astro catalogs database

pirogronian:
Objectives:
* make easier adding new astro catalogs (should will solve #154 ) (being also general, without separate instantiations for stars, DSOs and planets) [done]
* make dynamic objects loading more flexible
* make step towards dynamic objects unloading
Discuss subjects:
* nature of `AstroDatabase::Entry` (extension of `Selection` or common base as with `CatEntry`?)[1]
* or is this a good opportunity to modify `Selection` class and merge it with `CatEntry`?[1]

[1] As suggested by @375gnu , we created `AstroObject` class which is merge of `AstroDatabase::Entry` and `CatEntry`. `CatEntry` as separate class is hereby removed.

28.03.2019: As first data loader is implemented, we are ready for first data loading tests.
29.03.2019: As 2622774 commit has came, all tested data loads without errors.
23.04.2019: I decided to close first phase of this branch. Having moderately tested database for stars and dsos, I decided to integrate it with rest of the code. To get it work properly I had to adapt an octree first. Now I'm in the middle of an integration process and I'm looking for compilation errors (or, rather, lots of compilation errors is looking for me). Up to now celengine is integrated. When everything will compile and  build, I will ready to see how everything blows up during runtime... ;-)
26.04.2019: Good news: it works! Bad news - it's slow. 30 fps instead of 60... Probably it's due to different built octree.
30.04.2019: Octree rewritten. Performance greatly improved, but still worse than with original implementation.
New bug: Blinking and disappearing star, where rotating around its planet.
Old bug: DSOs still not present.
~15.06.2019: Fixed objects blining by small hack, fixed DSO display.
06.06.2019: Added new names handling, intended to enable incorporation of bodies and locations. Additional optimisation by using shared pointers to name strings. Octree tuning greatly improved perfomance, now FPS is comparable with master branch. Due to big amount of changes, tests are welcome.
~25.07.2019: Bodies and SolarSystem classes and surrounding infrastructure is integrated with AstroDatabase. Formally it fulfills goals of this branch, but another goal - make objects dynamically unloadable, seems to be at fingertips.


375gnu:
Please add Gaia and Kepler catalog names.

Not sure about merging Selection with CatEntry and AstroDatabase::Entry because the change will too intrusive and Celestia doesn't have any unit tests to catch possible regressions. So I'd suggest merging CatEntry into AstroDatabase::Entry as a 1st step and only then when this change is merged and tested start merging Selection with  AstroDatabase::Entry.

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+AstroCatalog::IndexNumber SimpleAstroCatalog::getIndexNumberByName(const std::string &name)
+{
+    if (compareIgnoringCase(name, m_prefix, m_prefix.length()) == 0)
+    {
+        IndexNumber num;
+        // Use scanf to see if we have a valid catalog number; it must be
+        // of the form: <prefix> <non-negative integer>  No additional
+        // characters other than whitespace are allowed after the number.
+        char extra[4];
+        if (std::sscanf(name.c_str() + m_prefix.length(), " %u %c", &num, extra) == 1)
```
375gnu: what if you have 2 spaces?

375gnu: space before %u is not required

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
+
+#include <cstdio>
+
+#include "astrocat.h" 
+
+AstroCatalog::IndexNumber SimpleAstroCatalog::getIndexNumberByName(const std::string &name)
+{
+    if (compareIgnoringCase(name, m_prefix, m_prefix.length()) == 0)
+    {
+        IndexNumber num;
```
375gnu: better use `unsigned int` and `static_cast` in return

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+        {
+            return num;
+        }
+    }
+    return InvalidIndex;
+}
+
+std::string SimpleAstroCatalog::getNameByIndexNumber(IndexNumber index)
+{
+    return fmt::sprintf("%s %d", m_prefix, index);
```
375gnu: return fmt::sprintf("%s %u", m_prefix, static_cast<unsigned>(index));

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const std::string& SimpleAstroCatalog::getPrefix() const
```
375gnu: inline

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+    return m_prefix;
+}
+
+AstroCatalog::IndexNumber TychoAstroCatalog::getIndexNumberByName(const std::string& name)
+{
+    int len = m_prefix.size();
+    if (compareIgnoringCase(name, m_prefix, len) == 0)
+    {
+        unsigned int tyc1 = 0, tyc2 = 0, tyc3 = 0;
+        if (std::sscanf(std::string(name, len, std::string::npos).c_str(),
```
375gnu: better use string::substr here

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+                   " %u-%u-%u", &tyc1, &tyc2, &tyc3) == 3)
```
375gnu: space is unrequired

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+std::string TychoAstroCatalog::getNameByIndexNumber(IndexNumber index)
+{
+    uint32_t tyc3 = index / 1000000000;
+    index -= tyc3 * 1000000000;
+    uint32_t tyc2 = index / 10000;
+    index -= tyc2 * 10000;
+    uint32_t tyc1 = index;
+    return fmt::sprintf("%s %d-%d-%d", m_prefix, tyc1, tyc2, tyc3);
```
375gnu: d -> u

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+AstroCatalog::IndexNumber CelestiaAstroCatalog::getIndexNumberByName(const std::string &name)
+{
+    char extra[4];
+    IndexNumber index = SimpleAstroCatalog::getIndexNumberByName(name);
+    if (index != InvalidIndex)
+        return index;
+    if (std::sscanf(name.c_str(), "#%u %c", &index, extra) == 1)
```
375gnu: like above

src/celengine/astrocat.h
```diff
@@ -0,0 +1,68 @@
...
+#pragma once
+
+#include <climits>
+#include <string>
+#include <celutil/util.h>
+#include <fmt/printf.h>
+
+class AstroCatalog
+{
+public:
```
375gnu: one space before `public/private`

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,157 @@
...
+    return catalogNumberToString(nr);
+}
+
+std::string AstroDatabase::getObjectNameList(AstroCatalog::IndexNumber nr, int max) const
+{
+    string names;
+    NameDatabase::NumberIndex::const_iterator iter = m_nameDB.getFirstNameIter(nr);
+    while (iter != m_nameDB.getFinalNameIter() && iter->first == nr && max > 0)
+    {
+        if (names.size() > 0)
```
375gnu: !names.empty()

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,157 @@
...
+        return "#{Invalid catalog}";
+    return m_catalogs.find(catalog)->second->getNameByIndexNumber(nr);
+}
+
+std::string AstroDatabase::getObjectName(AstroCatalog::IndexNumber nr, bool i18n) const
+{
+    NameDatabase::NumberIndex::const_iterator iter = m_nameDB.getFirstNameIter(nr);
+    if (iter != m_nameDB.getFinalNameIter() && iter->first == nr)
+    {
+        if (i18n && iter->second != _(iter->second.c_str()))
```
375gnu:
```if (i18n)
{
    const char* localized = _(iter->second.c_str());
    if (iter->second != localized)
        return localized;
}
return iter->second;
```

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,157 @@
...
+}
+
+std::string AstroDatabase::catalogNumberToString(AstroCatalog::IndexNumber nr) const
+{
+    return m_catalogs.find(Celestia)->second->getNameByIndexNumber(nr);
+}
+
+std::string AstroDatabase::catalogNumberToString(int catalog, AstroCatalog::IndexNumber nr) const
+{
+    if (m_catalogs.count(catalog) == 0)
```
375gnu:
```
auto it = m_catalogs.find(catalog);
if (it != m_catalogs.end())
    return it->second->getNameByIndexNumber(nr);
return "#{Invalid catalog}";
```

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,157 @@
...
+{
+    AstroCatalog::IndexNumber nr = m_nameDB.getCatalogNumberByName(name);
+    if (nr != AstroCatalog::InvalidIndex)
+        return nr;
+    for (const auto ci : m_catalogs)
+    {
+        AstroCatalog::IndexNumber inr = ci.second->getIndexNumberByName(name);
+        if (inr == AstroCatalog::InvalidIndex)
+            continue;
+        if (ci.first == Celestia)
```
375gnu: better use `?:` here

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,157 @@
...
+            return _(iter->second.c_str());
+        else
+            return iter->second;
+    }
+    return catalogNumberToString(nr);
+}
+
+std::string AstroDatabase::getObjectNameList(AstroCatalog::IndexNumber nr, int max) const
+{
+    string names;
```
375gnu: to optimize memory allocations use `string names("", N)` where N is large enough to keep full name, i suppose 127 will be ok. or `string names; names.reserve(N);`

src/celengine/astrodb.h
```diff
@@ -0,0 +1,60 @@
...
+#include "astroobj.h"
+#include "astrocat.h"
+
+class AstroDatabase {
+public:
+    
+    typedef std::unordered_map<AstroCatalog::IndexNumber, AstroCatalog::IndexNumber> CrossIndex;
+    typedef std::unordered_map<AstroCatalog::IndexNumber, AstroObject*> MainIndex;
+    
+    enum Catalog
```
375gnu: `enum class Catalog`

src/celengine/astrocat.cpp
pirogronian: This code is mostly copied from `stardb.cpp`. I don't know how to handle mutliple spaces. The same for others cases with `sscanf`.

src/celengine/astrocat.h
pirogronian: Should or shouldn't be? And space for sure, not new line, for example?

src/celengine/astrodb.h
pirogronian: Ok, however I don't understand a difference...

375gnu: `class enum` is safe variant of `enum`, for example, this code will not compile:
```
class enum Foo {A, B, C};
void func(Foo);
...
func(1);
```

pirogronian: Ok

src/celengine/astrocat.h
375gnu: One space should be before any access specifier, see coding-standards.html

src/celengine/astrocat.cpp
375gnu: I incorrectly read the spec, I have thought that whitespace can be input for %c

src/celengine/astrodb.h
pirogronian: There is a problem with it. Adding `class` prevent usage of enumeration as array index (which break `static const char *CatalogPrefix[MaxBuiltinCatalog];` for example).

pirogronian:
First basic test passed. Any ideas and/or contributions for more sophisticated unit tests are welcome.

src/celengine/stardataloader.cpp
```diff
@@ -0,0 +1,289 @@
...
+/*        if (isNewStar)
+            star = new Star();*/
+
+        if (ok)
+        {
+//            clog << " About to create star with nr " << star->getMainIndexNumber() << endl;
+            ok = Star::createStar(star, disposition, starData, resourcePath, !isStar, m_db);
+            if (star->getDetails() == nullptr)
+//                 clog << " Created star has null details!\n";
+            star->loadCategories(starData, disposition, resourcePath);
```
375gnu: guess what's wrong with these lines

src/celengine/stardataloader.cpp
```diff
@@ -0,0 +1,289 @@
...
+        } //else clog << " Errors while preparing star, skipping creation.\n";
+        delete starDataValue;
+
+        if (ok)
+        {
+
```
375gnu: unneeded extra new

src/celengine/stardataloader.cpp
```diff
@@ -0,0 +1,289 @@
...
+                    if (next != string::npos)
+                    {
+                        length = next - startPos;
+                        ++next;
+                    }
+                    string starName = objName.substr(startPos, length);
+                    m_db->addName(catalogNumber, starName);
+                    clog << " Adding name \"" << starName << "\" for entry nr " << catalogNumber << endl;
+                    if (starName != _(starName.c_str()))
+                        m_db->addName(catalogNumber, _(starName.c_str()));
```
375gnu: 2 calls to `_()`

src/celengine/dataloader.cpp
```diff
@@ -0,0 +1,25 @@
...
+    {
+        if (getSupportedContentType() != Content_Unknown && DetermineFileType(fname) != getSupportedContentType())
+        {
+            clog << "Error while loading content from \"" << fname << "\": wrong file content type.\n";
+            return false;
+        }
+    }
+
+    fstream stream(fname, ios::in);
+    if (!stream.good()) {
```
375gnu: move { to the next line

src/celengine/stardataloader.cpp
pirogronian: Indeed :-)

src/celengine/stardataloader.cpp
pirogronian: New line, I presume.

src/celengine/astrodb.cpp
pirogronian: With current implementation, such a memory preallocation leads to string with junks at the beginning and normal content after preallocated section.

375gnu:
On 3/30/19, Łukasz Buczyński <notifications@github.com> wrote:

> With current implementation, such a memory preallocation leads to string
> with junks at the beginning and normal content after preallocated section.

yes, it seems that the proper one is
```
std::string s;
s.reserve(100);
```

src/celengine/astrocat.h
```diff
@@ -0,0 +1,69 @@
...
+
+class TychoAstroCatalog : public SimpleAstroCatalog
+{
+ public:
+    TychoAstroCatalog() : SimpleAstroCatalog("TYC") {}
+    virtual IndexNumber getIndexNumberByName(const std::string&);
+    virtual std::string getNameByIndexNumber(IndexNumber);
+};
+
+class CelestiaAstroCatalog : public SimpleAstroCatalog
```
375gnu: What the purpose of this class?

src/celengine/astrocat.h
```diff
@@ -0,0 +1,69 @@
...
+    virtual const std::string& getName() const = 0;
+};
+
+class SimpleAstroCatalog : public AstroCatalog
+{
+ protected:
+    std::string m_prefix;
+ public:
+    SimpleAstroCatalog(const std::string& pref) { m_prefix = pref; }
+    virtual IndexNumber getIndexNumberByName(const std::string&);
```
375gnu: please prefer C++11 syntax for overridden methods:
```
IndexNumber getIndexNumberByName(const std::string&) overrride;
```

With it you don't introduce new virtual functions by mistake.

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,227 @@
+
+#include "astrodb.h"
+
+const char *AstroDatabase::CatalogPrefix[AstroDatabase::MaxBuiltinCatalog] = { "CEL", "HD", "Gliese", "SAO", "HIP", "TYC" };
```
375gnu:
```
constexpr array<const char *, AstroDatabase::MaxBuiltinCatalog> = {...};
```

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+AstroCatalog::IndexNumber AstroDatabase::findCatalogNumberByMainIndex(int catalog, AstroCatalog::IndexNumber nr) const
+{
+    std::unordered_map<int, CrossIndex*>::const_iterator it = m_celxindex.find(catalog);
+    if (it == m_celxindex.end())
+        return AstroCatalog::InvalidIndex;
+    if (it->second->count(nr) > 0)
+        return it->second->at(nr);
+    return AstroCatalog::InvalidIndex;
```
375gnu: as above

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+bool AstroDatabase::isInCrossIndex(int catalog, AstroCatalog::IndexNumber nr) const
+{
+    std::unordered_map<int, CrossIndex*>::const_iterator it = m_catxindex.find(catalog);
+    if (it == m_catxindex.end())
+        return false;
+    if (it->second->count(nr) == 0)
+        return false;
+    return true;
```
375gnu: as above

src/celengine/astrocat.h
pirogronian: Hm, I added it by convention, but in fact it's indeed unnecessary.

src/celengine/astroobj.h
```diff
@@ -1,14 +1,24 @@
...
 class UserCategory;
 
-class CatEntry
-{
+class AstroObject {
+    AstroCatalog::IndexNumber m_mainIndexNumber { AstroCatalog::InvalidIndex };
+    AstroDatabase *m_db { nullptr };
+public:
+    AstroObject() = default;
+    AstroObject(AstroDatabase *db, AstroCatalog::IndexNumber nr = AstroCatalog::InvalidIndex) { m_db = db; m_mainIndexNumber = nr; }
```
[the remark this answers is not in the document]
pirogronian: What is a proper syntax in this case? I'm not C++11 expert...

src/celengine/star.cpp
```diff
@@ -1140,6 +1135,8 @@ void Star::setLuminosity(float lum)
 
 StarDetails* Star::getDetails() const
 {
+    if (details == nullptr)
```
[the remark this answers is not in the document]
pirogronian: Technically it's not true, but indeed I added it for debug only. But it debugged indeed...

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,67 @@
...
+AstroCatalog::IndexNumber TychoAstroCatalog::getIndexNumberByName(const std::string& name)
+{
+    int len = m_prefix.size();
+    if (compareIgnoringCase(name, m_prefix, len) == 0)
+    {
+        unsigned int tyc1 = 0, tyc2 = 0, tyc3 = 0;
+        if (std::sscanf(name.substr(len, std::string::npos).c_str(),
+                   " %u-%u-%u", &tyc1, &tyc2, &tyc3) == 3)
+        {
+            return (tyc3 * 1000000000 + tyc2 * 10000 + tyc1);
```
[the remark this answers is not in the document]
pirogronian: I don't know nuances of Tycho catalog naming rules. I just copied it from legacy code. What should be changed here?

src/celengine/astroobj.h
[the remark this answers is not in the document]
375gnu:
```
AstroObject(AstroDatabase *db, AstroCatalog::IndexNumber nr = AstroCatalog::InvalidIndex) : m_db(db), m_mainIndexNumber(nr) {}
```

src/celengine/star.cpp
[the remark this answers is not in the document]
375gnu: maybe clog is unbuffered so you see the message.

src/celengine/astrocat.cpp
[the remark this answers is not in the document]
375gnu: some pseudo code: tycho name -> tycho catalog number -> through crossindex get a main index.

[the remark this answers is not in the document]
pirogronian: I know it, but it's generating unique catalog number what is my problem. Tycho name contains 3 numbers. Current code quarantee unique catalog number for any Tycho name. Is there better algorithm You suggest? I red [Tycho-2 format specification](http://tdc-www.harvard.edu/catalogs/tycho2.format.html) but don't understand so much.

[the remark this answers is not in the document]
375gnu: Okay, it seams that the routine is correct. But its name is quite unclear, I assumed that it returns internal Celestia's index number, not a catalog's number.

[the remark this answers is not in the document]
pirogronian: Indeed. At the beginning I couldn't decide which nomenclature to use...

src/celengine/dsodb.cpp
```diff
@@ -234,7 +234,7 @@ bool DSODatabase::load(istream& in, const string& resourcePath)
         objType = tokenizer.getNameValue();
 
         bool   autoGenCatalogNumber   = true;
-        uint32_t objCatalogNumber       = DeepSkyObject::InvalidCatalogNumber;
+        uint32_t objCatalogNumber       = AstroCatalog::InvalidIndex;
```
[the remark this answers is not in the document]
pirogronian: There is a lots of spaces left by previous coders...

pirogronian:
I see lots of isues found by Codacy, around code copied from legacy files. I'm not sure, how to write it better. Is it critical to get rid of it?
I see also fails of Travis and AppVeyor, caused by in-class constants declarations. I'll try to fix that.
One more issue. I have no skills with GTK and no chance to even test Mac and Win frontends. It obviously should fail to build with my changes.

375gnu:
> I see lots of isues found by Codacy, around code copied from legacy files. I'm not sure, how to write it better. Is it critical to get rid of it?

For me they sound like false positive. We can ignore them for now.

> One more issue. I have no skills with GTK and no chance to even test Mac and Win frontends. It obviously should fail to build with my changes.

I'll take care of them.

src/celengine/astrocat.cpp
```diff
@@ -25,18 +25,18 @@ const std::string& SimpleAstroCatalog::getName() const
...
+AstroCatalog::IndexNumber TychoAstroCatalog::nameToCatalogNumber(const std::string& name)
 {
-    int len = m_prefix.size();
+    int len = m_prefix.length();
     if (compareIgnoringCase(name, m_prefix, len) == 0)
     {
         unsigned int tyc1 = 0, tyc2 = 0, tyc3 = 0;
-        if (std::sscanf(name.substr(len, std::string::npos).c_str(),
+        if (std::sscanf(name.c_str(),
                    " %u-%u-%u", &tyc1, &tyc2, &tyc3) == 3)
```
375gnu: either missing `%*s` in the format string or `+ len` after name.c_str()

src/celengine/selection.cpp
```diff
@@ -142,7 +142,7 @@ string Selection::getName(bool i18n) const
                     if (parentStar != nullptr)
                     {
                         string buf;
-                        buf = fmt::sprintf("#%d", parentStar->getMainIndexNumber());
+                        buf = fmt::sprintf("#%d", parentStar->getIndex());
```
375gnu: `buf = fmt::sprintf("#%d/%s", parentStar->getIndex(), name);` and remove the line below this one

src/celestia/qt/qtcelestialbrowser.cpp
```diff
@@ -188,7 +188,7 @@ QVariant StarTableModel::data(const QModelIndex& index, int role) const
         {
         case NameColumn:
             {
-                uint32_t hipCatNo = star->getMainIndexNumber();
+                uint32_t hipCatNo = star->getIndex();
```
375gnu: *Just a marker for a comment regarding crossindexes here*

src/celengine/astrodb.cpp
```diff
@@ -11,6 +11,35 @@ AstroObject *AstroDatabase::getObject(AstroCatalog::IndexNumber nr) const
...
 
+AstroObject *AstroDatabase::getObject(const std::string &name) const
+{
+    return getObject(nameToIndex(name));
+}
+
+Star *AstroDatabase::getStar(AstroCatalog::IndexNumber nr) const
+{
+    Star *star = static_cast<Star*>(getObject(nr));
+    if (m_stars.count(star) > 0) return star;
```
375gnu: good place for ?: :)

src/celengine/astrodb.cpp
```diff
@@ -11,6 +11,35 @@ AstroObject *AstroDatabase::getObject(AstroCatalog::IndexNumber nr) const
...
+
+Star *AstroDatabase::getStar(const std::string &name) const
+{
+    return getStar(nameToIndex(name));
+}
+
+DeepSkyObject *AstroDatabase::getDSO(AstroCatalog::IndexNumber nr) const
+{
+    DeepSkyObject *dso = static_cast<DeepSkyObject*>(getObject(nr));
+    if (m_dsos.count(dso) > 0) return dso;
```
375gnu: And for ?: here

test/astrodb.cpp
```diff
@@ -19,7 +19,19 @@ bool objectNames(AstroDatabase &db, AstroCatalog::IndexNumber nr)
...
+    cout << "Names of object nr " << nr << ": " << db.getObjectNameList(nr, 1024) << endl;
+    return true;
+}
+
+bool objectNames(AstroDatabase &db, const std::string &name)
+{
+    AstroObject *o = db.getObject(name);
+    if (o == nullptr)
+    {
+        cerr << "Object named " << name << " doesn't exists!\n";
```
375gnu: fmt::fprintf

375gnu:
Currently your code still assumes HIP as the default catalog

pirogronian:
Hm, do You mean so called main index of AstroObject? I implement it for backward compatibility. However, we can assume, that any provided catalog number in stc file (or in names.dat and crossindex.dat) is valid a Hipparcos number and can be add to internal cross index. Then automatic index number can be dropped.

375gnu:
> Hm, do You mean so called main index of AstroObject? I implement it for backward compatibility.

Maybe I don't understand the idea of your changes.

I assumed that we'll have the following: stars.dat is the main catalog file, it has stars from different catalogs, not only HIP or TYC, but also from Gaia for example. stars.dat can be considered as big array of stars, so stars.dat index is Celestia internal index and can be changed on any stars.dat recompilation. So I assumed that the mainIndex is that internal index. While catalogNumber is a number from a crossindex file (an actual index from a particular catalog).

Just to note an idea for the further improvement:

Currently we don't have crossindex sources, but we should extend our tools to generate them from a single stars.txt. For example:
Now we have line:
12  0.040917099 -35.960224856 1003.590769 8.43 K4III

It should become
0.040917099,-35.960224856,1003.590769,8.43,K4III,HIP 12,HD 123,GAIA 234,TYC 12-3467-1
(numbers are fictional)

> However, we can assume, that any provided catalog number in stc file (or in names.dat and crossindex.dat) is valid a Hipparcos number and can be add to internal cross index. Then automatic index number can be dropped.

Yes  we need to assume this for a long time for backward compatibility.

pirogronian:
Your assumptions are correct. Main/internal index persists only for compatibility with old code. I also forgot to populate Hipparcos cross index while loading... However, if I could rewrite all rest code accessing this internal index, it can be eliminated and cross indexes would point directly pnto objects... 
But wait, in cross index data files there should be some kind of id referencing objects... Or maybe only pure Catalog to catalog mapping?

375gnu:
I see this as crossindex remain similar to that what we have now:
```
internalIndex1 catalogIndex1
internalIndex2 catalogIndex2
```

Now internalIndexes are HIP catalog numbers and catalogIndexes are numbers from other catalogs.

In my idea internalIndexes become pure celestia-only indexes and for any catalog-to-catalog mappings an intermediate step of getting internalIndex (HIP->internal->TYC/Gaia/HD/etc, Gaia->internal->HIP/Kepler/etc), that's not perfect but there is no 1-to-1 mapping for different catalogs for obvious reasons.

So we should still have internal main indexes.

pirogronian:
However in this case we wold have to be able bind internal index to every hosted object. If we do it automatically, then we would have no guarantion that particular object will have the same id between application sessions. So, will we do it manually for all existing objects? Or by external tool? This would demand for Celestia "main" database stored externally on disk. I though we can avoid this...

375gnu:
That's not a problem at all. That id is just an index inside stars.dat, of course it will not change during different sessions and even users. We can even make it compatible with the current database where that id is a HIP index.

375gnu:
this should interested to you https://sourceforge.net/p/celestia/feature-requests/44/

pirogronian:
So, we will have one big stars.dat with all possible stars or other objects? It would be difficult and inpractical. Maybe add support for universal id to many kinds of input data? Or, at least, more than one stars.dat formatted file?*
*Of course itcwoldnt be enough, because I want one database for not only stars.

375gnu:
Maybe we should raise the question on the forum to gather opinions from other people? At least @LukeCEL works with stars.dat so we we should consider his point.

pirogronian:
Fully agreed.

src/celestia/qt/qtcelestialbrowser.cpp
pirogronian: I dont understand...

src/celengine/xindexdataloader.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include "dataloader.h"
+
+class CrossIndexDataLoader : public AstroDataLoader
+{
+public:
+    static constexpr const char CROSSINDEX_FILE_HEADER[] = "CELINDEX";
+    int catalog;
+    CrossIndexDataLoader() = default;
+    CrossIndexDataLoader(AstroDatabase *db) { m_db = db; }
```
375gnu: prefer moder syntax `CrossIndexDataLoader(AstroDatabase *db) : m_db(db) {}`

src/celengine/astrocat.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    uint32_t tyc3 = index / 1000000000;
+    index -= tyc3 * 1000000000;
+    uint32_t tyc2 = index / 10000;
+    index -= tyc2 * 10000;
+    uint32_t tyc1 = index;
+    return fmt::sprintf("%s %u-%u-%u", m_prefix, tyc1, tyc2, tyc3);
+}
+
+constexpr int HipparcosAstroCatalog::MaxCatalogNumber;
```
375gnu: you should define a value for `constexpr` value

src/celengine/stardataloader.cpp
```diff
@@ -0,0 +1,384 @@
...
+        }
+//         clog << "End parsing entry.\n";
+    }
+
+    clog << "Successfully parsed " << successCount << " stars or barycenters.\n";
+
+    return true;
+}
+
+constexpr const char StarBinDataLoader::FILE_HEADER[];
```
375gnu: remove `constexpr` and define array's content

src/celengine/processoctree.cpp
```diff
@@ -0,0 +1,251 @@
+
+#include "star.h"
+#include "deepskyobj.h"
+#include "astrooctree.h"
+#include "processoctree.h"
+
+using namespace Eigen;
+
+static constexpr double SQRT3 = 1.732050807568877;
```
375gnu: static is unneeded in such expressions. `constexpr` means "this can be evaluated at compile time", in case of numbers it's like C macros but better.

src/celengine/processoctree.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+static constexpr double MAX_STAR_ORBIT_RADIUS = 1;
+
+void create5FrustumPlanes(Frustum::PlaneType *frustumPlanes, Vector3d position, Quaternionf orientation, float fovY, float aspectRatio)
+{
+    Vector3d planeNormals[5];
+    Eigen::Matrix3f rot = orientation.toRotationMatrix();
+    double h = (float) tan(fovY / 2);
```
375gnu: remove `(float)`

src/celengine/processoctree.cpp
```diff
@@ -0,0 +1,251 @@
...
+using namespace Eigen;
+
+static constexpr double SQRT3 = 1.732050807568877;
+
+static constexpr double MAX_STAR_ORBIT_RADIUS = 1;
+
+void create5FrustumPlanes(Frustum::PlaneType *frustumPlanes, Vector3d position, Quaternionf orientation, float fovY, float aspectRatio)
+{
+    Vector3d planeNormals[5];
+    Eigen::Matrix3f rot = orientation.toRotationMatrix();
```
375gnu: Quaternionf rot = orientation.conjugate();

src/celengine/processoctree.cpp
```diff
@@ -0,0 +1,251 @@
...
+    double h = (float) tan(fovY / 2);
+    double w = h * aspectRatio;
+    planeNormals[0] = Vector3d(0.0, 1.0, -h);
+    planeNormals[1] = Vector3d(0.0, -1.0, -h);
+    planeNormals[2] = Vector3d(1.0, 0.0, -w);
+    planeNormals[3] = Vector3d(-1.0, 0.0, -w);
+    planeNormals[4] = Vector3d(0.0, 0.0, -1.0);
+    for (int i = 0; i < 5; i++)
+    {
+        planeNormals[i] = rot.cast<double>().transpose() * planeNormals[i].normalized();
```
375gnu: planeNormals[i] = rot * planeNormals[i].normalized();

src/celengine/processoctree.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+using namespace Eigen;
+
+static constexpr double SQRT3 = 1.732050807568877;
+
+static constexpr double MAX_STAR_ORBIT_RADIUS = 1;
+
+void create5FrustumPlanes(Frustum::PlaneType *frustumPlanes, Vector3d position, Quaternionf orientation, float fovY, float aspectRatio)
+{
+    Vector3d planeNormals[5];
```
375gnu: I suppose that for Frustum double precision is sufficient

375gnu:
I'd suggest using LukeCEL's stars.dat (it contains 2M stars) for performance testing.

I'm not familiar with the original code so I need to read and understand it first to evaluate your changes.

src/celengine/processoctree.h
```diff
@@ -0,0 +1,67 @@
...
+    StarProcesor& procesor,
+    const Eigen::Vector3d& obsPos,
+    const Frustum::PlaneType *frustumPlanes,
+    float limitFactor,
+    OctreeProcStats * = nullptr);
+
+void processVisibleStars(
+    const OctreeNode *node,
+    StarProcesor& procesor,
+    Eigen::Vector3d position,
```
375gnu: &position and &orientation

src/celengine/processoctree.h
```diff
@@ -0,0 +1,67 @@
...
+    DsoProcesor& procesor,
+    const Eigen::Vector3d& obsPos,
+    const Frustum::PlaneType *frustumPlanes,
+    float limitFactor,
+    OctreeProcStats * = nullptr);
+
+void processVisibleDsos(
+    const OctreeNode *node,
+    DsoProcesor& procesor,
+    Eigen::Vector3d position,
```
375gnu: see above

src/celengine/astrooctree.cpp
```diff
@@ -1,177 +1,121 @@
...
             delete i;
-        delete m_children;
-    }
 }
 
-bool OctreeNode::add(Star *star)
+bool OctreeNode::add(LuminousObject *obj)
 {
-    m_stars.emplace_back(star);
+    m_objects.insert(make_pair(obj->getAbsoluteMagnitude(), obj));
```
375gnu: what will happen if 2/3/etc object have the same magnitude?

pirogronian: This is multimap. All should sit here without conflicts.

pirogronian:
Octree structure is proven to remain intact, but old bugs remains.

src/celengine/processoctree.cpp
```diff
@@ -0,0 +1,335 @@
...
+        return false;
+    if (selection.star()->getOctreeNode() == node)
+        return true;
+    return false;
+}
+
+void create5FrustumPlanes(Frustum::PlaneType *frustumPlanes, Vector3d position, Quaternionf orientation, float fovY, float aspectRatio)
+{
+    Vector3f planeNormals[5];
+    Quaternionf rot = orientation.conjugate();
```
375gnu: it looks like i told a shit, please revert a variant with matrix. `rot = orientation.toRotationMatrix();... planeNormals[i] = rot.transpose() * planeNormals[i].normalized();` this may be the cause of visibility bugs

pirogronian: Bug remains...

src/celengine/astrodb.cpp
```diff
@@ -339,3 +339,18 @@ AstroCatalog::IndexNumber AstroDatabase::getAutoIndex()
...
+float AstroDatabase::avgDsoMag() const
+{
+    float avg = 0;
+    size_t n = m_dsos.size();
+    for(const auto & dso : m_dsos)
+    {
+        if (dso->getAbsoluteMagnitude() > 8)
+            avg += dso->getAbsoluteMagnitude();
+        else
+            n--;
```
375gnu: inverted logic :) why not just `n=0; ...avg +=...;n++;`?

pirogronian: Such an approach was in original code. So, why not...

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,396 @@
...
+    return catalogNumberToString(nr);
+}
+
+std::vector<std::string> AstroDatabase::getObjectNameList(AstroCatalog::IndexNumber nr, int max) const
+{
+    std::vector<std::string> ret;
+    NameDatabase::NumberIndex::const_iterator iter = m_nameDB.getFirstNameIter(nr);
+    while (iter != m_nameDB.getFinalNameIter() && iter->first == nr && max > 0)
+    {
+        ret.emplace_back(iter->second);
```
375gnu: `emplace_back` moves an object into a container, `push_back` should be used instead

src/celengine/astrodb.cpp
```diff
@@ -0,0 +1,396 @@
...
+    {
+        if (max == 0)
+            break;
+        AstroCatalog::IndexNumber inr = indexToCatalogNumber(it.first, nr);
+        if (inr == AstroCatalog::InvalidIndex)
+        {
+//             cout << "Invalid cross index entry for catalog " << it.first << "[" << nr << "]\n";
+            continue;
+        }
+        ret.emplace_back(it.second->catalogNumberToName(inr));
```
375gnu: Again, it `catalogNumberToName` returns a new string, then `emplace_back` is ok and preferred, but if it return an existing object, then use `push_back`

src/celengine/astrooctree.cpp
```diff
@@ -0,0 +1,389 @@
...
+    return child;
+}
+
+bool OctreeNode::isInFrustum(const Frustum::PlaneType *planes) const
+{
+    for (unsigned int i = 0; i < 5; ++i)
+    {
+        const Frustum::PlaneType& plane = planes[i];
+
+        float r = m_scale * plane.normal().cwiseAbs().sum() * 1.1;
```
375gnu: `* 1.1` looks like hack, we need to check Frustum calculation

pirogronian: Indeed, however I spent already too many time and pain on it... For now I leave it for someone smarter.

375gnu:
@pirogronian you might be interested in http://liu.diva-portal.org/smash/get/diva2:1247648/FULLTEXT01.pdf "Interactive out-of-core rendering and filtering of one billion stars measured by the ESA Gaia mission". They discuss octree for 10^9 stars.

src/celestia/celestiacore.cpp
```diff
@@ -1187,7 +1194,7 @@ void CelestiaCore::charEntered(const char *c_p, int modifiers)
             if (typedTextCompletionIdx >= 0) {
                 string::size_type pos = typedText.rfind('/', typedText.length());
                 if (pos != string::npos)
-                    typedText = typedText.substr(0, pos + 1) + typedTextCompletion[typedTextCompletionIdx];
+                    typedText = typedText.substr(0, pos + 1) + (const string&) typedTextCompletion[typedTextCompletionIdx];
```
375gnu: why do need a cast here?

src/celestia/celestiacore.cpp
```diff
@@ -3971,19 +3988,19 @@ class SolarSystemLoader : public EnumFilesHandler
...
     ContentType contentType;
     ProgressNotifier* notifier;
 
-    CatalogLoader(OBJDB* db,
+    CatalogLoader(AstroDataLoader &loader,
                   const std::string& typeDesc,
                   const ContentType& contentType,
                   ProgressNotifier* pn) :
-        objDB      (db),
+        m_loader      (loader),
```
375gnu: this class doesn't use `m_` style. don't mix styles.

src/celengine/universe.cpp
```diff
@@ -218,22 +196,22 @@ bool Universe::isMarked(const Selection& sel, int priority) const
...
+    ClosestStarFinder(double _maxDistance, const Universe* _universe);
     ~ClosestStarFinder() = default;
-    void process(const Star& star, float distance, float appMag);
+    virtual void process(const Star *star, double distance, float appMag) override;
 
 public:
-    float maxDistance;
-    float closestDistance;
-    Star* closestStar;
+    double maxDistance;
```
375gnu: don't mix absolutely different things, if you want a double precision, make another PR

src/celengine/astrooctree.cpp
```diff
@@ -0,0 +1,386 @@
...
+    m_parent(parent)
+{
+    for(int i = 0; i < 8; i++)
+        m_children[i] = nullptr;
+}
+
+OctreeNode::~OctreeNode()
+{
+    for (auto &i : m_children)
+        if (i != nullptr)
```
375gnu: it's safe to pass nullptr to delete

src/celengine/astrooctree.cpp
```diff
@@ -0,0 +1,386 @@
...
+    fmt::fprintf(cout, "\n");
+}
+
+OctreeNode::OctreeNode(const Vector3d& cellCenter, double scale, size_t maxObj, OctreeNode *parent) :
+    m_cellCenter(cellCenter),
+    m_scale(scale),
+    m_maxObjCount(maxObj),
+    m_parent(parent)
+{
+    for(int i = 0; i < 8; i++)
```
375gnu: hardcoded size

src/celengine/astrooctree.cpp
```diff
@@ -0,0 +1,386 @@
...
+    auto i = objectIterator(obj);
+    if (i == m_objects.end())
+        return false;
+    m_objects.erase(i);
+    obj->setOctreeNode(nullptr);
+    setDirty(true);
+    return true;
+}
+
+OctreeNode *OctreeNode::getChild(int i, bool create)
```
375gnu: get should get, not create

src/celengine/name.cpp
```diff
@@ -1,120 +1,38 @@
...
 {
-    NameIndex::const_iterator iter = nameIndex.find(name);
-
-    if (iter == nameIndex.end())
+    for (const auto &ch : *m_ptr)
     {
-        iter = nameIndex.find(ReplaceGreekLetterAbbr(name));
-        if (iter == nameIndex.end())
-            return InvalidCatalogNumber;
+        m_hash += ch;
```
375gnu: i'm afraid this hash will have 100500 collisions

src/celengine/name.cpp
```diff
@@ -1,120 +1,38 @@
...
-
-    std::vector<std::string> completion;
-    int name_length = UTF8Length(name);
-
-    for (NameIndex::const_iterator iter = nameIndex.begin(); iter != nameIndex.end(); ++iter)
-    {
-        if (!UTF8StringCompare(iter->first, name, name_length, true))
+        const char *s = m_canonical.str().c_str();
+        const char *l = gettext(s);
+        if (s == l)
```
375gnu: leave a comment why pointer comparison is safe here

src/celengine/name.h
```diff
@@ -1,58 +1,140 @@
...
 
+bool inline operator==(const Name& n1, const Name& n2)
+{
+    return n1.ptr() == n2.ptr() || n1.str() == n2.str();
+}
 
-    uint32_t getNameCount() const;
+bool inline operator!=(const Name& n1, const Name& n2)
+{
+    return n1.ptr() != n2.ptr() && n1.str() != n2.str();
```
375gnu: return `! (n1 == n2)`

src/celengine/name.h
```diff
@@ -1,58 +1,140 @@
...
+class Name
+{
+protected:
+    std::shared_ptr<std::string> m_ptr { nullptr };
+    size_t m_hash { 0 };
+    void makeHash();
+    const static std::string m_empty;
+public:
+    Name(const char *s) { m_ptr = std::make_shared<std::string>(s); makeHash(); }
+    Name(const std::string& s) { m_ptr = std::make_shared<std::string>(s); makeHash(); }
```
375gnu: &s?

src/celengine/name.h
```diff
@@ -1,58 +1,140 @@
...
+    }
+    NameInfo(const NameInfo &other)
+    {
+        m_canonical = other.m_canonical;
+        m_localized = other.m_localized;
+        m_object = other.m_object;
+    }
+    bool hasLocalized() const
+    {
+        if (m_canonical.ptr() == m_localized.ptr())
```
375gnu: return m_canonical.ptr() != m_localized.ptr()

src/celengine/name.h
375gnu: More general questions: 1) why do want a sharable name? 2) why are trying to own a pointer to an object you don't own?

src/celengine/universe.cpp
pirogronian: It will be hard to revert. Moreover, it's already part of stars and dsos common interface for objects handled by octree. I see no reason to maintain separate versions of it, double and float....

src/celengine/name.cpp
pirogronian: Maybe... I used hash before latest octree optimisation, to check if it will increase performance.

src/celengine/name.h
pirogronian: 1. On certain stage of developement I was afraid of number of string indexes. So I started looking for a method to save memory.
2. I don't understand what you mean. Probably because it's my first ever attempt to deal with shared_ptr. :-)

375gnu:
On 6/7/19, Łukasz Buczyński <notifications@github.com> wrote:
>  public:
> -    float maxDistance;
> -    float closestDistance;
> -    Star* closestStar;
> +    double maxDistance;
>
> It will be hard to revert. Moreover, it's already part of stars and dsos
> common interface for objects handled by octree. I see no reason to maintain
> separate versions of it, double and float....

The reason that we still need to support machines with limited
resources unable to handle a lot of objects and in their case double
precision is just a memory waste.

pirogronian:
> On 6/7/19, Łukasz Buczyński ***@***.***> wrote: public: - float maxDistance; - float closestDistance; - Star* closestStar; + double maxDistance; It will be hard to revert. Moreover, it's already part of stars and dsos common interface for objects handled by octree. I see no reason to maintain separate versions of it, double and float....
> The reason that we still need to support machines with limited resources unable to handle a lot of objects and in their case double precision is just a memory waste.

DSOs already have double precision positions. I only generalized it on other kinds of objects. Is it really so much increase of memory? Switching to doubles adds only 12 bytes per object, while sizeof(Star) have currently 184 bytes of size.

src/celestia/celestiacore.cpp
pirogronian: Because it has now Name type. However, `.str()` would be shorter, indeed.

375gnu: Not just shorter, but more clear

375gnu:
@pirogronian this might be interesting for you http://nomis80.org/code/octree.html

src/celengine/namedb.cpp
```diff
@@ -82,6 +82,13 @@ std::vector<Name> NameDatabase::getCompletion(const std::string& name, bool gree
             completion.push_back(iter->first);
         }
     }
+    for (NameMap::const_iterator iter = m_localizedIndex.begin(); iter != m_localizedIndex.end(); ++iter)
```
375gnu:
```
for(const auto& local : m_localizedIndex)
```
?

pirogronian: Indeed. Just copied existing code and forgot about optimisations.

pirogronian:
> `getLocalizedName()` and `getName(true)` do the same. Do we need 2 similar functions?

The code evolves constantly. Maybe some functions will be reventually emoved...

> instead of `str().c_str()` you can add `c_str()` to your object.

I was sure I have both...

pirogronian:
@375gnu
It's very sad. I hoped to unify shared resources handling using shared_ptr and enable, for example many lua scipts at once or "resident" scripts. However, I pribabky already notuciced considerable perfirmance drop due to shared poiters in names handling. It suggest to avoid its usage completely.

pirogronian:
I decided to remove this branch from 1.7 release, as amount of changes seems to be too big and time to finish it too long.

375gnu:
Nobody urges you :)

pirogronian:
More and more people is waiting for new official release. Not everyone can easly download code from git and build it.

375gnu:
We provide win32 & win64 builds, munix9 provides opensuse and appimage packages.

Imho current 1.7 doesn't have any killer feature to prefer it over 1.6.1. From time to time I think that we should release 1.6.1 with patches floating in different distros and 64bit build as 1.6.2.

pirogronian:
IMHO these new builds should be avaliable as an option on official download page. For now official Linux binary is 1.5.1 instead of 1.6.1 and all binaries are 32bit, what a shame!

pirogronian:
Well, with passing NameInfo by reference, CPU load drops to master level...

PR #249, opened 2019-03-26
Fix unrecognition names directly searched or loaded.

### LINKED ISSUE #248: Greek letters parsing regression
As Celestia store names in already parsed form, it doesn't recognize directly entered or loaded greek names. Suggestion: parse provided names when searching for objects.

pirogronian:
Fixes #248 . Feel free to merge if once approved.

### DIFF
--- a/src/celutil/utf8.cpp
+++ b/src/celutil/utf8.cpp
@@ -772,39 +772,45 @@ std::string ReplaceGreekLetterAbbr(const std::string& str)
     Greek *instance = Greek::getInstance();
     std::string ret = str;
 
-    if (str[0] >= 'A' && str[0] <= 'Z' &&
-        str[1] >= 'A' && str[1] <= 'Z')
+    if (str[0] >= 'A' && str[0] <= 'Z')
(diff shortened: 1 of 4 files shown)


src/celutil/utf8.cpp
```diff
@@ -772,39 +772,45 @@ std::string ReplaceGreekLetterAbbr(const std::string& str)
     Greek *instance = Greek::getInstance();
     std::string ret = str;
 
-    if (str[0] >= 'A' && str[0] <= 'Z' &&
-        str[1] >= 'A' && str[1] <= 'Z')
+    if (str[0] >= 'A' && str[0] <= 'Z')
```
375gnu: Could you explain what's wrong with this file?

pirogronian: I added replacement also for Greek letters names, not only for cannonical abbreviations.

375gnu: Ok. Could you move this change into its own commit with the small explanation as above?

PR #252, opened 2019-03-27
Copy/paste search console typed text. For now used in Qt only.

pirogronian:
Due to shortcuts conflicts, I decided to bind keys: F5 for copy and F6 for paste, as in Midnight Commander.

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -1234,6 +1234,16 @@ void CelestiaAppWindow::createMenus()
     connect(gotoObjAct, SIGNAL(triggered()), this, SLOT(gotoObject()));
     navMenu->addAction(gotoObjAct);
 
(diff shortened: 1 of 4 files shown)


src/celestia/qt/qtappwin.cpp
```diff
@@ -425,7 +425,7 @@ void CelestiaAppWindow::init(const QString& qConfigFileName,
 
     connect(fullScreenAction, SIGNAL(triggered()), this, SLOT(slotToggleFullScreen()));
     viewMenu->addAction(fullScreenAction);
-
+
```
375gnu: extra spaces. you can a git hook to prevent commits with extra spaces. in `.git/hooks/` you have a `pre-commit.sample`, just remove `.sample`

pirogronian: Thanks. Recently my text editor lost ability to highlight extra spaces, not sure why.

PR #254, opened 2019-04-02
Remove dead code

375gnu:
The less garbage we have the better


375gnu:
> As I understand, shader programs weren't used anymore, as well as jpg/png support code?

Yes, we don't need those shaders anymore. They are OpenGL1 extension written in asm-like language and they are superseded by GLSL with OpenGL1.5 extension and OpenGL2.0 core.

As now we have OpenGL2.1 only those files were just confusing users :)

Jpg/png part is OSX specific and it was added years ago to support scenarios like "building Celestia without jpg or png support" or "building Celestia with OSX own jpg reader" as those scenarios are not supported anymore it's better to remove unmaintained code.

PR #259, opened 2019-04-06
Fix crash on create new lua rotation object.

pirogronian:


src/celestia/celx_rotation.cpp
```diff
@@ -12,17 +12,17 @@
 #include "celx.h"
 #include "celx_internal.h"
 #include "celx_vector.h"
+#include <fmt/printf.h>
```
375gnu: please remove this...

src/celestia/celx_rotation.cpp
```diff
@@ -12,17 +12,17 @@
...
 
 using namespace Eigen;
 
+#define dbg fmt::fprintf(cout, __FILE__ "[%u]\n", __LINE__)
```
375gnu: and these 2 lines

PR #264, opened 2019-04-08
Fix eigen::quaternion related crash

pirogronian:


pirogronian:
I tried to initialize `orientationUniv` in constructors, but it didn't help. So, `memcpy`...

PR #265, opened 2019-04-08
Fix alignment issues in CELX routines

375gnu:
Not finished yet, should address the same issue with celx_vector.cpp, etc


375gnu:
I'm not very good with memory allocators especially if they have additional requirements as lua does. Another problem is that different eigen types require different alignment afair, so a custom allocator may be to greedy in general case.

pirogronian:
On the other side, I'm for the first time faced with memory alignment problem... I scrached my head where to start :-)

Issue #269, opened 2019-04-15
compile error char_traits ambiguous


krisztianpinter:
trying to compile newest celestia with newest libfmt from github, getting "reference to ‘char_traits’ is ambiguous" in modelfile.cpp:371:75. complete error message attached.

this is a linux mint,
gcc -v says 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.11) 

[error.txt](https://github.com/CelestiaProject/Celestia/files/3081916/error.txt)

pirogronian:
@krisztianpinter
If you have any programming skills, try add `std::` prefox to `char_traits` or other ambigous tokens. Meanwhile I'm, preparing fix.

pirogronian:
@krisztianpinter
Please check out my PR #270 .

375gnu:
I'd say that's an fmt issue not our. @krisztianpinter why do you need the latest fmt?

krisztianpinter:
@375gnu i don't. i just didn't bother to pick a version. my plan is to put celestia on amazon linux, and it does not seem to have access to libfmt out of the box. so i just pulled it from github and compiled instead. i can use an older version if i know which older version to use. or maybe instead of compiling, i hunt down a good host for the package and install from there.

and btw yeah, i agree that it sounds fishy that a lib update breaks existing code. but i don't know c++ in depth enough to judge.

pirogronian:
@krisztianpinter 
It seems that `fmt` has its own `char_traits` template class: [https://github.com/fmtlib/fmt/search?q=char_traits&unscoped_q=char_traits](https://github.com/fmtlib/fmt/search?q=char_traits&unscoped_q=char_traits).
However, it's in namespace `internal`. Maybe it is his bug that is shown outside. Could You post an issue to them?

375gnu:
I have reported https://github.com/fmtlib/fmt/issues/1117

375gnu:
@krisztianpinter you don't need to compile fmt on your own, we provide git submodules for less common  dependencies, so i suggest doing the following:
* rm /path/to/local/fmt/installation
* git clone https://github.com/CelestiaProject/Celestia && cd Celestia
* git submodule update --init

then follow usual build procedure.

It seems we should update build instructions :)

krisztianpinter:
sigh. so much ado about nothing! :)
i think i can close this issue now?

375gnu:
[...]

PR #270, opened 2019-04-16
Fix (hopefuly) compile error reported in #269

pirogronian:
Fix (hopefuly) compile error reported in #269

### DIFF
--- a/src/celmodel/modelfile.cpp
+++ b/src/celmodel/modelfile.cpp
@@ -16,8 +16,6 @@
 
 
 using namespace cmod;
-using namespace std;
-
 
 // Material default values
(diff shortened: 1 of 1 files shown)


375gnu:
I'd prefer a bug report to fmt repo done instead of fixing their issues.

pirogronian:
If fmt authors introduced their own version of char_traits it doesn't seem to necessarily be a bug. However, can be reported for being sure.

375gnu:
If internal API interfere with public one then for it's a bug. Even if it won't be fixed there is no need to change a whole file, just adding `std::` to `char_traits` is enough.

375gnu:
fmt was fixed so i closing the PR

PR #271, opened 2019-04-21
Faster ring drawing using GL2.0 methods

375gnu:
Now only during the 1st frame ring shape is calculated, following frames are drawn much faster.

ringSystemSections option was deleted. If we want perfect circle shape of our rings we can do this using GLSL.

I have one issue unclarified regarding API used, so I'll keep it unmerged until I found an answer.


SevenSpheres:
@375gnu, please revert this per FarGetaNik [here](https://celestia.space/forum/viewtopic.php?p=146667#p146667)

PR #272, opened 2019-04-23
[WIP] Rewrite some vector math eigen-way

375gnu:
Just a cleanup [and speedup]


375gnu:
mathematically incorrect, because to rotate a vector using a quaternion on should do `q * v * q.conjugate()` instead of simple `q * v`

PR #275, opened 2019-04-27
Added debug octree stats.

pirogronian:
It displays, next to fps counter, stats of processing star and DSO octrees while looking for visible objects. From left to right numbers means: 
1. total objects count in visited nodes
2. total visited nodes count
3. max length of nodes path while traversing
I think this is usefull for reference while testing octree behaviour. Maybe display placement is a bit unfortunate, but I'm not familiar with render code enough to create better one.


pirogronian:
I didn't noticed any overhead, but of course, I can hide it.

375gnu:
you can ignore codacy results, sometimes they are false positive

Issue #287, opened 2019-05-17
All code should be namespaced


375gnu:
Some classes or structures have too generic names (e.g. Marker) that can cause name conflicts with 3rd party software used by Celestia (e.g. Marker is defined in system OSX library).

To prevent such conflicts and allow usage of Celestia as library in other projects we should namespace our code.

PR #288, opened 2019-05-17
Move math code into the celmath namespace

375gnu:
In context of #287


375gnu:
Initially I thought about common namespace, but 1) we already have 2 independent namespaces: celmath and cmod, 2) if basically I prefer different namespaces for different subsystems because 3) eventually we may want to make some parts of Celestia independent libraries (I hope Celestia will evolve and prosper :))

PR #289, opened 2019-05-20
Cmake cleanups

375gnu:


375gnu:
It's actually does the same what `find_package` does. So if you have required libraries then find_package finds them and pkg-config is nop. Of course you may have a very special kind of handcrafted packages which install their contents into location not supported by find_package but they still provide a valid pkg-config data, but in such cases it's better to pass additional variables to cmake to point it to proper locations. Explicit is better then implicit :)

PR #291, opened 2019-05-25
Move GL calls out of frontends, stage 1

375gnu:
I want to make frontends render-agnostic as much as possible, so I want to move direct GL calls out of them introducing helpers in the `Render` class.

f647c2d isn't finished, currently it uses `enum` as a key, but with `string` we won't need to update render.h and recompile all dependent files. And we don't call "OpenGL Info" dialog so often that we should care about performance. What do you think @pirogronian ? Maybe it worth to move some code into celestiacore (but qt version is too different).

0fae18b can be split into its own PR

### DIFF
--- a/src/celengine/command.cpp
+++ b/src/celengine/command.cpp
@@ -638,24 +638,30 @@ CommandCapture::CommandCapture(std::string _type,
 {
 }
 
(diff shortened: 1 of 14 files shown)


375gnu:
i'm merging a simplified version of the PR, other bits are in WIP state

PR #301, opened 2019-06-18
Disable GLContext usage as it does nothing now

375gnu:
maybe later when we have different renderers we'll restore it so i have used `#if 0` instead of removal. but why not to have smaller resulting binary?


pirogronian:
I suggest use #ifdef GL_CONTEXT or so. For future use it would need just to set flag on, instead of seeking and removing all appropriate #if 0s.

375gnu:
good point

PR #308, opened 2019-06-28
Fix assert in Renderer::renderMarker

f00zz:


375gnu:
Thanks for contribution!

PR #311, opened 2019-07-02
Use vertex buffer when displaying orbits

f00zz:
I noticed that there's already code in place to display orbits with vertex buffers, but disabled. It was mostly working except that colors are ignored (when displaying faded orbits), so I fixed it.

This isn't exactly modern OpenGL (glVertexPointer/etc are deprecated) but should be a bit more efficient than using immediate mode.

### DIFF
--- a/src/celengine/curveplot.cpp
+++ b/src/celengine/curveplot.cpp
@@ -31,7 +31,7 @@
 #if DEBUG_ADAPTIVE_SPLINE
 #define USE_VERTEX_BUFFER 0
 #else
-#define USE_VERTEX_BUFFER 0
+#define USE_VERTEX_BUFFER 1
 #endif
 
 #include "curveplot.h"
@@ -164,7 +164,7 @@ public:
         vbobj(0),
         currentStripLength(0)
     {
-        data = new Vector4f[capacity];
+        data = new Vertex[capacity];
     }
 
     ~HighPrec_VertexBuffer()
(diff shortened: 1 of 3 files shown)


f00zz:
Hi @375gnu, thanks for reviewing.

> And we have a problem with faded orbits when USE_VERTEX_BUFFER is set to 1: the are rendered as solid curves actually.

Right, that's why I added a color component to the vertex buffer, so that faded curves are rendered correctly. Otherwise we would need two different vertex buffers (one for solid curves and another one for faded curves).

I'll look into using shaders and VertexObject.

375gnu:
On 7/2/19, f00zz <notifications@github.com> wrote:

> Right, that's why I added a color component to the vertex buffer, so that
> faded curves are rendered correctly. Otherwise we would need two different
> vertex buffers (one for solid curves and another one for faded curves).

Indeed, you did it right, it's just due to my local config faded
orbits were disabled. Everything works fine.

> I'll look into using shaders and VertexObject.

Actually VertexObject is not ready for dynamically updated VBOs. If
have some changes in my WIP branches but currently it works with RO
buffers only.

If you don't object I'll merge this commit.

f00zz:
> If you don't object I'll merge this commit.

No problem, it's your call. It will still need to be ported to modern OpenGL eventually, like you have been doing.

Thank you for maintaining Celestia!

375gnu:
> No problem, it's your call. It will still need to be ported to modern OpenGL eventually, like you have been doing.

In my plans (or dreams) we'll have several renderers: gl2.1 for older hw, gles2 for raspberry pi and similar boards and 4.x for modern hw.

> Thank you for maintaining Celestia!

Join us :)

PR #312, opened 2019-07-03
Reimplement globular clusters using GLSL

375gnu:


375gnu:
Hi @pirogronian , could you check if this works for you? I have tested it with amd, intel and software drivers.

I still need to figure out why I have a different image.

pirogronian:
Globulars looks normal for me.
celestia

375gnu:
They look the same on their "start" distance, but when you go close you see the difference, possibly the problem is in point vs quad sizing.

pirogronian:
@375gnu Do You mean this white center, which seems to have distinguishable borser?

375gnu:
@pirogronian  Not only the center, but all stars:

That's initial image:
Экранны здымак_2019-07-04_18-44-41

That's after i get closer:
Экранны здымак_2019-07-04_18-42-19

Left is a master version, this PR on the right site. As you can see stars a much smaller on the right image.

pirogronian:
Indeed. Here is my comprision (master on right):
celestia-3
However I feel new display is much more realistic.

375gnu:
Finally. Original size are a bit larger but too blurry because of this.

PR #313, opened 2019-07-06
DSO .h file fixes

375gnu:
1. Render flags are 64 bit now
2. Modern C++ recommends to mark methods with `override` instead of `virtual` in this case you really override an existing method instead of introducing a new one by mistake.

### DIFF
--- a/src/celengine/globular.h
+++ b/src/celengine/globular.h
@@ -69,7 +69,7 @@ class Globular : public DeepSkyObject
                                             float pixelSize);
     GlobularForm* getForm() const;
 
-    virtual unsigned int getRenderMask() const;
+    virtual uint64_t getRenderMask() const;
     virtual unsigned int getLabelMask() const;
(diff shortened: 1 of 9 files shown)


src/celengine/globular.h
```diff
@@ -52,42 +52,42 @@ class Globular : public DeepSkyObject
...
- private:
-    float detail{ 1.0f };
-    std::string* customTmpName{ nullptr };
+    float         detail{ 1.0f };
+    std::string*  customTmpName{ nullptr };
     GlobularForm* form{ nullptr };
-    float r_c{ R_c_ref };
-    float c{ C_ref };
-    float tidalRadius{ 0.0f };
+    float         r_c{ R_c_ref };
```
pirogronian: Why these 3 whitespaces?

375gnu: I prefer names of variables are aligned.

PR #314, opened 2019-07-07
Terminator made for GL2

375gnu:


375gnu:
Hmm, we need rectangles as part of the overlay layer, so I will move drawRect there.

PR #325, opened 2019-07-19
Provide build files for Debian and derivatives

375gnu:


375gnu:
it works at least locally :) will test together with @munix9 in OBS

PR #326, opened 2019-07-21

lrm25:
Don't have a windows box with visual studio yet, so that still needs to be done


lrm25:
Hang on, let me get these merge records out.

PR #327, opened 2019-07-21

lrm25:


lrm25:
Still needs the Windows commit, I don't have Visual Studio yet.

src/celestia/qt/qtselectionpopup.cpp
```diff
@@ -180,20 +180,24 @@ SelectionPopup::SelectionPopup(const Selection& sel,
...
-    if (selection.body() != nullptr)
-    {
         // Alternate surface submenu
         QMenu* surfacesMenu = createAlternateSurfacesMenu();
         if (surfacesMenu != nullptr)
             addMenu(surfacesMenu);
 
+        if (selection.body()->hasPrimary())
+        {
+            QAction *action = new QAction(_("Primary"), this);
```
375gnu: I'd prefer a more descriptive title like "Select primary body"

src/celestia/gtk/menu-context.cpp
```diff
@@ -66,6 +67,10 @@ void menuContext(float, float, Selection sel)
             AppendMenu(popup, NULL, "S_ync Orbit", gtk_action_group_get_action(app->agMain, "SyncSelection"));
             /* Add info eventually:
              * AppendMenu(popup, NULL, "_Info", 0); */
+            if (sel.body()->hasPrimary())
+            {
+                AppendMenu(popup, GTK_SIGNAL_FUNC(handleContextPrimary), "Primary", NULL);
```
375gnu: The same as for Qt

src/celestia/qt/qtselectionpopup.h
```diff
@@ -59,6 +60,10 @@ Q_OBJECT
     void selectionInfoRequested(Selection& sel);
 
  private:
+    bool hasPrimary(Body *body);
```
375gnu: it seems these 4 prototypes shouldn't be here

src/celengine/body.h
```diff
@@ -361,6 +364,10 @@ class Body : public CatEntry
     void setName(const std::string& name);
     void recomputeCullingRadius();
 
+    bool hasPrimaryStar() const;
+    bool hasPrimaryBody(int classification) const;
+    bool hasPrimaryAroundBarycenter() const;
```
375gnu: just hasBarycenter()

src/celengine/body.h
```diff
@@ -240,6 +240,9 @@ class Body : public CatEntry
     PlanetarySystem* getSatellites() const;
     void setSatellites(PlanetarySystem*);
 
+    bool hasPrimary() const;
+    CatEntry* getPrimary() const;
```
375gnu: I'm not sure that CatEntry* is a correct type here, I suppose Selection should be used here

src/celengine/body.cpp
```diff
@@ -492,6 +492,90 @@ void Body::setSatellites(PlanetarySystem* ssys)
...
+        case Body::Comet:
+            return hasPrimaryStar();
+        case Body::Moon:
+        case Body::MinorMoon:
+            return hasPrimaryBody(Body::Planet | Body::DwarfPlanet) ||
+                   hasPrimaryAroundBarycenter();
+        case Body::Spacecraft:
+            return hasPrimaryBody(Body::Planet | Body::DwarfPlanet | Body::Asteroid |
+                                  Body::Comet | Body::Moon | Body::MinorMoon) ||
+                   hasPrimaryStar();
```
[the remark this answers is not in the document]
lrm25: How about "SpacecraftPrimaryBody"?

lrm25:
> I have a general concern whether your new methods should be part of the `Body` class. Backend should supply minimal required data to a frontend, and then it's a frontend's job to process it. As we don't have yet any basic frontend class, then you can either move your code to celestia/celestiacore.cpp or add a new celestia/helper.cpp.

I'm thinking helper.cpp since the functions don't really seem to fit in celestiacore.cpp.

src/celengine/body.cpp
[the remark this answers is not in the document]
375gnu: > How about "SpacecraftPrimaryBody"?

Yes, this sound much better.

lrm25:
For the helper module, would you prefer static functions or a class?

375gnu:
I'd prefer functions, the simpler architecture you help the better :)

lrm25:
Let me fix the commits.

PR #328, opened 2019-07-21
Remove constexpr for pow and exp for clang

eyvallah:
pow and exp are not constexpr using clang

This allows building on macOS


375gnu:
OSX image in Travis CI is broken for quite a while so we were not aware in clang limitattions

PR #332, opened 2019-08-01
[WIP] Overlay::rect() to be used instead of direct GL calls

375gnu:
+ shaderpoperties fix

### DIFF
--- a/src/celengine/overlay.cpp
+++ b/src/celengine/overlay.cpp
@@ -11,14 +11,19 @@
 #include <cstdarg>
 #include <celutil/utf8.h>
 #include <GL/glew.h>
+#include <Eigen/Core>
+#include <fmt/printf.h>
 #include "vecgl.h"
 #include "overlay.h"
+#include "render.h"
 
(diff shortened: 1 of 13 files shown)


src/celengine/overlay.cpp
```diff
@@ -161,27 +166,106 @@ void Overlay::print(const char* s)
...
+    default:
+        break;
     }
+    prog = r.getShaderManager().getShader(shader);
+
+    if (prog == nullptr)
+        return;
 
-    if (fill)
+    constexpr array<short, 8> texels = {0, 1,  1, 1,  1, 0, 0, 0};
```
pirogronian: Double spaces in two cases.

375gnu: actually one space is missing.

these are texture coordinates x0, y0, x1, y1, x2, y2, x3, y3 so with additional spaces it's seen easier where one texel finishes and another one starts.

375gnu:
I realized that I can simplify shaderprops a little bit more

PR #338, opened 2019-08-03
[appveyor] Use previous image

375gnu:
it looks like after the recent update we have issues with vcpkg. so let's just use an old image.


375gnu:
Because there are only 2 persons who can review PR, and I can't do that with my own :)

PR #339, opened 2019-08-04

lrm25:


src/celestia/helper.cpp
```diff
@@ -0,0 +1,99 @@
...
+            return true;
+        }
+    }
+    return false;
+}
+
+
+bool Helper::hasPrimary(Body* body)
+{
+    switch (body->getClassification()) {
```
375gnu: Please move `{` to the next line

src/celestia/helper.cpp
```diff
@@ -0,0 +1,99 @@
...
+        case Body::Planet:
```
375gnu: Please align `case` with `switch`

src/celestia/helper.cpp
```diff
@@ -0,0 +1,99 @@
...
+    return false;
+}
+
+
+Selection Helper::getPrimary(Body* body)
+{
+    PlanetarySystem* system = body->getSystem();
+    if (system == nullptr)
+    {
+        return Selection((Body*)nullptr);
```
375gnu: Just `return Selection()`. And `{}` are not required in a single line body (but you may leave them, we are not very consistent in this).

src/celestia/helper.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    Body* primaryBody = system->getPrimaryBody();
+    if (primaryBody != nullptr) {
```
375gnu: Please move `{` to the next line

src/celestia/helper.cpp
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    Star* primaryStar = system->getStar();
+    if (primaryStar != nullptr)
+    {
+        return Selection(primaryStar);
+    }
+
+    return Selection((Body*)nullptr);
```
375gnu: `return Selection()`

src/celestia/helper.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _HELPER_H_
+#define _HELPER_H_
+
+class Body;
+class Selection;
+
+class Helper
+{
+    public:
+        static bool hasPrimary(Body* body);
```
375gnu: Please fix indents: before `public` & `private` we use only 1 space, and 4 before methods.

Add `const` to `Body*` to tell a compiler that we don't modify `body`.

375gnu:
great. i'll implement a win32 version

PR #340, opened 2019-08-04
Use common base for reference counting classes.

pirogronian:


src/celengine/frame.h
```diff
@@ -23,14 +24,12 @@
  * (which specifies the coordinate axes at a given time) and the
  * nestingDepth() method (which is used to check for recursive frames.)
  */
-class ReferenceFrame
+class ReferenceFrame : public RefCounted
 {
+ protected:
```
375gnu: 1. Why `protected`?
1. Why do you override empty destructor with another empty one?

src/celengine/refcounted.cpp
```diff
@@ -0,0 +1,4 @@
+
```
375gnu: this file is not required, it's effectively empty

src/celengine/refcounted.cpp
```diff
@@ -0,0 +1,4 @@
...
+#include "refcounted.h"
+
+RefCounted::~RefCounted() {}
```
375gnu: and this is something absolutely weird, you mark `RefCounted::~RefCounted()` as 0 but here add an implementation

src/celengine/refcounted.h
```diff
@@ -0,0 +1,26 @@
+
```
375gnu: Please add a license

src/celengine/refcounted.h
```diff
@@ -0,0 +1,26 @@
...
+#pragma once
+
+class RefCounted
+{
+ protected:
```
375gnu: Why `protected`?

src/celengine/refcounted.h
```diff
@@ -0,0 +1,26 @@
...
+    mutable int m_refCount { 0 };
+    virtual ~RefCounted() = 0;
```
375gnu: not RAII

src/celengine/refcounted.h
```diff
@@ -0,0 +1,26 @@
...
+ public:
+    int addRef() const
```
375gnu: This method modifies internal state, why it's `const`?

src/celengine/solarsys.cpp
```diff
@@ -480,8 +480,8 @@ static bool CreateTimeline(Body* body,
         if (value->getType() != Value::ArrayType)
         {
             clog << "Error: Timeline must be an array\n";
-            delete defaultBodyFrame;
-            delete defaultOrbitFrame;
+            defaultBodyFrame->release();
```
375gnu: not RAII

src/celengine/refcounted.cpp
pirogronian: I thought exactly the same, but without it there is a linker error. I found this solution on the Internet...

src/celengine/frame.h
pirogronian: 1. To prevent manual destruction. I followed original implementation.
2. Again I followed original implementation. I presumed it's because of virtuality of descructor.

src/celengine/refcounted.cpp
pirogronian: As above...

src/celengine/refcounted.h
pirogronian: Again, I just followed original implementation :-)

pirogronian:
I wanted shared_ptr but you mentioned its poor performance. This implementation may be a bit rough, but it has fast lookup (pure pointer). Using shared_ptr for names nadling slowed Celestia down noticeably in my branch #247 .

375gnu:
@pirogronian 
In #247 your simply replaced plain pointers with shared_ptr, and because we have quite lot names, you noticed slowdown. Here we don't need so many objects so the slowdown should be neglible. And as a bonus your change should be thread safe so less changes in future.

pirogronian:
@375gnu
Using shared_ptr just because we have relatively small objects using it is temporary, while I want something more reliable. In the future number of objects using timelines can grow up (I plan to make it common in AstroObject). Moreover, I have to replace  shared pointers of names with something faster.

375gnu:
«Preliminary optimization is the root of all evil». How many objects will use shared timelines, how much it will slowdown data processing? Until we have numbers I consider shared_ptr a much better solution.

pirogronian:
Actually all bodies use timelines and so, timeline phases, which are already refcounted. All I did is making some code common. Lack of refcounting blocks me in my attempts to implement dynamic object modification and unloading.

375gnu:
Neither you is a C++ guru to implement correctly such refcounted object nor me to verify it. Let's use shared_ptr and only if we have noticeable slowdown think about alternative implementations.

pirogronian:
I didn't know that for implementing simple refcounting someone have to be a c++ guru... :-) Especially if this implementation already was in the code. However, I will did what you demand.

375gnu:
@pirogronian as shared_ptr is so slow, could you implement refcounting class looking more or less like smart pointer?

pirogronian:
@375gnu I hope so. I mentioned one proposition in [first comment](https://github.com/CelestiaProject/Celestia/pull/340#issuecomment-518029503).

PR #344, opened 2019-08-05
Introducing shared_ptr instead self-implemented of refcounted objects.

pirogronian:
Second commit increased CPU load by about 10% on my machine. Should I proceed or should there  be a better implementation?

### DIFF
--- a/src/celengine/body.cpp
+++ b/src/celengine/body.cpp
@@ -544,7 +544,7 @@ UniversalCoord Body::getPosition(double tdb) const
 {
     Vector3d position = Vector3d::Zero();
 
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     Vector3d p = phase->orbit()->positionAtTime(tdb);
     ReferenceFrame* frame = phase->orbitFrame();
 
@@ -569,7 +569,7 @@ UniversalCoord Body::getPosition(double tdb) const
  */
 Quaterniond Body::getOrientation(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     return phase->rotationModel()->orientationAtTime(tdb) * phase->bodyFrame()->getOrientation(tdb);
 }
 
@@ -578,7 +578,7 @@ Quaterniond Body::getOrientation(double tdb) const
  */
 Vector3d Body::getVelocity(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
 
     ReferenceFrame* orbitFrame = phase->orbitFrame();
 
@@ -599,7 +599,7 @@ Vector3d Body::getVelocity(double tdb) const
  */
 Vector3d Body::getAngularVelocity(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
 
     Vector3d v = phase->rotationModel()->angularVelocityAtTime(tdb);
 
@@ -622,7 +622,7 @@ Vector3d Body::getAngularVelocity(double tdb) const
  */
 Matrix4d Body::getLocalToAstrocentric(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     Vector3d p = phase->orbitFrame()->convertToAstrocentric(phase->orbit()->positionAtTime(tdb), tdb);
     return Eigen::Transform<double, 3, Affine>(Translation3d(p)).matrix();
 }
@@ -633,7 +633,7 @@ Matrix4d Body::getLocalToAstrocentric(double tdb) const
 Vector3d Body::getAstrocentricPosition(double tdb) const
 {
     // TODO: Switch the iterative method used in getPosition
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     return phase->orbitFrame()->convertToAstrocentric(phase->orbit()->positionAtTime(tdb), tdb);
 }
 
@@ -642,7 +642,7 @@ Vector3d Body::getAstrocentricPosition(double tdb) const
  */
 Quaterniond Body::getEclipticToFrame(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     return phase->bodyFrame()->getOrientation(tdb);
 }
 
@@ -652,7 +652,7 @@ Quaterniond Body::getEclipticToFrame(double tdb) const
  */
 Quaterniond Body::getEclipticToEquatorial(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     return phase->rotationModel()->equatorOrientationAtTime(tdb) * phase->bodyFrame()->getOrientation(tdb);
 }
 
@@ -662,7 +662,7 @@ Quaterniond Body::getEclipticToEquatorial(double tdb) const
  */
 Quaterniond Body::getEclipticToBodyFixed(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     return phase->rotationModel()->orientationAtTime(tdb) * phase->bodyFrame()->getOrientation(tdb);
 }
 
@@ -672,7 +672,7 @@ Quaterniond Body::getEclipticToBodyFixed(double tdb) const
 // meridian, and z-axis at a right angle the xy plane.
 Quaterniond Body::getEquatorialToBodyFixed(double tdb) const
 {
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
     return phase->rotationModel()->spin(tdb);
 }
 
--- a/src/celengine/body.cpp
+++ b/src/celengine/body.cpp
@@ -217,7 +217,7 @@ void Body::markUpdated()
 }
 
 
-const shared_ptr<const ReferenceFrame>& Body::getOrbitFrame(double tdb) const
+const ReferenceFrame::SharedConstPtr& Body::getOrbitFrame(double tdb) const
 {
     return timeline->findPhase(tdb)->orbitFrame();
 }
@@ -229,7 +229,7 @@ const Orbit* Body::getOrbit(double tdb) const
 }
 
 
-const shared_ptr<const ReferenceFrame>& Body::getBodyFrame(double tdb) const
+const ReferenceFrame::SharedConstPtr& Body::getBodyFrame(double tdb) const
 {
     return timeline->findPhase(tdb)->bodyFrame();
 }
--- a/src/celengine/body.cpp
+++ b/src/celengine/body.cpp
@@ -217,7 +217,7 @@ void Body::markUpdated()
 }
 
 
-const ReferenceFrame* Body::getOrbitFrame(double tdb) const
+const shared_ptr<const ReferenceFrame>& Body::getOrbitFrame(double tdb) const
 {
     return timeline->findPhase(tdb)->orbitFrame();
 }
@@ -229,7 +229,7 @@ const Orbit* Body::getOrbit(double tdb) const
 }
 
 
-const ReferenceFrame* Body::getBodyFrame(double tdb) const
+const shared_ptr<const ReferenceFrame>& Body::getBodyFrame(double tdb) const
 {
     return timeline->findPhase(tdb)->bodyFrame();
 }
@@ -546,7 +546,7 @@ UniversalCoord Body::getPosition(double tdb) const
 
     auto phase = timeline->findPhase(tdb);
     Vector3d p = phase->orbit()->positionAtTime(tdb);
-    ReferenceFrame* frame = phase->orbitFrame();
+    auto frame = phase->orbitFrame();
 
     while (frame->getCenter().getType() == Selection::Type_Body)
     {
@@ -580,7 +580,7 @@ Vector3d Body::getVelocity(double tdb) const
 {
     auto phase = timeline->findPhase(tdb);
 
-    ReferenceFrame* orbitFrame = phase->orbitFrame();
+    auto orbitFrame = phase->orbitFrame();
 
     Vector3d v = phase->orbit()->velocityAtTime(tdb);
     v = orbitFrame->getOrientation(tdb).conjugate() * v + orbitFrame->getCenter().getVelocity(tdb);
@@ -603,7 +603,7 @@ Vector3d Body::getAngularVelocity(double tdb) const
 
     Vector3d v = phase->rotationModel()->angularVelocityAtTime(tdb);
 
-    ReferenceFrame* bodyFrame = phase->bodyFrame();
+    auto bodyFrame = phase->bodyFrame();
     v = bodyFrame->getOrientation(tdb).conjugate() * v;
     if (!bodyFrame->isInertial())
     {
--- a/src/celengine/frame.cpp
+++ b/src/celengine/frame.cpp
@@ -729,7 +729,7 @@ FrameVector::createRelativeVelocityVector(const Selection& _observer,
 
 FrameVector
 FrameVector::createConstantVector(const Vector3d& _vec,
-                                  const ReferenceFrame* _frame)
+                                  const shared_ptr<const ReferenceFrame>& _frame)
 {
     FrameVector fv(ConstantVector);
     fv.vec = _vec;
--- a/src/celengine/frame.cpp
+++ b/src/celengine/frame.cpp
@@ -697,12 +697,6 @@ FrameVector::FrameVector(FrameVectorType t) :
 {
 }
 
-
-FrameVector::~FrameVector()
-{
-}
-
-
 FrameVector
 FrameVector::createRelativePositionVector(const Selection& _observer,
                                           const Selection& _target)
@@ -729,7 +723,7 @@ FrameVector::createRelativeVelocityVector(const Selection& _observer,
 
 FrameVector
 FrameVector::createConstantVector(const Vector3d& _vec,
-                                  const shared_ptr<const ReferenceFrame>& _frame)
+                                  const ReferenceFrame::SharedConstPtr& _frame)
 {
     FrameVector fv(ConstantVector);
     fv.vec = _vec;
--- a/src/celengine/frame.cpp
+++ b/src/celengine/frame.cpp
@@ -29,31 +29,9 @@ static const double ANGULAR_VELOCITY_DIFF_DELTA = 1.0 / 1440.0;
 /*** ReferenceFrame ***/
 
 ReferenceFrame::ReferenceFrame(Selection center) :
-    centerObject(center),
-    refCount(0)
+    centerObject(center)
 {
 }
-
-
-int
-ReferenceFrame::addRef() const
-{
-    return ++refCount;
-}
-
-
-int
-ReferenceFrame::release() const
-{
-    --refCount;
-    assert(refCount >= 0);
-    int refCountCopy = refCount;
-    if (refCount <= 0)
-        delete this;
-
-    return refCountCopy;
-}
-
 
 // High-precision rotation using 64.64 fixed point path. Rotate uc by
 // the rotation specified by unit quaternion q.
@@ -691,8 +669,6 @@ FrameVector::FrameVector(const FrameVector& fv) :
     vec(fv.vec),
     frame(fv.frame)
 {
-    if (frame != nullptr)
-        frame->addRef();
 }
 
 
@@ -705,11 +681,7 @@ FrameVector::operator=(const FrameVector& fv)
     target = fv.target;
     vec = fv.vec;
 
-    if (frame != nullptr)
-        frame->release();
     frame = fv.frame;
-    if (frame != nullptr)
-        frame->addRef();
 
     return *this;
 }
@@ -728,8 +700,6 @@ FrameVector::FrameVector(FrameVectorType t) :
 
 FrameVector::~FrameVector()
 {
-    if (frame != nullptr)
-        frame->release();
 }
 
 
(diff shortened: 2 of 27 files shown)


src/celengine/timeline.cpp
```diff
@@ -27,9 +27,7 @@ Timeline::~Timeline()
     for (auto phase : phases)
     {
         // Remove the phase from whatever phase tree contains it.
-        phase->getFrameTree()->removeChild(phase);
-
-        phase->release();
+        phase->getFrameTree()->removeChild(phase.get());
```
375gnu: I believe  FrameTree owns its children so it should accept sharedptr not raw.

src/celengine/timeline.cpp
```diff
@@ -46,14 +44,13 @@ Timeline::appendPhase(TimelinePhase* phase)
             return false;
     }
 
-    phase->addRef();
-    phases.push_back(phase);
+    phases.push_back(shared_ptr<TimelinePhase>(phase));
```
375gnu: have you tried to leave `phases.push_back(phase);`? compiler should generate calls to create shared_ptr

src/celengine/timeline.h
```diff
@@ -38,7 +39,7 @@ class Timeline
     void markChanged();
 
 private:
-    std::vector<TimelinePhase*> phases;
+    std::vector<std::shared_ptr<TimelinePhase> > phases;
```
375gnu: in modern c++ it's not required to add a space between > and > in templates

src/celengine/timelinephase.cpp
```diff
@@ -34,19 +34,14 @@ TimelinePhase::TimelinePhase(Body* _body,
...
-    refCount(0)
+    m_owner(_owner)
 {
     // assert(owner == orbitFrame->getCenter()->getFrameTree());
-    m_orbitFrame->addRef();
-    m_bodyFrame->addRef();
 }
 
 
 TimelinePhase::~TimelinePhase()
```
375gnu: `~TimelinePhase() = default` in header file

src/celengine/timeline.cpp
pirogronian: I believe it's done this way in second commit.

src/celengine/timeline.cpp
pirogronian: I'm nearly sure 've tried, also hoping compiler will do the rest.

src/celengine/timeline.h
pirogronian: Good to know. :-)

375gnu:
> Second commit increased CPU load by about 10% on my machine. Should I proceed or should there be a better implementation?

Damn, i didn't expected so high growth.

Can we live with the current code?

pirogronian:
> Damn, i didn't expected so high growth.

I believe it's due to pass shared_ptr as value. I try to make it passing as a reference.

> Can we live with the current code?

Yes, but also mainteining current set of features. So, good bye dynamic unloading...

375gnu:
> Yes, but also mainteining current set of features. So, good bye dynamic unloading...

Why you can't unload objects with current ref counters?

pirogronian:
> Why you can't unload objects with current ref counters?

I can and I tried it in this PR. I only wanted to make common base for such a ref counters. You know the rest :-)

pirogronian:
With passing shared_ptr as reference, even with additional shared_ptr usage in Timeline, CPU load seems to remain normal.

src/celengine/frametree.cpp
```diff
@@ -172,23 +170,22 @@ FrameTree::addChild(TimelinePhase* phase)
...
-        (*iter)->release();
         children.erase(iter);
         markChanged();
     }
 }
 
 
 /*! Return the child at the specified index. */
-TimelinePhase*
+const TimelinePhase*
```
375gnu: why do you have raw pointer here?

i suppose we *should* always return smart pointers, we *can* accept raw pointers (if we don't store them anywhere)

src/celengine/frametree.h
```diff
@@ -39,9 +40,9 @@ class FrameTree
 
     ReferenceFrame* getDefaultReferenceFrame() const;
 
-    void addChild(TimelinePhase* phase);
-    void removeChild(TimelinePhase* phase);
-    TimelinePhase* getChild(unsigned int n) const;
+    void addChild(std::shared_ptr<const TimelinePhase> &phase);
+    void removeChild(std::shared_ptr<const TimelinePhase> &phase);
+    const TimelinePhase* getChild(unsigned int n) const;
```
375gnu: and raw pointer here

src/celengine/solarsys.cpp
```diff
@@ -518,7 +509,7 @@ static bool CreateTimeline(Body* body,
         const Timeline* timeline = body->getTimeline();
         if (timeline->phaseCount() == 1)
         {
-            const TimelinePhase* phase = timeline->getPhase(0);
+            auto phase = timeline->getPhase(0);
```
375gnu: here you can use raw pointers to speedup a little bit:
`auto phase = timeline->getPhase(0).get();`

src/celestia/celx_phase.cpp
```diff
@@ -28,12 +28,10 @@ class PhaseReference
     PhaseReference(const TimelinePhase& _phase) :
     phase(&_phase)
 {
-        phase->addRef();
```
375gnu: Actually you should remove PhaseReference and use shared_ptr instead directly

src/celengine/frametree.cpp
pirogronian: Probably because I wanted to check impact on perfomance on the first place. Now I'm implementing it more comprehently and I'm stuck with domino effect... Probably will be finished tomorrow.

pirogronian:
After 246aa4c CPU still remains normal. However, lua interface for frames is not tested yet.

pirogronian:
CPU usage still normal :-)

src/celengine/frame.cpp
```diff
@@ -728,8 +700,6 @@ FrameVector::FrameVector(FrameVectorType t) :
 
 FrameVector::~FrameVector()
```
375gnu: can be replaced with `= default`in .h file

src/celengine/observer.cpp
```diff
@@ -816,16 +814,15 @@ void Observer::setFrame(ObserverFrame::CoordinateSystem cs, const Selection& ref
...
  *  universal coordinates will not change.
  */
-void Observer::setFrame(const ObserverFrame& f)
+void Observer::setFrame(const shared_ptr<const ObserverFrame>& f)
 {
-    if (frame != &f)
+    if (frame != f)
     {
-        auto* newFrame = new ObserverFrame(f);
+        auto newFrame = make_shared<ObserverFrame>(*f);
```
375gnu: `auto* newFrame = new ObserverFrame(f);` was required due to ownership and `new`/`delete` issues, now with `shared_ptr` we don't need this and instead of `newFrame` `f` should be used.

src/celengine/parseobject.h
```diff
@@ -46,12 +47,12 @@ RotationModel* CreateRotationModel(Hash* rotationData,
 
 RotationModel* CreateDefaultRotationModel(double syncRotationPeriod);
 
-ReferenceFrame* CreateReferenceFrame(const Universe& universe,
+std::shared_ptr<const ReferenceFrame> CreateReferenceFrame(const Universe& universe,
```
375gnu: why `const` here?

src/celengine/parseobject.h
```diff
@@ -46,12 +47,12 @@ RotationModel* CreateRotationModel(Hash* rotationData,
...
                                      Value* frameValue,
                                      const Selection& defaultCenter,
                                      Body* defaultObserver);
 
-TwoVectorFrame* CreateTopocentricFrame(const Selection& center,
+std::shared_ptr<const TwoVectorFrame> CreateTopocentricFrame(const Selection& center,
```
375gnu: and here?

src/celengine/timelinephase.h
```diff
@@ -42,7 +41,7 @@ class TimelinePhase
         return m_endTime;
     }
 
-    ReferenceFrame* orbitFrame() const
+    const std::shared_ptr<const ReferenceFrame>& orbitFrame() const
```
375gnu: why `const` here?

src/celengine/timelinephase.h
```diff
@@ -52,7 +51,7 @@ class TimelinePhase
         return m_orbit;
     }
 
-    ReferenceFrame* bodyFrame() const
+    const std::shared_ptr<const ReferenceFrame>& bodyFrame() const
```
375gnu: and here

src/celengine/timelinephase.h
```diff
@@ -78,48 +77,44 @@ class TimelinePhase
         return m_startTime <= t && t < m_endTime;
     }
 
-    static TimelinePhase* CreateTimelinePhase(Universe& universe,
+    static std::shared_ptr<const TimelinePhase> CreateTimelinePhase(Universe& universe,
```
375gnu: again

src/celengine/timelinephase.h
```diff
@@ -78,48 +77,44 @@ class TimelinePhase
...
+    TimelinePhase& operator=(const TimelinePhase& phase) = delete;
 
 private:
     Body* m_body;
 
     double m_startTime;
     double m_endTime;
 
-    ReferenceFrame* m_orbitFrame;
+    std::shared_ptr<const ReferenceFrame> m_orbitFrame;
```
375gnu: again `const`

src/celengine/timelinephase.h
```diff
@@ -78,48 +77,44 @@ class TimelinePhase
...
     Orbit* m_orbit;
-    ReferenceFrame* m_bodyFrame;
+    std::shared_ptr<const ReferenceFrame> m_bodyFrame;
```
375gnu: and again

src/celestia/celx_phase.cpp
```diff
@@ -98,7 +73,7 @@ static int phase_timespan(lua_State* l)
 
     celx.checkArgs(1, 1, "No arguments allowed for to phase:timespan");
 
-    const TimelinePhase* phase = this_phase(l);
+    auto& phase = *this_phase(l);
```
375gnu: reference to pointer == pointer to pointer

src/celengine/body.cpp
```diff
@@ -544,9 +544,9 @@ UniversalCoord Body::getPosition(double tdb) const
 {
     Vector3d position = Vector3d::Zero();
 
-    const TimelinePhase* phase = timeline->findPhase(tdb);
+    auto phase = timeline->findPhase(tdb);
```
375gnu: why did you remove `const` here

src/celengine/timelinephase.h
pirogronian: I prefer to declare references as const whenever possible to avoid unneeded-modification-bugs.

src/celestia/celx_phase.cpp
pirogronian: Actually this is a reference to shared_ptr.

src/celengine/body.cpp
pirogronian: Probably because returned reference is const already.

pirogronian: Probably because returned reference is const already.

src/celengine/timelinephase.h
375gnu: Celestia already has a bit more than allowed `const_cast`s :)

pirogronian: I didn't see any yet. I may start removing it in another PR.

375gnu: It's better not to introduce new

pirogronian: I dont, do I?

375gnu: Potentially with `const` put here and there once we will face a situation when api change or static_const is required.

pirogronian: Still feel const is better. If someone really need to modify particular data, then maybe it's a time to rethink general approach. Personally, I like to know exactly where and who modify particular data.

src/celengine/frametree.h
```diff
@@ -92,15 +92,15 @@ class FrameTree
 private:
     Star* starParent;
     Body* bodyParent;
-    std::vector<TimelinePhase*> children;
+    std::vector<TimelinePhase::SharedConstPtr > children;
```
375gnu: extra space

src/celengine/observer.cpp
```diff
@@ -57,7 +57,7 @@ static Vector3d slerp(double t, const Vector3d& v0, const Vector3d& v1)
 
 Observer::Observer()
 {
-    frame = new ObserverFrame();
+    frame = make_shared<ObserverFrame>();
```
375gnu:
```
Observer::Observer() :
    frame(make_shared<ObserverFrame>())
{
...
```

src/celengine/observer.cpp
```diff
@@ -83,13 +83,12 @@ Observer::Observer(const Observer& o) :
     locationFilter(o.locationFilter),
     displayedSurface(o.displayedSurface)
 {
-    frame = new ObserverFrame(*o.frame);
+    frame = make_shared<ObserverFrame>(*o.frame);
```
375gnu: like above

src/celengine/observer.cpp
```diff
@@ -816,25 +814,19 @@ void Observer::setFrame(ObserverFrame::CoordinateSystem cs, const Selection& ref
...
     {
-        auto* newFrame = new ObserverFrame(f);
-
-        if (frame != nullptr)
-        {
-            convertFrameCoordinates(newFrame);
-            delete frame;
-        }
-        frame = newFrame;
+        convertFrameCoordinates(f);
```
375gnu: could you check `convertFrameCoordinates`? i'm not sure that your change is equivalent to the original code

src/celengine/observer.cpp
```diff
@@ -1460,8 +1446,6 @@ ObserverFrame& ObserverFrame::operator=(const ObserverFrame& f)
 
 ObserverFrame::~ObserverFrame()
```
375gnu: this can be moved to .h (= default)

src/celengine/observer.h
```diff
@@ -66,30 +69,30 @@ class ObserverFrame
...
+    const ReferenceFrame::SharedConstPtr& getFrame() const;
 
     UniversalCoord convertFromUniversal(const UniversalCoord& uc, double tjd) const;
     UniversalCoord convertToUniversal(const UniversalCoord& uc, double tjd) const;
     Eigen::Quaterniond convertFromUniversal(const Eigen::Quaterniond& q, double tjd) const;
     Eigen::Quaterniond convertToUniversal(const Eigen::Quaterniond& q, double tjd) const;
 
-    static UniversalCoord convert(const ObserverFrame* fromFrame,
-                                  const ObserverFrame* toFrame,
+    static UniversalCoord convert(const ObserverFrame::SharedConstPtr &fromFrame,
```
375gnu: please be consistent with `&` positions. in some places you have Foo &f, in other Foo& f

src/celengine/parseobject.cpp
```diff
@@ -1307,7 +1307,7 @@ getFrameCenter(const Universe& universe, Hash* frameData, const Selection& defau
 }
 
 
-static BodyFixedFrame*
+static shared_ptr<const BodyFixedFrame>
```
375gnu: BodyFixedFrame::Ptr

src/celestia/celx_phase.cpp
375gnu: if it possible to put copy of shared_ptr to luauserdata? if not then it's better to put a raw pointer to data instead of the raw pointer to a smart pointer

pirogronian: That's how it's implemented already. Lua userdata is always presented to user as a raw pointer by default. So, by default we always have raw pointer to smart pointer. :-)

375gnu: okay

src/celengine/solarsys.cpp
```diff
@@ -368,17 +363,15 @@ TimelinePhase* CreateTimelinePhase(Body* body,
         rotationModel = new ConstantOrientation(Quaterniond::Identity());
     }
 
-    TimelinePhase* phase = TimelinePhase::CreateTimelinePhase(universe,
+    auto phase = TimelinePhase::CreateTimelinePhase(universe,
                                                               body,
```
375gnu: incorrect indentation

src/celengine/solarsys.cpp
```diff
@@ -407,7 +400,7 @@ Timeline* CreateTimelineFromArray(Body* body,
         bool isFirstPhase = iter == timelineArray->begin();
         bool isLastPhase = *iter == timelineArray->back();
 
-        TimelinePhase* phase = CreateTimelinePhase(body, universe, phaseData,
+        auto phase = CreateTimelinePhase(body, universe, phaseData,
```
375gnu: incorrect indentattion

src/celengine/solarsys.cpp
```diff
@@ -629,22 +617,18 @@ static bool CreateTimeline(Body* body,
...
-            delete defaultOrbitFrame;
-            delete orbitFrame;
             delete rotationModel;
             return false;
         }
 
         // We finally have an orbit, rotation model, frames, and time range. Create
         // the object timeline.
-        TimelinePhase* phase = TimelinePhase::CreateTimelinePhase(universe,
+        auto phase = TimelinePhase::CreateTimelinePhase(universe,
```
375gnu: incorrect indentation

src/celengine/timelinephase.cpp
```diff
@@ -126,12 +80,12 @@ TimelinePhase::CreateTimelinePhase(Universe& universe,
         return nullptr;
     }
 
-    TimelinePhase* phase = new TimelinePhase(body,
+    auto phase = make_shared<const TimelinePhase>(body,
```
375gnu: indentation

src/celengine/timelinephase.h
```diff
@@ -78,48 +78,41 @@ class TimelinePhase
         return m_startTime <= t && t < m_endTime;
     }
 
-    static TimelinePhase* CreateTimelinePhase(Universe& universe,
+    static TimelinePhase::SharedConstPtr CreateTimelinePhase(Universe& universe,
```
375gnu: indentation

src/celestia/celx_phase.cpp
```diff
@@ -211,15 +186,14 @@ static int phase_gc(lua_State* l)
 {
     CelxLua celx(l);
 
-    PhaseReference* ref = static_cast<PhaseReference*>(celx.checkUserData(1, Celx_Phase));
-    if (ref == nullptr)
+    auto & ref = *this_phase(l);
+    if (ref)
```
375gnu: incorrect

should be smth like
```
auto t = this_phase(l);
if (t == nullptr)
{
   auto& ref = *t;
``

File: src/celestia/celx_phase.h [hunk 33]
```diff
@@ -12,10 +12,17 @@
...
 class TimelinePhase;
 
+inline int celxClassId(const std::shared_ptr<const TimelinePhase>&)
+{
+    return Celx_Phase;
+}
+
 extern void CreatePhaseMetaTable(lua_State* l);
-extern int phase_new(lua_State* l, const TimelinePhase& phase);
+extern int phase_new(lua_State* l,  const shared_ptr<const TimelinePhase>& phase);
```
375gnu: extra space before const

src/celestia/celx_phase.h
```diff
@@ -12,10 +12,17 @@
 #ifndef _CELX_PHASE_H_
 #define _CELX_PHASE_H_
 
+#include <memory>
+
 struct lua_State;
 class TimelinePhase;
 
+inline int celxClassId(const std::shared_ptr<const TimelinePhase>&)
```
375gnu: i don't see where this fn is used

src/celestia/celx.cpp
```diff
@@ -1851,7 +1851,7 @@ void CelxLua::newFrame(const ObserverFrame& f)
     frame_new(m_lua, f);
 }
 
-void CelxLua::newPhase(const TimelinePhase& phase)
+void CelxLua::newPhase(const shared_ptr<const TimelinePhase>& phase)
```
375gnu: not changed to TimelinePhase::Ptr

src/celengine/solarsys.cpp
pirogronian: What do you mean?

src/celengine/timelinephase.h
pirogronian: Again, what do you mean?

src/celestia/celx_phase.cpp
pirogronian: `this_phase` already checks for nullptr.

src/celestia/celx_phase.h
pirogronian: In `celx_internal.h` at line 316.

src/celengine/solarsys.cpp
375gnu: `body` should start just under `universe`

src/celengine/timelinephase.h
375gnu: all arguments if placed on several lines should start on the same column

src/celestia/celx_phase.cpp
375gnu: in this case this code is too obscure. references and smart pointers pointing to nullptr is a bad practice

pirogronian: Even if program should never enter this code with null pointer? Maybe `assert(t != nullptr)`?

pirogronian: @375gnu So, are you sure I should do additional check in every place i get pointer to data? And should it be `if { }` statement or just `assert` would be enough?

375gnu: I'd prefer `assert`

pirogronian:
I can't see any remained notice...

375gnu:
@pirogronian 
https://github.com/CelestiaProject/Celestia/pull/344/files/3a05fa914f88ecabe292ccf3d884da196418c719#diff-7be42549281eeed64f5a9370badd11c8L521

375gnu:
@pirogronian now it looks good, could you cleanup commit history?

375gnu:
@pirogronian why did you merge dirty history?

pirogronian:
I was too fast, sorry. Could you revert it? I will later restore this branch and clean it.

375gnu:
@pirogronian I already merged all your commits into a single one

PR #346, opened 2019-08-06
Fix mingw

dlshcbmuipmam:
Now you can build the source with Mingw + Msys2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,7 @@ include(GNUInstallDirs)
 include(CheckIncludeFiles)
 include(rpath)
 include(install_to_extras_subdir)
-if (NOT MSVC)
+if (NOT MSVC AND NOT MINGW)
   include(cotire)
 else()
   macro(cotire _target)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ include(GNUInstallDirs)
 include(CheckIncludeFiles)
 include(rpath)
 include(install_to_extras_subdir)
-if (NOT MSVC)
+if (NOT MSVC AND NOT MINGW)
```
375gnu: Actually only MSVC has issues with CMake >= 3.12. On the other hand cotire is not updated for quite a while, it has issue with C headers included into C++ files, so we will need to remove it once.

dlshcbmuipmam: Probably it is good to remove the whole cotire if it doesn't provide much, it was causing some weird namespace bug in my mingw test build.

By the way I have modified my PR about the GIT_COMMIT in the CMakeList.txt . I found that it should be "%h" in MSVC and '%h' in Mingw in order to compile the code correctly, weird behaviour.

PR #347, opened 2019-08-07
Fix mouse move resulting in wrong movement in simulation

eyvallah:
The mouse movement mapped to celestiacore is incorrect.

The offset distance should be the current location offset by the location the last time mouseMoveEvent is called, whereas the current implementation uses the current location offset by  the location the last time when the cursor is still visible. To resolve it saveLocalCursorPos and saveGlobalCursorPos is in both occasion.

### DIFF
--- a/src/celestia/qt/qtglwidget.cpp
+++ b/src/celestia/qt/qtglwidget.cpp
@@ -204,20 +204,23 @@ void CelestiaGlWidget::mouseMoveEvent(QMouseEvent* m)
             // Hide the cursor.
             setCursor(QCursor(Qt::BlankCursor));
             cursorVisible = false;
-
-            // Save the cursor position
-            QPoint pt;
-            pt.setX(x);
-            pt.setY(y);
-
-            // Store a local and global location
-            saveLocalCursorPos = pt;
-            pt = mapToGlobal(pt);
-            saveGlobalCursorPos = pt;
+        }
+        else
+        {
+            // Calculate mouse delta from local coordinate then move it back to the saved location
+            appCore->mouseMove(x - saveLocalCursorPos.rx(), y - saveLocalCursorPos.ry(), buttons);
         }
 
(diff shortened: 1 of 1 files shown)


375gnu:
With your change objects are "trembling" when being moved or rotated. Current implementation keeps cursor in its original place so mouse movements are not limited by a screen and we can rotate object by 360 deg, etc.

eyvallah:
@375gnu in that case, it seems to be a Mac only issue, the line QCursor::setPos(saveGlobalCursorPos); doesn't restore the cursor location, causing there to be larger movement as you drag away from the place where dragging started, can you test if removing the next line "QCursor::setPos(saveGlobalCursorPos);" combining with this pull request acts correctly on other platform.

375gnu:
With your suggestion there is no more "trembling" but movements are still limited by window borders and cursor doesn't restore its original position.

eyvallah:
> With your suggestion there is no more "trembling" but movements are still limited by window borders and cursor doesn't restore its original position.
> 
> UPD: That's actually like gtk+, glut and native osx frontends behave. But native win32 behaves exactly like current qt version which I find much more convenient.
> 
> @eyvallah why do you consider such behaviour incorrect?
> 
> Some explanations - https://github.com/CelestiaProject/Celestia/blob/master/src/celestia/win32/winmain.cpp#L3755

it's a qt issue on mac, i think. QCursor::setPos doesn't set the cursor back for me, it is where the problem lies.

375gnu:
@eyvallah as they set with Info.plist or should be granted manually?

eyvallah:
@375gnu  no need for Info.plist item, it  has something to do with what Apple called "Trusted Event" since such events are invoked by app instead of the user, which is very general.  it will request the permission (which will result in a popup) when related API is triggered, if denied, the API wouldn't work and popup wouldn't be happening again

PR #351, opened 2019-08-08
Fix cmake luajit

dlshcbmuipmam:
vcpkg install the header files in include/luajit which is not included in the findluajit script

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -101,7 +101,7 @@ Install required packages:
 
 Install optional packages:
 
-  vcpkg install qt5
+  vcpkg install qt5 luajit
 
 Configure and build 32-bit version:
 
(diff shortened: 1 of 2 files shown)


INSTALL
```diff
@@ -101,7 +101,7 @@ Install required packages:
 
 Install optional packages:
 
-  vcpkg install qt5
+  vcpkg install qt5 luajit
```
375gnu: vcpkg installs lua & luajit into the same place, at least the version I have.

Could you add a sentence that if a user wants luajit then s/he might need to remove lua?

dlshcbmuipmam: @375gnu I think you are using an old vcpkg, it has been fixed in https://github.com/microsoft/vcpkg/pull/5863/commits 

Right now I am using 2019.07 tag version. It is no need to remove lua in order to install luajit.

375gnu: great

Issue #350, opened 2019-08-07
MinGW debug build fail


dlshcbmuipmam:
The debug build in MSYS2 + MinGW by 
`cmake .. -G "MSYS Makefiles" -DENABLE_WIN=OFF -DCMAKE_BUILD_TYPE=Debug`
takes forever to link the celestia-qt.exe. This can be related directly to MinGW, or gcc9.1 (the compiler in my MinGW). 

This should not be a high priority issue, but it is better to keep an eye on what cause this problem.

dlshcbmuipmam:
I accidentally found a solution, use the lld linker instead of the default linker
cmake .. -G "MSYS Makefiles" -DENABLE_WIN=OFF -DCMAKE_CXX_FLAGS='-fuse-ld=lld'

375gnu:
@dlshcbmuipmam could you document the workaround in INSTALL.txt?

dlshcbmuipmam:
> @dlshcbmuipmam could you document the workaround in INSTALL.txt?

@375gnu Sure.

PR #355, opened 2019-08-12
Update INSTALL instruction for Mingw debug build

dlshcbmuipmam:
#350

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -168,6 +168,17 @@ Configure and build:
 
 Instead of N, pass a number of CPU cores you want to use during a build.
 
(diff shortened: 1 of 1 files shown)


dlshcbmuipmam:
@pirogronian  This is for Mingw build only, so pacman is the only choice :-) 

Debug builds on the other compiler doesn't need this tweak as far as I know, though it seems like that the lld linker is really nice.

dlshcbmuipmam:
Wait, I was too late, I miss the -DCMAKE_BUILD_TYPE=Debug flag in this instruction.

375gnu:
No problem, just create a new PR :)

PR #358, opened 2019-08-12
Qt - Prepend APPDIR for absolute paths for AppImage builds

munix9:
Added cmake option ENABLE_APPIMAGE and compiler definition APPIMAGE
Prepend APPDIR env var for absolute paths for AppImage builds,
see https://docs.appimage.org/packaging-guide/environment-variables.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,17 +21,18 @@ set(DISPLAY_NAME "Celestia")
 #
 #
 #
-option(ENABLE_CELX    "Enable celx scripting, requires Lua library? (Default: on)" ON)
-option(ENABLE_SPICE   "Use spice library? (Default: off)" OFF)
-option(ENABLE_NLS     "Enable interface translation? (Default: on)" ON)
-option(ENABLE_GLUT    "Build simple Glut frontend? (Default: on)" OFF)
-option(ENABLE_GTK     "Build GTK2 frontend (Unix only)? (Default: off)" OFF)
-option(ENABLE_QT      "Build Qt frontend? (Default: on)" ON)
-option(ENABLE_WIN     "Build Windows native frontend? (Default: on)" ON)
-option(ENABLE_THEORA  "Support video capture to OGG Theora? (Default: on)" ON)
-option(ENABLE_TOOLS   "Build different tools? (Default: off)" OFF)
-option(NATIVE_OSX_APP "Support native OSX paths read data from (Default: off)" OFF)
-option(FAST_MATH      "Build with unsafe fast-math compiller option (Default: off)" OFF)
+option(ENABLE_CELX     "Enable celx scripting, requires Lua library? (Default: on)" ON)
+option(ENABLE_SPICE    "Use spice library? (Default: off)" OFF)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -120,6 +122,10 @@ if(_UNIX)
   find_package(PkgConfig)
 endif()
 
+if(_UNIX AND ENABLE_APPIMAGE)
```
375gnu: On the line 69 you set ENABLE_APPIMAGE to OFF on non Unix-platforms

src/celestia/qt/qtappwin.cpp
```diff
@@ -208,6 +208,12 @@ void CelestiaAppWindow::init(const QString& qConfigFileName,
         QString dataDir = QApplication::applicationDirPath() + "/../Resources";
 #else
         QString dataDir = CONFIG_DATA_DIR;
+#ifdef APPIMAGE
+        QString appimageDir = QString::fromLocal8Bit(::getenv("APPDIR"));
```
375gnu: just a remark: in qt it's preferred to use qgetenv because on some platforms (i suppose latest windows) getenv isn't available. No need to change it here, as this code is Unix-only

src/celestia/qt/qtappwin.cpp
```diff
@@ -208,6 +208,12 @@ void CelestiaAppWindow::init(const QString& qConfigFileName,
...
+        if (!appimageDir.isEmpty()) {
```
375gnu: please move `{` to a new line

src/celestia/qt/qtmain.cpp
```diff
@@ -76,6 +76,12 @@ int main(int argc, char *argv[])
     QString splashDir = QApplication::applicationDirPath() + "/../Resources/splash/";
 #else
     QString splashDir = SPLASH_DIR;
+#ifdef APPIMAGE
+    QString appimageDir = QString::fromLocal8Bit(::getenv("APPDIR"));
+    if (!appimageDir.isEmpty()) {
```
375gnu: please move { to a new line

src/celestia/qt/qtmain.cpp
```diff
@@ -92,6 +98,11 @@ int main(int argc, char *argv[])
     QString localeDir = QApplication::applicationDirPath() + "/../Resources/locale";
 #else
     QString localeDir = LOCALEDIR;
+#ifdef APPIMAGE
+    if (!appimageDir.isEmpty()) {
```
375gnu: please move { to a new line

munix9:
Thank you for your comments.
I don't like my solution very much.
It would be better to define the query for getenv("APPDIR") centrally for Qt (then more likely qgetenv("APPDIR")) and for GTK - or even for the whole application - and then place the result in the appropriate places in the code.
But someone with more experience in C/C++ than me would have to do that.

375gnu:
I have some plans to implement a new base application class moving all common routines used in different frontends there. But for now I can add some code lines to celestia/helper.cpp

PR #361, opened 2019-08-15
C++17-like filesystem API + small fixes

375gnu:
This library will be used as a fallback if std::experimental::filesystem is not available.

PR #363, opened 2019-08-16
use const char * on debug print

eyvallah:


src/celengine/category.cpp
```diff
@@ -99,13 +99,13 @@ void UserCategory::cleanup()
     while(!m_objlist.empty())
     {
         auto it = m_objlist.begin();
-        DPRINTF(1, "Removing object: %s\n", it->getName());
+        DPRINTF(1, "Removing object: %s\n", it->getName().c_str());
```
375gnu: this is not required because #361 reimplemented DPRINTF using fmtlib, so now it's valid to use %s with std::string

src/celengine/category.cpp
```diff
@@ -99,13 +99,13 @@ void UserCategory::cleanup()
...
         removeObject(*it);
     }
     while(!m_catlist.empty())
     {
         auto it = m_catlist.begin();
-        DPRINTF(1, "Removing category: %s\n", (*it)->name());
+        DPRINTF(1, "Removing category: %s\n", (*it)->name().c_str());
```
375gnu: The same as above

Issue #106, opened 2018-11-12
Draw orbits lines: complete/fading


pirogronian:
We were asked for implement orbits display modes: complete vs fading.
As I can see, there are only fading for now.
Should it be part of oncoming 1.7 release?

375gnu:
core & qt are finished, native win and gtk are in progress

PR #365, opened 2019-08-16
Update win32's view options dialog

375gnu:
Closes: #106


375gnu:
@dlshcbmuipmam that's an Render->View options dialog, first enable Orbits then orbits for particular  object types (e.g. Planets), then scale down to see complete orbits and when you check/uncheck "Show fading orbits" then they should change their appearance.

375gnu:
it seems that reseditor add garbage to resource.h

PR #368, opened 2019-08-17
Fix errors introduced in filesystem branch

### LINKED ISSUE #367: Xcode's clang seems to dislike passing object in varargs
the cmake generated Makefile and Xcode project works, however the native project I'm maintaining is still throwing error (in addition to the class name clash) in compiling with the fmt commit.

for passing string/fs in DebugPrint:
Can not pass object of non-trivial type /xxx/ through variadic function; call will abort at runtime

for the vprintf in debug.cpp:
No matching function for call to 'vfprintf'
note: candidate function [with S = const char *, Char = char] not viable: no known conversion from 'va_list' (aka '__builtin_va_list') to 'basic_format_args<typename basic_printf_context_t<internal::basic_buffer<char> >::type>' (aka 'basic_format_args<basic_printf_context<back_insert_iterator<fmt::v5::internal::basic_buffer<char> >, char> >') for 3rd argument

375gnu:
@eyvallah does this change fix #367 ?

### DIFF
--- a/src/celutil/debug.cpp
+++ b/src/celutil/debug.cpp
@@ -7,44 +7,7 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-#ifdef _MSC_VER
-#include <windows.h>
-#endif
-#include <cstdio>
-#include <cstdarg>
-#include <config.h>
-#include <fmt/printf.h>
-
-static int debugVerbosity = 0;
-
-#if defined(DEBUG) || defined(_DEBUG)
-void DebugPrint(int level, const char *format, ...)
-{
-    va_list args;
-    va_start(args, format);
-
-    if (level <= debugVerbosity)
-    {
-#ifdef _MSC_VER
-        if (IsDebuggerPresent())
-        {
-            char buf[1024];
-            fmt::vsprintf(buf, format, args);
-            OutputDebugStringA(buf);
-        }
-        else
-        {
-            fmt::vfprintf(stdout, format, args);
-        }
-#else
-        fmt::vfprintf(stderr, format, args);
-#endif
-    }
-
-    va_end(args);
-}
-#endif /* DEBUG */
-
+int debugVerbosity = 0;
 
 void SetDebugVerbosity(int dv)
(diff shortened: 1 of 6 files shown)


src/celutil/debug.cpp
```diff
@@ -10,38 +10,30 @@
...
-#include <config.h>
+#include <iostream>
 #include <fmt/printf.h>
 
 static int debugVerbosity = 0;
 
 #if defined(DEBUG) || defined(_DEBUG)
-void DebugPrint(int level, const char *format, ...)
+template <typename... T>
+void DebugPrint(int level, const char *format, const T & ... args)
```
eyvallah: the template function should be declared and implemented in header?, this is not helping with the error of passing an object as varargs

375gnu: correct, i just checked g++ debug.cpp, will fix it later

375gnu:
@eyvallah does this updated pr fix your issue #367

375gnu:
@eyvallah does this updated pr fix your issue #367

375gnu:
@eyvallah does this updated pr fix your issue #367

src/celutil/debug.h
```diff
@@ -10,31 +10,41 @@
...
-
 #if !defined(_DEBUG) && !defined(DEBUG)
-#define DPRINTF(level, args...)
+#define DPRINTF(level, format, ...)
 #else
-#define DPRINTF(level, args...) DebugPrint(level, args)
-extern void DebugPrint(int level, const char *format, ...);
-#endif
-
+#define DPRINTF(level, format, ...) DebugPrint(level, format __VA_OPT__(,) __VA_ARGS__)
```
eyvallah: drop this define and it passes.

375gnu: could you provide an error log with line? this is perfectly valid c++11 syntax so it should not cause any issues in theory.

375gnu: i saw that particular error in travis but now i have other errors :)

375gnu:
@eyvallah now it should be ok

PR #372, opened 2019-08-23
get rid of glBegin/end in texture font

eyvallah:


eyvallah:
for later

375gnu:
I actually want to replace bitmap fonts with truetype so we have better localization support.

On the other hand glEnableClientState family is not supported in OpenGL ES 2.0 and it's too slow for this usage.

PR #373, opened 2019-08-25
Use context menu handler for callback

eyvallah:
only modified the qt version.

### DIFF
--- a/src/celestia/celestiacore.h
+++ b/src/celestia/celestiacore.h
@@ -197,8 +197,6 @@ class CelestiaCore // : public Watchable<CelestiaCore>
         ShowFrame     = 0x010,
     };
 
-    typedef void (*ContextMenuFunc)(float, float, Selection);
-
  private:
     class OverlayImage
     {
@@ -317,8 +315,6 @@ class CelestiaCore // : public Watchable<CelestiaCore>
     int getOverlayElements() const;
     void setOverlayElements(int);
 
-    void setContextMenuCallback(ContextMenuFunc);
-
     void addWatcher(CelestiaWatcher*);
     void removeWatcher(CelestiaWatcher*);
 
@@ -371,6 +367,16 @@ class CelestiaCore // : public Watchable<CelestiaCore>
 
     void setCursorHandler(CursorHandler*);
     CursorHandler* getCursorHandler() const;
+
+    class ContextMenuHandler
+    {
+    public:
+        virtual ~ContextMenuHandler() = default;
+        virtual void requestContextMenu(float, float, Selection) = 0;
+    };
+
+    void setContextMenuHandler(ContextMenuHandler*);
+    ContextMenuHandler* getContextMenuHandler() const;
 
     void toggleReferenceMark(const std::string& refMark, Selection sel = Selection());
     bool referenceMarkEnabled(const std::string& refMark, Selection sel = Selection()) const;
@@ -485,14 +491,13 @@ class CelestiaCore // : public Watchable<CelestiaCore>
     MovieCapture* movieCapture{ nullptr };
     bool recording{ false };
 
-    ContextMenuFunc contextMenuCallback{ nullptr };
-
     Texture* logoTexture{ nullptr };
 
     Alerter* alerter{ nullptr };
     std::vector<CelestiaWatcher*> watchers;
     CursorHandler* cursorHandler{ nullptr };
     CursorShape defaultCursorShape{ CelestiaCore::CrossCursor };
+    ContextMenuHandler* contextMenuHandler{ nullptr };
 
     std::vector<Url*> history;
     std::vector<Url*>::size_type historyCurrent{ 0 };
--- a/src/celestia/gtk/menu-context.cpp
+++ b/src/celestia/gtk/menu-context.cpp
@@ -42,16 +42,18 @@ static GtkMenu* CreateAlternateSurfaceMenu(const vector<string>& surfaces);
 static AppData* app;
 
 
-/* Initializer. Sets the appData, since there is no way to pass it. */
-void initContext(AppData* a)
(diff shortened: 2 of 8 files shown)


eyvallah:
@375gnu I brought the change also to other platforms, I can't test it though because I'm on Mac only, the gtk one is apparently wrong, there needs to be some mechanism to pass the app argument.

src/celestia/celestiacore.h
```diff
@@ -372,6 +368,16 @@ class CelestiaCore // : public Watchable<CelestiaCore>
     void setCursorHandler(CursorHandler*);
     CursorHandler* getCursorHandler() const;
 
+    class ContextMenuHandler
+    {
+    public:
+        virtual ~ContextMenuHandler() {};
```
375gnu: {} can be replaced with `= default`

src/celestia/gtk/menu-context.cpp
```diff
@@ -25,33 +25,25 @@
...
-
-/* There is no way to pass the AppData struct to the menu at this time. This
- * keeps the global variable local to this file, at least conceptually. */
-static AppData* app;
-
-
-/* Initializer. Sets the appData, since there is no way to pass it. */
-void initContext(AppData* a)
+/* Initializer. Sets the appData */
+GTKContextMenuHandler::GTKContextMenuHandler(AppData* app) : CelestiaCore::ContextMenuHandler(), app(app)
```
375gnu: please change style:
```
GTKContextMenuHandler::GTKContextMenuHandler(AppData* _app) :
    CelestiaCore::ContextMenuHandler(),
    app(_app)
```

src/celestia/gtk/menu-context.h
```diff
@@ -21,11 +21,14 @@
 
 #include "common.h"
 
+class GTKContextMenuHandler : public CelestiaCore::ContextMenuHandler {
```
375gnu: move { to the next line

src/celestia/gtk/menu-context.h
```diff
@@ -21,11 +21,14 @@
...
+public:
```
375gnu: before public/private we use 1 space indent

src/celestia/kde/kdeapp.cpp
```diff
@@ -106,7 +106,7 @@ static uint32 FilterOtherLocations = ~(Location::City |
                     Location::Vallis |
                     Location::Mare);
 
-KdeApp::KdeApp(std::string config, std::string dir, std::vector<std::string> extrasDirs, bool fullscreen, bool disableSplash) : KMainWindow(0, 0)
+KdeApp::KdeApp(std::string config, std::string dir, std::vector<std::string> extrasDirs, bool fullscreen, bool disableSplash) : KMainWindow(0, 0), CelestiaCore::ContextMenuHandler()
```
375gnu: no need to change kde frontend, it's dead

src/celestia/qt/qtappwin.cpp
```diff
@@ -182,7 +176,7 @@ void FPSActionGroup::updateFPS(int fps)
 }
 
 CelestiaAppWindow::CelestiaAppWindow(QWidget* parent) :
-    QMainWindow(parent)
+    QMainWindow(parent), CelestiaCore::ContextMenuHandler()
```
375gnu: please move CelestiaCore::ContextMenuHandler() to the next line

src/celestia/win32/winmain.cpp
```diff
@@ -2349,6 +2343,17 @@ class WinAlerter : public CelestiaCore::Alerter
     }
 };
 
+class WinContextMenuHandler : public CelestiaCore::ContextMenuHandler
+{
+public:
+    WinContextMenuHandler() : CelestiaCore::ContextMenuHandler() {};
```
375gnu: i don't think that you need to define a constructor here

src/celestia/gtk/menu-context.h
eyvallah: i fixed this one but it seems our win32 code doesn't follow that

eyvallah:
i think ill stick with the delegate class instead of the lambda for now, to match with alerter, cursor handler, watcher etc.

375gnu:
> the gtk one is apparently wrong,

yes, i tried to compile it and it threw a lot of errors, and i'm not able to fix them yet as C code doesn't map to C++ properly.

eyvallah:
@375gnu falls back to the old way for gtk for now, can you verify if it passes compilation?

src/celestia/gtk/menu-context.h
```diff
@@ -21,11 +21,15 @@
...
 
-/* Initializer Function */
-void initContext(AppData* a);
+    void requestContextMenu(float, float, Selection sel);
 
-/* Entry Function */
-void menuContext(float, float, Selection sel);
+ private:
+    AppData *appData;
+}
```
375gnu: missing ';'

src/celestia/gtk/menu-context.h
```diff
@@ -21,11 +21,15 @@
...
+    GTKContextMenuHandler(AppData* app);
 
-/* Initializer Function */
-void initContext(AppData* a);
+    void requestContextMenu(float, float, Selection sel);
 
-/* Entry Function */
-void menuContext(float, float, Selection sel);
+ private:
+    AppData *appData;
```
375gnu: it seems that now you don't need to store AppData * anymore

pirogronian:
@375gnu I think class is just more general and flexible than function.

eyvallah:
@375gnu yes, as pirogronian said, it's more flexible and it's in line with the other delegate/handler method in the celestiacore.

375gnu:
@eyvallah feel free to merge, we usually merge own pr ourselves

PR #379, opened 2019-08-27
[WIP] [qt] Dump more technical information

375gnu:


375gnu:
@pirogronian @dlshcbmuipmam @eyvallah do you have any ideas what to display? I would like to have as much info as possible so we can narrow down issue in case of problem reports.

Also do you have any suggestion regarding design?

PR #381, opened 2019-08-29
define macros for print levels and avoid direct use of fmt::printf

eyvallah:


src/celestia/oggtheoracapture.cpp
```diff
@@ -309,13 +309,13 @@ bool OggTheoraCapture::start(const std::string& filename,
...
     yuv.uv_stride=video_x/2;
 
-    fmt::printf(_("OggTheoraCapture::start() - Theora video: %s %.2f(%d/%d) fps quality %d %dx%d offset (%dx%d)\n"),
-                filename.c_str(),
-                (double)video_hzn/(double)video_hzd,
-                video_hzn,video_hzd,
-                video_q,
-                video_x,video_y,
-                frame_x_offset,frame_y_offset);
+    DPRINTF(PRINT_LEVEL_VERBOSE,
```
pirogronian: I think _(...) may stay here. It's better to have localized messages.

src/celestia/oggtheoracapture.cpp
```diff
@@ -424,7 +424,7 @@ void OggTheoraCapture::cleanup()
 
     if(outfile)
     {
-        fmt::printf(_("OggTheoraCapture::cleanup() - wrote %d frames\n"), video_frame_count);
+        DPRINTF(PRINT_LEVEL_VERBOSE, "OggTheoraCapture::cleanup() - wrote %d frames\n", video_frame_count);
```
pirogronian: Same here.

src/celutil/debug.h
```diff
@@ -20,6 +20,10 @@
 #undef DPRINTF // OSX has DPRINTF
 #endif // DPRINTF
 
+#define PRINT_LEVEL_ERROR       0
```
pirogronian: Verbosity levels are usually ordered from Error to Debug, ie Debug is more verbise than normal info and normal info is mire verbose than error. In more sophisticated systems I saw following levels:
error
warning
info
verbose
debug

src/cel3ds/3dsread.cpp
```diff
@@ -620,18 +620,18 @@ M3DScene* Read3DSFile(ifstream& in)
     unsigned short chunkType = readUshort(in);
     if (chunkType != M3DCHUNK_MAGIC)
     {
-        DPRINTF(0, "Read3DSFile: Wrong magic number in header\n");
+        DPRINTF(PRINT_LEVEL_ERROR, "Read3DSFile: Wrong magic number in header\n");
```
375gnu: Please rename PRINT_LEVEL_ to LOG_LEVEL_

375gnu:
on the other hand why not to use https://github.com/gabime/spdlog?

eyvallah:
> on the other hand why not to use https://github.com/gabime/spdlog?

let's leave it to another PR :)

eyvallah:
@pirogronian pls approve

PR #384, opened 2019-09-01
Remove F5 F6 shortcuts for copy and paste to solve conflicts with setTargetSpeed

dlshcbmuipmam:
Previously, because of the shortcut setting in Qt, pressing F5 F6 have no effect on the traveling speed, while F1, F2, F3, F4, F7 work.

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -1173,12 +1173,12 @@ void CelestiaAppWindow::createMenus()
     navMenu->addAction(gotoObjAct);
 
     QAction *copyAction = new QAction(QIcon(":/icons/clip_copy.png"), _("Copy search console text"), this);
-    copyAction->setShortcut(QString("F5"));
+    // copyAction->setShortcut(QString("F5")); // conflict with setTargetSpeed
     connect(copyAction, &QAction::triggered, this, &CelestiaAppWindow::copyText);
     navMenu->addAction(copyAction);
 
     QAction *pasteAction = new QAction(QIcon(":/icons/clip_paste.png"), _("Paste into search console"), this);
-    pasteAction->setShortcut(QString("F6"));
+    // pasteAction->setShortcut(QString("F6")); // conflict with setTargetSpeed
     connect(pasteAction, &QAction::triggered, this, &CelestiaAppWindow::pasteText);
     navMenu->addAction(pasteAction);
 
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -1121,16 +1121,6 @@ void CelestiaAppWindow::createMenus()
     connect(copyImageAction, SIGNAL(triggered()), this, SLOT(slotCopyImage()));
     fileMenu->addAction(copyImageAction);
 
-    QAction* copyURLAction = new QAction(QIcon(":/icons/clip_copy.png"), _("Copy &URL"), this);
-    copyURLAction->setShortcut(QKeySequence::Copy);
-    connect(copyURLAction, SIGNAL(triggered()), this, SLOT(slotCopyURL()));
-    fileMenu->addAction(copyURLAction);
-
-    QAction* pasteURLAction = new QAction(QIcon(":/icons/clip_paste.png"), _("&Paste URL"), this);
-    //pasteURLAction->setShortcut(QKeySequence::Paste);  // conflicts with cycle render path command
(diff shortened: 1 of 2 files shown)


375gnu:
@dlshcbmuipmam  @pirogronian I suppose it's better to replace them with other combinations rather them simply disabling. What about Ctrl+Shift+C/Ctrl+Shift+V?

dlshcbmuipmam:
@375gnu At least its seems like Ctrl+Shift+C corresponds to the copy image in the qt client. 

I think we should document the shortcuts in somewhere, I know there is a wiki page https://en.wikibooks.org/wiki/Celestia/Navigation but it doesn't include the client specific shortcut, so that we can discuss it in a systematic way.

dlshcbmuipmam:
And perhaps not every function worths a shortcut in a GUI application?

375gnu:
> And perhaps not every function worths a shortcut in a GUI application?

Maybe. So generally i'm ok with your change but let's wait for @pirogronian, he implemented this feature.

In the future we can we use the same Ctrl-C/Ctrl-V shortcuts. If a console is open then insert into it, else insert as URL.

pirogronian:
If shortcuts I intriduced needs to be changed or removed, let it be. But copying stuff from/to search console should have any, I think. Making clipvoard iperations context sensitive looks nice also. It could be even implemented inside CelestiaCore by external clipboard handler, context menu is.

dlshcbmuipmam:
@375gnu @pirogronian Actually, Ctrl-V for copying URL is commented out, saying "conflicts with cycle render path command", do you have any idea on what the render path command is?

dlshcbmuipmam:
@375gnu @pirogronian Since I cannot find any "conflicts with cycle render path command", I decided to allow the use of Ctrl+V.

I have combined both the copy and paste for search console and URL into `copyTextOrURL()` and `pasteTextOrURL()`, see if you agree on the change or not.

375gnu:
> Actually, Ctrl-V for copying URL is commented out, saying "conflicts with cycle render path command", do you have any idea on what the render path command is?

Previously Celestia had different render paths for different old OpenGL versions and extensions and users were able to change them in runtime.

Now we head to proper OpenGL 2.1 with some extensions so both other paths and means to switch them were removed.

pirogronian:
Indeed, @375gnu eliminated render paths some time ago. I dint know if it will be bring back in any form.

src/celestia/qt/qtappwin.cpp
```diff
@@ -1172,14 +1162,14 @@ void CelestiaAppWindow::createMenus()
     connect(gotoObjAct, SIGNAL(triggered()), this, SLOT(gotoObject()));
     navMenu->addAction(gotoObjAct);
 
-    QAction *copyAction = new QAction(QIcon(":/icons/clip_copy.png"), _("Copy search console text"), this);
-    // copyAction->setShortcut(QString("F5")); // conflict with setTargetSpeed
-    connect(copyAction, &QAction::triggered, this, &CelestiaAppWindow::copyText);
+    QAction *copyAction = new QAction(QIcon(":/icons/clip_copy.png"), _("Copy URL / Copy search console text"), this);
```
375gnu: maybe it worth to shorted it a bit: "Copy URL / console text"?

src/celestia/qt/qtappwin.cpp
```diff
@@ -1172,14 +1162,14 @@ void CelestiaAppWindow::createMenus()
...
+    copyAction->setShortcut(QString("Ctrl+C"));
+    connect(copyAction, &QAction::triggered, this, &CelestiaAppWindow::copyTextOrURL);
     navMenu->addAction(copyAction);
 
-    QAction *pasteAction = new QAction(QIcon(":/icons/clip_paste.png"), _("Paste into search console"), this);
-    // pasteAction->setShortcut(QString("F6")); // conflict with setTargetSpeed
-    connect(pasteAction, &QAction::triggered, this, &CelestiaAppWindow::pasteText);
+    QAction *pasteAction = new QAction(QIcon(":/icons/clip_paste.png"), _("Paste URL / Paste into search console"), this);
```
375gnu: the same here: "Paste URL / console text"

375gnu:
> I dint know if it will be bring back in any form.

Besides OpenGL2.1 for older PC I hope to have OpenGL4 and OpenGL ES 2 renderers. None of them can be switched in runtime without restart.

375gnu:
@pirogronian please approve or reject :)

PR #390, opened 2019-09-05
fix mac os x build and enable sandbox, codesign, etc

eyvallah:
1. modernize the project for opening
2. enabled sandboxing and code signing
3. remove precompiled libs frameworks and header in the repo
4. fix some compilation error


eyvallah:
@375gnu here is the edition for 1.6.1 working with current Xcode. however I would really discourage using it for 1.7.0. the project was falling behind liek

macosx/CelestiaAppCore.mm
```diff
@@ -22,6 +22,10 @@
 #include "qtcapture.h"
 #import <Carbon/Carbon.h>
 
+// HACK: SetThemeCursor is only available in Carbon.h without LP64 defined.
+//       However it is still present in 64bit system, thus, we just extern it.
+extern "C" OSStatus SetThemeCursor(ThemeCursor inCursor);
```
375gnu: What if it will be removed in the next OSX release? I'd prefer to load if dynamically and only if it's actually present in the system then use it. Otherwise omit calls to SetThemeCursor

eyvallah: it is meant to stay for long, since modern ways to change cursor should actually still map to this function in the Carbon framework although Carbon is deprecated since like 6 years ago if I'm not mistaken. anyway I will edit the hack out and adopt the modern method

eyvallah:
changed to use NSCursor
a few other changes including:
1. use scaling instead of full resolution (the text rendering is not really good in retina 2x screens) 
2. always setNeedsDisplay, rendering will only invoked by system on a regular basis, so don't bother to check needsDispay value
3. Change deployment target to macOS 10.7 (Release in 2011, well Mac users tend to update, so I really doubt if there are any people running pre-10.7 os x and wanting an update of Celestia), so we don't have to deal with C++ ABI.

PR #393, opened 2019-09-06
Make render process more const.

pirogronian:


375gnu:
If change our render routines to store some internal state we will need to remove `const` from all methods or mark all affected members as mutable. i don't think it's a good idea unless really have much compact code. have you tried to check such cases on godbolt.org?

PR #397, opened 2019-09-10
allow specification of config file, data dir in a sandboxed environment

eyvallah:
1. added a window for setting path of config file/data dir
2. remove legacy code for single core CPU, patch for ATI Radeon 9200


eyvallah:
@375gnu this pr should be ready. it added a window allowing user to choose where the "celestia.cfg" and the data directory (current directory) are located. the window appears when user clicked the menu item in menubar or Celestia initialization fails.

PR #402, opened 2019-09-12
making celestia app store ready

eyvallah:
1. the splash image view was disabled before and rendering as an image with alpha, set it to enabled (the nib files are so old that Xcode updated them)
2. use asset catalog in place of image files


eyvallah:
again, it's a mac only PR. I think it's almost ready for App Store distribution. 

things to be done
1. 1024x1024 icon
2. determine a build number
3. compile thrid party libs
4. app store stuff: introduction, etc
5. update credits, author, copyright, etc

PR #409, opened 2019-09-21
remove legacy codes and fix warnings in macosx

eyvallah:


eyvallah:
should be ok mostly are removed codes, replace for deprecation

PR #422, opened 2019-09-30
Merging AstroDB - phase I

pirogronian:


pirogronian:
I'm probably able to fix Eigen-related compile errors in AppVeyor, but lack of pthread is worse issue. Anyone know how to enable multithreading on Win? Or I just have to disable searching for pthread on Win?

375gnu:
i'll check what i can do with threads on win as i have a vm with vs2105

375gnu:
But the main question: have you tested your branch with stars.dat from Celestia Origin?

pirogronian:
No. I once checked, if CO package would fit on my local disk, but nothing more. I still wait for lighter version or exposing it as a repo. Or is a smallet package already avaliable?

375gnu:
Not yet. You can try to download only required files most torrent
clients allow this. Or I can download it and send to you.

pirogronian:
Yes, please.

375gnu:
@pirogronian https://drive.google.com/open?id=1_GFgkcBXYRwoQv6WfOFHLvvKsKa2b5YG

pirogronian:
Thank You. Just tested. No visible errors on standard output nor internal console. With standard luminosity limit sky is considerable more populated and CPU load is higher, but FPS are fine. Any otherts tests suggested?

375gnu:
i'd like to test with larger db, but we don't have any

pirogronian:
@375gnu Apparenlty errors about phtread library wasn't fatal. I consider this first merging stage ready. Could it be merged now or should I commit full integration?

375gnu:
Please give move a few days to run some tests. Unfortunately I'm busy now with 1.6.2 preparation.

pirogronian:
Ok, no problem. Glad to see 1.6.2 progress.

375gnu:
Hi @pirogronian 
I run quick tests* against this PR and #431 as well. Unfortunately numbers are disappointing: this PR consumes the same CPU power (in my case it was 70% of a single Ryzen 2400 core**, constant 60 FPS) but x2.3 times more memory (856Mb vs 369Mb) comparing to master, time to load star.db is <1 s. With #431 situation gets even worth: 100% core load and only 35 FPS *, memory consumption grows to 1067Mb, Celestia needs 5-6 seconds to load star.db.

*star.db from Celestia Origin was used with limiting magnitude set to max (15), 1080p, Earth hidden behind the left window border.

**numbers measured with 3 instances running, with only 1 or 2 instances master and #422 consume 63% CPU, #423 shows FPS up to 40.

astrodb

pirogronian:
It could be prize paid for dynamic content. current database is static and "highly optimised" to once loaded data, according creators themself. But increased CPU usage is a bad and a bit surprising news.

375gnu:
That's the `sudo perf top -p $PID` output with the current database:
master
And here with #431:
astro2

So there are three places to optimize:
* OctreeNode::isInFrustrum()
* processVisibleStars
* internal Octree representation, it seems it's recalculated too often

+ the 4th place: star.db loading

pirogronian:
Octree is recalculated if data is changing. I assume You tested it with static data. Anyway, which optimisation is more urgent: CPU or memory usage? Is high CPU usage confirmed with more than one test?

375gnu:
Both :) I believe memory usage optimization is a easier thing, just return to single precision data. But High CPU usage makes Celestia unusable in certain cases, for example on startup FPS is about 5. Your should investigate why internal rb-tree is rebalanced so often.

pirogronian:
You confuses me with your fear of doubles. 4 additional bytes of double number times, say, 10 millions of objects gives ~40 MB. And do You really think single precision parameters in case of space simulation program is a good thing?
Cpu is probably affected by iterator of tree. I have some idea about it. Perhaps would decrease memory usage as well.

375gnu:
I fear because it's too naive and straightforward approach. There are a lot of objects which don't need doubles, at least solar system objects. And now we have not +80Mb but +700Mb, almost order of magnitude growth.

pirogronian:
So it's clear indication that this grow is _not_ caused by doubles. Moreover, singles could work for bodies and locations lical to nerby star, but stars and bodies not associated with stars (there was a request to support this type of object) should have doubles. So, will we support both precisions (if so, then how) to save some memory?

PR #431, opened 2019-10-05
[WIP] Final AstroDB merge

pirogronian:


pirogronian:
Performance of an octree seems to be considerably improved, but it's visible only if max objects per node is the same (~10), which, in turn, is responsible (at least partially) for big memory consumption. This number can be enlarged, which will save some memory, but also will increase CPU load (and startup time). So, I'll seek another area for memory optimisation.

pirogronian:
[This](http://jsteemann.github.io/blog/2016/06/14/how-much-memory-does-an-stl-container-use/) page suggest the main problem can be used container type. But is there any proper replacmenent for map/set? Unordered versions have only slightly less memory consumption.

pirogronian:
Due to memory consumption issues and major changes in master, I decided to merge AstroDB in yet another PR.

PR #433, opened 2019-10-07
[WIP]support keeping 1.6.1 data while updating

eyvallah:


macosx/Migrator.m
```diff
@@ -0,0 +1,53 @@
...
+        [self migrateDataBaseFrom0To1];
+    }
+}
+
++ (void)migrateDataBaseFrom0To1 {
+    if ([self dataBaseVersion] != 0)
+        return;
+
+    // get the real home path
+    const char *home = getpwuid(getuid())->pw_dir;
```
375gnu: we have HomeDir() here https://github.com/CelestiaProject/Celestia/blob/1.6.x/src/celutil/unixdirectory.cpp#L144

eyvallah: @375gnu HomeDir will return the result of getenv("HOME"), this is the sandboxed home path (which should be ~/Library/Containers/space.celestia.Celestia/Data). Since 1.6.1 is non sandboxed, the preference file is located in the non-sandboxed home path(~), the only way to get this seems to be using getpwuid(getuid())

eyvallah:
this should work however i'll test it tomorrow on a macOS that can run 1.6.1

PR #440, opened 2019-10-11
move languages

eyvallah:
this moves English to en, English UI files to Base. zh_CN to zh-Hans, zh_TW to zh-Hant. and enabled additional warning checking suggested by Xcode.


eyvallah:
Modify the Chinese languages so other Chinese speaking regions (HK, Macau, Singapore, etc) should be able to use either simplified or traditional Chinese. It works on mac, guess it should also work on other platforms? @375gnu

375gnu:
Files from `locale` subdir should work, but not from `po` & `po2`, because they expect standard locale id (lang_TERRITORY.CHARSET@modifier, where only `lang` is mandatory). So even if you want to share files from `locale` you still need to create additional files in both `po` & `po2`. And you can't simply add ad-hock modifier and expect that it will be used by any users because they use standard locales. For example, my system knows only the following chinese locales:
```
# zh_CN GB2312
# zh_CN.GB18030 GB18030
# zh_CN.GBK GBK
# zh_CN.UTF-8 UTF-8
# zh_HK BIG5-HKSCS
# zh_HK.UTF-8 UTF-8
# zh_SG GB2312
# zh_SG.GBK GBK
# zh_SG.UTF-8 UTF-8
# zh_TW BIG5
# zh_TW.EUC-TW EUC-TW
# zh_TW.UTF-8 UTF-8
```

According to https://github.com/CelestiaProject/Celestia/blob/1.6.x/po/translate_resources.pl#L72 on windows the situation is quite the same.

Also I would like not to introduce any features or possibly backward incompatible changes to 1.6.x branch even if they are useful. 1.6.x is just to an update to 1.6.1. For such changes we have 1.7 & `master` branch.

eyvallah:
@375gnu I'll edit this commit, keeping the change of renaming English.lproj to en.lproj. It seems Xcode deprecated zh_CN/zh_TW, so newly created files defaults to use zh-Hans/zh-Hant

eyvallah:
it turns out macOS only supports base internationalization since 10.8, so renaming English.lproj to en.lproj instead of Base.lproj. Since, some new strings were introduced for the config file selection window in 1.6.2, I localized it in simplified Chinese.

375gnu:
> it turns out macOS only supports base internationalization since 10.8,

It was released 7 years ago. Is something older is still relevant?

eyvallah:
@375gnu if we want to use base internationalization, we would have to drop support for 10.7. Since 1.6.2 is a minor fix, I would not do this in this branch.

For 1.7.0, I’ll get my swift port ready, it will be independent of the current macOS project, so there shouldn’t be a need to worry about maintaining changes from 1.6.2

375gnu:
okay. np.

PR #443, opened 2019-10-12
fix warnings compiling in Xcode

eyvallah:
this commit:

1. fixes all warnings when compiling in Xcode 11 (except for suggestion for base internationalization)
2. remove the full screen handling code since fullscreen is supported by Mac natively now
3. always use libjpeg and libpng for image loading. since macOS uses libjpeg internally, it shouldn't be a problem


eyvallah:
@375gnu you can check the last files on the list since the others are just objective-c++ or UI :)

macosx/CGLInfo.m
```diff
@@ -27,7 +27,7 @@ static BOOL ExtensionSupported(const char *ext)
         return false;
 
     char *extensions = gExtensions;
-    int len = strlen(ext);
+    unsigned long len = strlen(ext);
```
375gnu: `size_t` is more correct here

PR #445, opened 2019-10-13
fix warnings in master

eyvallah:


src/celengine/rectangle.h
```diff
@@ -25,7 +25,7 @@ class Rect
         Filled      = 0x0002,
     };
 
-    Rect() = default;
+    Rect() = delete;
```
375gnu: yep, as i remember some compilers tend to remove it implicitly

PR #453, opened 2019-10-18
[win32] Remove copy of CelestiaCore::runScript from win frontend

375gnu:


375gnu:
@pirogronian it works, i have checked. MessageBox is called by WinAlerter so generally nothing changes. But this PR is required for #455

PR #454, opened 2019-10-18
[WIP] (More) dynamic unified runtime database

pirogronian:


pirogronian:
According to e=most recent debug info, all stars, DSOs, bodies, octree nodes and cross index records have average size ~ 330 MB:

> AstroDatabase::dumpStats()
>  Objects: 2461185 (* 56 B = 131.441467 MB)
>  Stars: 2449017 (* 104 B = 242.898712 MB)
>  DSOs: 11087 (* 160 B = 1.691742 MB)
>  Bodies: 1081 (* 352 B = 0.362885 MB)
>  Names: 43493
>  Localized names: 3
>  Octree nodes: 496328 (* 152 B = 71.946960 MB)
>  CrossIndex records: 400386 (* 16 B = 6.109406 MB)

So, high memory consumption os probably due to some other structures, probably indexes.

CMakeLists.txt
```diff
@@ -285,6 +285,8 @@ if(OCTREE_DEBUG)
   add_definitions(-DOCTREE_DEBUG)
 endif()
 
+include(FindThreads)
```
375gnu: why did you add this check? `std::thread` is a c++11 feature available on all platforms. While this check look for posix threads.

src/celengine/astrooctree.cpp
```diff
@@ -0,0 +1,529 @@
...
+        return false;
+    }
+    OctreeNode *node = this;
+    while(node->m_objects.size() < node->m_maxObjCount && node->getChildrenCount() > 0)
+    {
+//         fmt::fprintf(cout, "Node %i[%i/ %i]->%i\n", node, node->m_objects.size(), node->m_maxObjCount, node->getChildrenCount());
+        int i = node->getBrightestChildId();
+        if (i < 0)
+        {
+            fmt::fprintf(clog, "No brightest child of node %i!!\n", node);
```
375gnu: `node` is a pointer. why do you use `%i`?

375gnu:
now performance is better, 50fps but still 1 cpu core is 100% loaded

astrodb

375gnu:
That's default DB:

orig

CMakeLists.txt
pirogronian: AFAIR without it linker throwed error.

src/celengine/astrooctree.cpp
pirogronian: Indeed. I just didn't pay much attention.

pirogronian:
Not good. On my machine results was as good as with standard DB and I hoped this problem is in general solved. And while looking into processVisibleStars I cant see any places to optimise.

375gnu:
astrodb

pirogronian:
Good to know, however now I'm convinced that dynamic octree cannot be considerably better tuned. So I decided to recreate static octree for non-moving objects. But dont know, when it will be done...

375gnu:
btw, i stil think that making db using doubles only is not a great idea. first, doubles are slower than floats, some lowpower hardware lacks their support (on them either double is just an alias for float or implemented in software).

as we are planning plugable modules, we can provide 2 implementations: with doubles and floats.

PR #455, opened 2019-10-18
Scripting refactor

375gnu:
New wrappers don't replace existing scripting classes but complement them so we have a consistent interface.

We still have hardcoded legacy and lua support because we don't have a plugin infrastructure yet, but it's much easier to adapt new code for new languages and and implement the as plugins.


375gnu:
stupid m$v$: locally it compiles without eef670c, in appveyor 50/50

Issue #463, opened 2019-10-20
Pluto's orbit


SevenSpheres:
In Celestia 1.7 Ceres, Haumea, Makemake, and Eris have purple orbits, but Pluto-Charon has a blue orbit (same as major planets). This is because Pluto-Charon is defined as a reference point, so the orbit color can't be specified. I see three ways to fix this:
1. Define the Pluto-Charon barycenter as an object instead of a reference point (what Celestia Origin does)
2. Specify the color of dwarf planet orbits in the start script (what Project Echoes does)
3. Change the color of dwarf planet orbits in the source code

I prefer option 2 because it means the user can easily change orbit colors.

SevenSpheres:
I implemented option 2 and commited it to my dwarfplanets-test branch, but it's not showing up.

375gnu:
4. check that a reference point is a star's child, get its children and assign orbit's colour of the refpoint according to its children type.

SevenSpheres:
**375gnu**, that would still require changing the default color of dwarf planet orbits.

375gnu:
@SevenSpheres why?

SevenSpheres:
The default color for dwarf planet orbits is blue. Dwarf planets are defined with purple orbits in dwarfplanets.ssc (which makes sense, so we can tell dwarf planets from major planets), but it's impossible to change the orbit color for Pluto-Charon because it's a reference point. I listed three ways this can be fixed.

I think you were saying that the orbit color of a reference point (i.e. Pluto-Charon) should match that of its first child (i.e. Pluto). In that case the default orbit color wouldn't need to change.

I still think the default orbit color should change, preferably in the start script and not in the source code. Then it won't need to be specified in dwarfplanets.ssc.

SevenSpheres:
Okay, you did my option 3. Maybe the addition to the start script could be in an addon?

375gnu:
I don't think that start script is a good place for such changes. Actually it's worth to add new configuration options to allow control over orbit colors and other things.

SevenSpheres:
Now I notice that Pluto-Charon has no label (but maybe this has been fixed now?)

375gnu:
No, it's not fixed. I'll take care of it.

PR #466, opened 2019-10-21
Override refPoint orbit color according to a child type (Closes: #463)

375gnu:
i believe there is a better way to fix this, so i'll won't merge it now


375gnu:
#471 provides better implementation

PR #476, opened 2019-10-25
Resolve name conflict with Atlas

SevenSpheres:
There are two objects named Atlas; a star and a moon. I renamed the moon to "Atlas (moon)".

Maybe I should just update all the data files, since Askaniy seems to have disappeared...

### DIFF
--- a/data/minormoons.ssc
+++ b/data/minormoons.ssc
@@ -576,7 +576,7 @@
     Albedo 0.5
 }
 
-"Atlas:Saturn XV:1980 S 28" "Sol/Saturn"   # Cassini-epoch data
+"Atlas (moon):Saturn XV:1980 S 28" "Sol/Saturn"   # Cassini-epoch data
 {
     Class "minormoon"
     Texture     "asteroid.jpg"


data/dwarfplanets.ssc
```diff
@@ -37,8 +37,6 @@ ReferencePoint "Pluto-Charon" "Sol"
 {
     CustomOrbit "pluto"
 
-    OrbitColor  [ 0.556863 0.235294 0.576471 ]
```
375gnu: why did you remove this color?

SevenSpheres: Since the orbit color for dwarf planets is now specified in the source code, that line is no longer needed

SevenSpheres: Added it back (I can use rebase now!)

pirogronian:
I think this shpuld be resolved by namespace. Altlas the star is accessible from main namespace, while Atlas the moon is accessible from Saturn namespace, I mean to find from console we have to enter "Saturn/" first. So, technically, there is no reason for change.

SevenSpheres:
> I mean to find from console we have to enter "Saturn/" first.

Yes, this is _because_ there is a star with the same name.

375gnu:
Pirogronian pointed a better way, so I agree, no need to change.

PR #481, opened 2019-10-28
update .travis.yml

eyvallah:
make macOS non-failable, and update Xcode

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 dist: xenial
-osx_image: xcode10.2
+osx_image: xcode11.1
 language: cpp
 sudo: true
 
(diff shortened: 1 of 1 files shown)


eyvallah:
urgh, Xcode 11 tries to use the c++ filesystem which is only supported on macOS 10.15, however CI is running in macOS 10.14, that's why the macOS build is throwing errors. I'll see to it later.

eyvallah:
it still defaults to C++ 17 when not linking with c++fs, i think we should introduce a macro to disable use of c++ filesystem in filesystem.h rather than relying totally on cplusplus version @375gnu what do you think?

375gnu:
Why apple clang uses C++17 despite C++11 set by cmake?

eyvallah:
I'm closing this until travis offers 10.15 machine. the glew package fetched from brew is flawed in cmake config, and will make the linker try to link to OpenGL library in a non-existent path.

PR #494, opened 2019-11-05
Remove duplicate moons

SevenSpheres:
Remove duplicate Nix and Hydra


SevenSpheres:
> explain why you need to change load order

I think it makes more sense to have dwarf planets before asteroids in the solar system browser.

SevenSpheres:
> explain why you need to change load order

I think it makes more sense to have dwarf planets before asteroids in the solar system browser.

375gnu:
@SevenSpheres it doesn't matter in what order you load files

PR #503, opened 2019-11-11
Increase star render distance

SevenSpheres:
Based on [this thread](https://celestia.space/forum/viewtopic.php?f=3&t=17783) and [this one](https://celestia.space/forum/viewtopic.php?f=23&t=20030&start=20), I've increased the maximum star render distance to 1 billion light-years. Now Celestia should be able to render M87*, and other such distant objects.

### DIFF
--- a/src/celengine/stardb.cpp
+++ b/src/celengine/stardb.cpp
@@ -45,7 +45,7 @@ constexpr const char SAOCatalogPrefix[]       = "SAO ";
 // local group of galaxies. A larger value should be OK, but the
 // performance implications for octree traversal still need to be
 // investigated.
-constexpr const float STAR_OCTREE_ROOT_SIZE   = 10000000.0f;
+constexpr const float STAR_OCTREE_ROOT_SIZE   = 1000000000.0f;
 
 constexpr const float STAR_OCTREE_MAGNITUDE   = 6.0f;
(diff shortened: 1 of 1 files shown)


SevenSpheres:
I haven't tested it at all, since I don't have a C++ compiler and [there weren't any artifacts generated](https://ci.appveyor.com/project/375gnu/celestia/builds/28781977/artifacts).

375gnu:
The more distant objects use have the deeper levels of octree present, i.e. more time required to traverse it.

SevenSpheres:
Okay. Janus recompiled Celestia with the same one line change and had no problems (see the first link). You can check for yourself if you want. If there are really problems I can reduce the limit to 100 million light-years; this is still enough for M87*.

375gnu:
He didn't tested it with larger stars.db. I will test it the other day.

LafuenteAstronomy:
Maybe in the long run, you can probably test for around 100 GLY. Janus also made a commit regarding that but I forgot which number

SevenSpheres:
> Maybe in the long run, you can probably test for around 100 GLY star render limit.

It would be easy enough to increase the limit to 100 Gly - just add two more zeroes. However, I don't see the point, since this is larger than the observable universe.

PR #509, opened 2019-11-15
Remove old redundant models

SevenSpheres:


SevenSpheres:
> are you sure that *.3ds files are not sources for appropriate *.cmod files?

They probably are, but do they really need to be there?

Anthony-B-Russo10:
It’s always to have an archive of this stuff done were.

SevenSpheres:
> It’s always to have an archive of this stuff done were.

I honestly have no idea what you're trying to say here.

Anthony-B-Russo10:
The point is keep the files archived some where.

375gnu:
If they are sources then we need to keep them in the repository. We just don't need to install them.

SevenSpheres:
I added back the 3DS models. Now removing only the CMS models.

SevenSpheres:
> We just don't need to install them.

If I understand correctly how CMakeLists.txt works, the redundant models _are_ being installed.

PR #515, opened 2019-11-19
Initial implementation for TrueType fonts support

375gnu:
Not optimized, without RTL support and with old GL.

RTL and modern GL require overlay/console changes, now I just want a drop-in replacement so users can test it.

### DIFF
--- a/src/celttf/truetypefont.cpp
+++ b/src/celttf/truetypefont.cpp
@@ -1,0 +1,628 @@
+// truetypefont.cpp
+//
+// Copyright (C) 2019, Celestia Development Team
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include <config.h>
+#include <algorithm>
+#include <array>
+#include <iostream>
+#include <vector>
+#include <GL/glew.h>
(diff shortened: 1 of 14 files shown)


src/celttf/truetypefont.cpp
```diff
@@ -0,0 +1,628 @@
...
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include <config.h>
+#include <algorithm>
+#include <iostream>
+#include <vector>
+#include <unordered_map>
```
eyvallah: <array> needs to be included too

src/celttf/truetypefont.cpp
```diff
@@ -0,0 +1,628 @@
...
+        if (FT_Init_FreeType(&ft))
+        {
+            cerr << "Could not init freetype library\n";
+            return nullptr;
+        }
+    }
+
+    int psize = 0;
+    auto nameonly = ParseFontName(filename, psize);
+    return TextureFont::load(fs::path(CONFIG_DATA_DIR) / nameonly, psize, dpi);
```
eyvallah: CONFIG_DATA_DIR might not be available for non-qt ones, I think we should just leave it blank, since we already changed current directory to CONFIG_DATA_DIR before performing any operation

PR #520, opened 2019-11-20
GLUtil lib removal, part 1

375gnu:
so far only 2 calls are left: gluErrorString (hardly useful) and gluBuild2DMipmaps which has 2 alternatives but still should left as a fallback, but it can be removed in macos builds


375gnu:
omg, m$vc still has `near` and `far` keywords, like in dos & win3.x era

PR #527, opened 2019-11-22
do not use wordexp on unsupported os

eyvallah:
iOS and Android do not have wordexp.

although we cannot run celestia on these oses now, but give it time, it would happen


eyvallah:
@375gnu I'll try it, but it is a bit different, wordexp.h is available on iOS, but the function wordexp/wordfree is marked unavailable in wordexp.h on iOS/watchOS/tvOS. guess in that way cmake should compile a custom source file to determine if it is really available or not

eyvallah:
indeed, tested with iOS/andoid/macOS cmake toolchain, output is correct

PR #533, opened 2019-11-25
Use shaders to draw object as point

375gnu:


eyvallah:
this seems to have affected halo rendering

screenshot

375gnu:
@eyvallah what do you mean? i don't see any difference.

so far i have another issue.

eyvallah:
screenshot

@375gnu  it used to be brighter

375gnu:
this version is basically the same what we had disabled but with newer api. i don't like it, but this code performance impact is not so important, so we can optimize it later

375gnu:
@eyvallah regarding halo change, the same in this PR. I believe the problem is in star rendering, seems it has some implicit dependencies on other code parts, this present in 1.6.1 as well. Press Ctrl+S to change starfield rendering style and you'll see that Sun changes it's halo. I suppose it's better to create a separate issue for this problem.

375gnu:
ping @pirogronian

PR #548, opened 2019-11-30
Remove deprecated C++ features

375gnu:
std::unary_function & std::binary_function were deprecated in C++11
and removed in C++17


375gnu:
> according to Xcode clang, random_shuffle is also removed in C++17, see galaxy.cpp line 661

okay, will handle this

PR #553, opened 2019-12-01
Refactor render.cpp and extract some classes to own files

375gnu:
render.cpp is a huge beast, it takes time to compile almost like all other files in celengine/. with this change some its parts moved to their own files.

hdrfuncrender.cpp contains unused no functions, so I did not bother to extract them into compilable file. HDR rendering doesn't work now and i doubt we should bother implementing it with gl2.1, proper HDR requires gl3.0 features (framebuffer and floating point textures).

### DIFF
--- a/src/celengine/pointstarrenderer.h
+++ b/src/celengine/pointstarrenderer.h
@@ -1,0 +1,65 @@
+// pointstarrenderer.h
+//
+// Copyright (C) 2001-2019, the Celestia Development Team
+// Original version by Chris Laurel <claurel@gmail.com>
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#pragma once
+
+#include <Eigen/Core>
+#include <vector>
+#include "objectrenderer.h"
+#include "renderlistentry.h"
+
+class ColorTemperatureTable;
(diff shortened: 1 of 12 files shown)


src/celengine/pointstarrenderer.h
```diff
@@ -0,0 +1,57 @@
...
+
+class ColorTemperatureTable;
+class PointStarVertexBuffer;
+class Star;
+class StarDatabase;
+
+class PointStarRenderer : public ObjectRenderer<Star, float>
+{
+ public:
+    static constexpr const float StarDistanceLimit = 1.0e6f;
```
eyvallah: this actually results in undefined reference while linking when using c++11/14, c++17 is fine

375gnu: > this actually results in undefined reference while linking when using c++11/14, c++17 is fine

omg. why all that compilers are so different? g++ eats this fine and even clang (I assume you have clang) from  xcode11.2 does.

eyvallah: this happens while building on Xcode with a Xcode project rather than the cmake generated, I guess there are difference in cxx flag, but that's what I found

https://en.cppreference.com/w/cpp/language/static

If a const non-inline (since C++17) static data member or a constexpr static data member (since C++11) is odr-used, a definition at namespace scope is still required, but it cannot have an initializer. This definition is deprecated for constexpr data members (since C++17).

PR #556, opened 2019-12-03
Refactor asterisms and boundaries drawing

375gnu:
required to make renderer plugins

### DIFF
--- a/src/celengine/asterismrenderer.cpp
+++ b/src/celengine/asterismrenderer.cpp
@@ -1,0 +1,126 @@
+// asterismrenderer.cpp
+//
+// Copyright (C) 2018-2019, the Celestia Development Team
+//
+// This program is free software; you can redistribute it and/or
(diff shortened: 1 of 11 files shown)


src/celengine/asterismrenderer.cpp
```diff
@@ -0,0 +1,126 @@
...
+        m_vo.allocate(m_vtxTotal * 3 * sizeof(GLfloat), vtxBuf);
+        m_vo.setVertices(3, GL_FLOAT, false, 0, 0);
+        delete[] vtxBuf;
+    }
+
+    prog->use();
+    prog->color = defaultColor.toVector4();
+    m_vo.draw(GL_LINES, m_vtxTotal);
+
+    assert(m_asterisms->size() != m_vtxCount.size());
```
eyvallah: I think it should be == rather than !=

375gnu: yes. fixed.

PR #569, opened 2019-12-17
Update INSTALL & README* files

375gnu:


SevenSpheres:
Why list the subreddit? Only, what, three people use it.

PR #571, opened 2019-12-19
Shaders for fonts

375gnu:


src/celttf/truetypefont.cpp
```diff
@@ -73,6 +75,8 @@ struct TextureFontPrivate
     int toPos(wchar_t) const;
     void optimize();
 
+    const Renderer &m_renderer;
```
eyvallah: throws "Explicitly defaulted copy assignment operator is implicitly deleted" warning for the assign operator

PR #585, opened 2020-01-12
Fix red planets + tiny shader optimization

375gnu:
@eyvallah could you check if this fixes red Earth as reported by Selden?


eyvallah:
works fine for me now

375gnu:
yeah, on windows fixed as well

Issue #475, opened 2019-10-25
Additions to celestia.cfg


SevenSpheres:
Some things that could be added to celestia.cfg:

- [ ] SplashImage (see #438)
- [ ] Language (see #441)
- [ ] Label colors for all object types
- [ ] Orbit colors for all object types (see #463)

PR #588, opened 2020-01-18
Increase Star Render Distance FURTHER

LafuenteAstronomy:
I made a small change to the stardb.cpp file, increasing the star render distance from 1 GLY to 100 GLY. Would you mind checking if this works well with the current programming of Celestia? Thanks

### DIFF
--- a/src/celengine/stardb.cpp
+++ b/src/celengine/stardb.cpp
@@ -45,7 +45,7 @@ constexpr const char SAOCatalogPrefix[]       = "SAO ";
 // local group of galaxies. A larger value should be OK, but the
 // performance implications for octree traversal still need to be
 // investigated.
-constexpr const float STAR_OCTREE_ROOT_SIZE   = 1000000000.0f;
+constexpr const float STAR_OCTREE_ROOT_SIZE   = 100000000000.0f;
 
 constexpr const float STAR_OCTREE_MAGNITUDE   = 6.0f;
 //constexpr const float STAR_EXTRA_ROOM        = 0.01f; // Reserve 1% capacity for extra stars


375gnu:
This actually require a star database with so distant data, without it the change is useless.

GurrenLagannTSS:
Vision's addons do place real black holes that are beyond the current 1 Gly render limit, so this change might be needed If we want then to work in standard Celestia.

SevenSpheres:
All of Vision's addons released so far work with Janus' version of Celestia 1.7.0, which has exactly the same code change I made (1 Gly limit). I agree with @375gnu here; no one is going to define anything this distant, since it's larger than the observable Universe.

LafuenteAstronomy:
Then, I guess we need to probably merge the code with Janus' code so that GitHub version works as well. And while 100 billion may be beyond the Observable Universe, it may be the best that we can do, since the actual diameter is 93 billion, just 7 billion less.

LafuenteAstronomy:
If only Janus was here to give his insight

SevenSpheres:
> And while 100 billion may be beyond the Observable Universe, it may be the best that we can do, since the actual diameter is 93 billion, just 7 billion less.

The **radius** (which is what matters here) is 46.5 Gly, less than half of 100 Gly.

LafuenteAstronomy:
Perhaps 50 GLY can be made instead. However, for reasons of fictional addons, 100 can be good, it's the call of the developer if it can be tierh 50, 100 or unchanged.. We just need to integrate it with Janus' code, which enables such objects to be present in the first place.

SevenSpheres:
> We just need to integrate it with Janus' code, which enables such objects to be present in the first place.

I guess the Janus build I have increases the star render limit further than 1 Gly.

LafuenteAstronomy:
> > We just need to integrate it with Janus' code, which enables such objects to be present in the first place.
> 
> I guess the Janus build I have increases the star render limit further than 1 Gly.

If I remember, he had a 100 GLY commit. It may be the one you are using

375gnu:
For me render distance should be configured in celestia.cfg rather than hardcoded.

SevenSpheres:
> For me render distance should be configured in celestia.cfg rather than hardcoded.

[Agreed](https://github.com/CelestiaProject/Celestia/issues/475)

PR #591, opened 2020-01-21
Fix and enable structs in shaders

375gnu:
need to test this first on older hw


375gnu:
okay, as there is no reports of something broken let's merge it

PR #597, opened 2020-02-02
EHT-like zoom limit

GurrenLagannTSS:


375gnu:
what is eht, and why should we have this limit changed?

GurrenLagannTSS:
What is EHT - Event Horizon Telescope

Why we should change it - With it, not only we could see other black holes and/or their distant galaxies from other distant galaxies (and Earth), we could use it to simulate observations exoplanets from light-years away (eg. Proxima Centauri b and c, from Earth).

SevenSpheres:
Hmm good idea, but there would also need to be some changes to the magnitude limit:

magnitudelimit

Anthony-B-Russo10:
Still a good idea though.

PR #603, opened 2020-02-10
Fix missing header (GCC 4)

h3xx:
`clog` is no longer a built-in and requires iostream.

This bug was introduced in 4a38917e08a83314ca9d2cb8af3c0562acd29fb2.


h3xx:
This bug was introduced in 4a38917e08a83314ca9d2cb8af3c0562acd29fb2.

Issue #602, opened 2020-02-10
`BlendTexture` property is ignored


375gnu:
This apparently affects Celestia Origin's rogue planet addon as well (`BlendTexture` seems to be broken):

![rogueplanetbug](https://user-images.githubusercontent.com/56284207/74109301-5471fa00-4b3f-11ea-9315-ef1d714dbecc.png)

@375gnu I think this should be labeled "high priority".

_Originally posted by @SevenSpheres in https://github.com/CelestiaProject/Celestia/issues/541#issuecomment-583889172_

SevenSpheres:
> @375gnu I think this should be labeled "high priority".

To be clear this was referring to the star rendering issue (#541), not the `BlendTexture` issue.

Also, `BlendTexture` still works correctly for objects in the Solar System (e.g. Sedna). That's why I assumed this was the same as the star rendering issue.

375gnu:
@SevenSpheres could you test https://ci.appveyor.com/project/375gnu/celestia/builds/30729994/artifacts?

SevenSpheres:
That fixes it!

PR #607, opened 2020-02-11
Fix BlendTexture property (Fixes: #602)

375gnu:


SevenSpheres:
Confirming that this fixes the bug.

SevenSpheres:
This does indeed fix both #602 and #662, but it seems to make `BlendTexture` work incorrectly for Solar System bodies:
Sedna before this commit:
sedna-nobug
Sedna after this commit:
sednabug

375gnu:
I have reverted the change which influenced Sedna, though i might apply other changes.

375gnu:
but i have found another issues, so make this WIP for a while

375gnu:
now everything should be fine

PR #608, opened 2020-02-11
stars: don't allow incorrect semiAxes

375gnu:
all < 1 or at least one > 1


SevenSpheres:
[Pretty sure the consensus is against this.](https://celestia.space/forum/viewtopic.php?p=147985#p147985)

375gnu:
I strongly oppose idea that Celestia code should have fixes and workarounds for incorrect addons. So, taking into account that the community is against the proposed change, it won't be implemented.

PR #610, opened 2020-02-13
Allow building without NLS

375gnu:
Since the introduction of CMake we had `ENABLE_NLS` option but it was never properly implemented. It basically useless for end-users but I have a mingw & vcpkg based crosscompiling environment and it lacks gettext.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -48,7 +48,6 @@ init:
 
 install:
 - cd c:\tools\vcpkg
-- git pull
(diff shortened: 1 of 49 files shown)


SevenSpheres:
giterror
something was done incorrectly here

eyvallah:
what is NLS short for?

375gnu:
> something was done incorrectly here

i started the change on another computer, where i haven't setup git correctly

> what is NLS short for?

Native Language Support

Issue #526, opened 2019-11-21
Can't type some characters in search


SevenSpheres:
In Celestia 1.6.1, when the search box is open (by pressing Enter), pressing any key will type into the search box. In Celestia 1.7, it's impossible to type brackets in the search box; regardless of whether the search box is open, typing brackets will always adjust the limiting magnitude. This may also be the case for other characters; I haven't tried pressing every key.

This actually is a problem - in some catalogs of objects, particularly those derived from SIMBAD, there are object names with brackets.

375gnu:
@SevenSpheres i suppose the problem is with Qt interface only, isn't it?

SevenSpheres:
> i suppose the problem is with Qt interface only, isn't it?

Yes it is, I never noticed that (since I'm not checking everything with both frontends).

PR #613, opened 2020-02-15
[qt] Pass '[' and ']' keys to search console

375gnu:
Closes: #526


375gnu:
@SevenSpheres please test

SevenSpheres:
Fixes brackets at least, though there may or may not be other keys that don't work.

375gnu:
I don't see any other keys without modifiers used by Qt GUI

LukeCEL:
This fixes brackets. I haven't found any other keys that fail.

PR #619, opened 2020-02-19
[win32] remove uneeded menu separators

375gnu:


375gnu:
> looks good

if you are ok with the change please press "approve" instead of simply commenting

PR #620, opened 2020-02-20
WIP remove wglex

375gnu:


375gnu:
@SevenSpheres could you test this PR,  especially switching to/from full screen?

SevenSpheres:
I don't typically use fullscreen mode so I don't know what the behavior is in other builds, but when I switch to fullscreen in this build, the screen resolution becomes ridiculously low such that I can only see part of the screen. Attempting to exit fullscreen mode does nothing; I could only reset the screen resolution by logging out (through the CTRL+ALT+DELETE window).

375gnu:
ok, let's leave it for future and somebody good in windows

Issue #628, opened 2020-02-28
Browser doesn't sort numerically


LukeCEL:
Both the 1.6.1 browser and the 1.7 QT Celestial Browser treat numbers as if they should be sorted alphabetically. For example, instead of this putting these objects in this order:
- NGC 1
- NGC 2
- NGC 3
- NGC 4
- NGC 5

etc., they put objects in this order:
- NGC 1
- NGC 10
- NGC 100
- NGC 1000
- NGC 1001

SevenSpheres:
I mean, this isn't really an issue for me. Also, why is this tagged as "qt", since it isn't Qt-specific?

LukeCEL:
> Also, why is this tagged as "qt", since it isn't Qt-specific?

Good point

LukeCEL:
Update: @eyvallah fixed the bug for 1.7, now it only exists in 1.6.x.

SevenSpheres:
Why is this issue still open, since this has been done for the Qt, native Win, Android, and presumably macOS/iOS frontends? I see there's a "gtk" label, so has this not been done for the Gtk frontend?

375gnu:
No, gtk and native win interfaces were not updated (and i doubt they will be).

SevenSpheres:
This is the Windows native frontend.
![nativewinsort](https://user-images.githubusercontent.com/56284207/144753969-47ce6e83-04d6-4043-8f81-2e47a6dc59cd.png)

375gnu:
@SevenSpheres that's not browser. That's a search console.

SevenSpheres:
Well...
* In 1.6.x on Windows the search console sorts alphabetically, so this has been changed
* The Windows native frontend has no DSO browser, and the star browser apparently can't be sorted by name at all

375gnu:
Let's keep it open for a while until we decide the future of win32 & gtk frontends.

375gnu:
gtk was removed, win-native is doomed

PR #630, opened 2020-02-29
Natural sort based on numeric values in QT browsers

eyvallah:
fix sort in qt #628


eyvallah:
> Doesn't work, still looks like this:
> ngcsearch

only in qt browsers, not console.

LukeCEL:
It seems to work.

naturalsort

PR #639, opened 2020-03-03
Extract FramebufferObject to its own file

375gnu:
just for flexibility.


eyvallah:
btw I saw that this function is not used anywhere, renderGeometryShadow_GLSL

hv-work:
@eyvallah it's required for selfshadowing and for postprocessing effects like bloom/hdr and probably for spherical mirror projection

Issue #637, opened 2020-03-02
Wrong URLs cause Celestia to crash


eyvallah:
**Describe the bug**
If a wrong URL is entered into Celestia, Celestia crash immediately

**To Reproduce**
1. copy "cel://testurl"
2. Control(Command)+V paste into Celestia
3. Crash

**Expected behavior**
An alert containing error message should be displayed instead

**Desktop (please complete the following information):**
 - OS: macOS
 - Frontend: QT
 - Version: 1.7.0

**Additional context**
libc++abi.dylib: terminating with uncaught exception of type std::invalid_argument: stof: no conversion

PR #642, opened 2020-03-05
Fix failure if an incorrect URL is pasted

375gnu:
Closes: #637

+ small code cleanup

the fix part will be backported to 1.6.2

### DIFF
--- a/src/celengine/vertexobject.cpp
+++ b/src/celengine/vertexobject.cpp
@@ -29,10 +29,11 @@ VertexObject::~VertexObject()
 {
     delete m_attribParams;
 
-    if (GLEW_ARB_vertex_array_object)
+    if (m_vaoId != 0 && GLEW_ARB_vertex_array_object)
         glDeleteVertexArrays(1, &m_vaoId);
 
-    glDeleteBuffers(1, &m_vboId);
+    if (m_vboId != 0)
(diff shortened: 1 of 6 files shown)


SevenSpheres:
This fixes Celestia crashing, but when an invalid URL is pasted Celestia should show an error message. Instead nothing happens.

eyvallah:
I think a boolean value/error string could be returned when opening a url. or invoke an error in Alerter?

hv-work:
return type of `Url::goTo` should be changed to `bool` from `void`, the same for `CelestiaCore::goToUrl`, other is easy as well

eyvallah:
still crashes with cel://testurl, stof throws exception

375gnu:
@eyvallah i'm unable to reproduce your crash, could you build in debug mode and run under debugger to find the fault location?

375gnu:
okay, now it should catch more errors

not the best solution, but as i said once this should be rewritten from scratch, this is the ugliest part of Celestia. both from design view and implementation

eyvallah:
url.cpp, line 516

375gnu:
is it with the latest changes?

eyvallah:
> is it with the latest changes?

my bad, forgot to run make install

PR #643, opened 2020-03-05
Fix particles rendering

### LINKED ISSUE #549: Stars becoming yellow
@375gnu: remember the time I referenced this problem on the Celestia forums [here](https://celestia.space/forum/viewtopic.php?p=142395#p142395)? Well, I think I found the problem. The stars turn yellow as soon as a model with point sprites is loaded.

Here's a screenshot before the model is loaded:
beforesprites

Here's a screenshot after the model is loaded:
aftersprites

I have confirmed that the stars turn yellow right when the console displays:
> Loading model: "extras-standard/Nebula - Sprites/models/hltau.cmod"

375gnu:
This potentially could fix #549 (yellow stars). Or it could break everything. @LukeCEL, I know you have a lot of sprite models. Could you test them with this change?

### DIFF
--- a/src/celengine/rendcontext.cpp
+++ b/src/celengine/rendcontext.cpp
@@ -166,7 +166,6 @@ RenderContext::drawGroup(const Mesh::PrimitiveGroup& group)
     {
         glEnable(GL_POINT_SPRITE);
         glActiveTexture(GL_TEXTURE0);
-        glTexEnvi(GL_POINT_SPRITE, GL_COORD_REPLACE, GL_TRUE);
         glEnable(GL_VERTEX_PROGRAM_POINT_SIZE);
     }
 
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -2236,7 +2236,7 @@ ShaderManager::buildFragmentShader(const ShaderProperties& props)
     if (props.texUsage & ShaderProperties::DiffuseTexture)
     {
         if (props.texUsage & ShaderProperties::PointSprite)
-            source += "color = texture2D(diffTex, gl_TexCoord[0].st);\n";
+            source += "color = texture2D(diffTex, gl_PointCoord);\n";
         else
             source += "color = texture2D(diffTex, " + diffTexCoord + ".st);\n";
     }
@@ -2811,7 +2811,10 @@ ShaderManager::buildEmissiveFragmentShader(const ShaderProperties& props)
 
     if (props.texUsage & ShaderProperties::DiffuseTexture)
     {
-        source += "    gl_FragColor = " + colorSource + " * texture2D(diffTex, gl_TexCoord[0].st);\n";
+        if (props.texUsage & ShaderProperties::PointSprite)
+            source += "    gl_FragColor = " + colorSource + " * texture2D(diffTex, gl_PointCoord);\n";
+        else
+            source += "    gl_FragColor = " + colorSource + " * texture2D(diffTex, gl_TexCoord[0].st);\n";
     }
     else
     {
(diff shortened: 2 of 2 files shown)


LukeCEL:
Hi @375gnu - initial results look promising. I visited HL Tau in my spritestest repository and the stars didn’t turn yellow!

However, I discovered a few other bugs involving sprites. Because I haven’t tested them on another branch, I don’t know if they were caused by this PR or were there originally in the 1.7 branch. I will do some tests later.

Additionally, I found a bug that has been in Celestia since at least 1.6.1. i plan to create a new bug report.

375gnu:
> I visited HL Tau in my spritestest repository and the stars didn’t turn yellow!

At least my guess was correct

LukeCEL:
> However, I discovered a few other bugs involving sprites. Because I haven’t tested them on another branch, I don’t know if they were caused by this PR or were there originally in the 1.7 branch. I will do some tests later.

Okay, there is some strange behavior going on with the sprites. In 1.6.1 and 1.6.2, my spritestest objects display fine. This is true regardless of whether the textures are in the "medres" or "hires" directory.

In 1.7.0 (the latest commit), the spritestest objects display fine when the textures are in "medres". (The stars turn yellow, but that's besides the point.) However, when the textures are in "hires", the object "Disk Test" does not render with textures. It looks like this:


It was for this reason that I recently changed "hires" to "medres" in my spritestest repository.

In this branch when the textures are in "medres", HL Tau's Disk displays fine. However, it simply doesn't show anything for "Disk Test". When the textures are in "hires", HL Tau's Disk displays fine but "Disk Test" looks like the picture above.

TL;DR: I think this branch made some sprite models not work when the texture is in "medres". It also doesn't work when it's in "hires", but that was an issue from before.

Note: I have not comprehensively tested this with Vision's sprites, but it appears to be affected by the same problem, to some extent.

hv-work:
For me `hires` works fine:

spritetes-hires

hv-work:
@LukeCEL could you provide `/tmp/celestia-shaders.log` for both cases?

hv-work:
okay, the problem is clear with DSO object only medres textures are looked for

LukeCEL:
> other problems are still obscure for me

Yeah sorry haha, I kind of just dumped all the results from my testing in a single comment. Basically I think there are two issues going on here:

Problem 1: DSC objects with hires textures don't show the texture. I agree with what you said, "with DSO object only medres textures are looked for."

Problem 2: In this branch "fix/glsl-particles-drawing", DSC objects with medres textures don't render. When I say:
> it simply doesn't show anything for "Disk Test"
I mean to say that "Disk Test" shows nothing at all, as if no model was there.

375gnu:
Found why only medres is loaded with DSC objects.

In your SSC you have texture defined:
```
"Disk" "HL Tau"	
{
	Class "diffuse"
	Mesh "hltau.cmod"
	Texture "27px.png"
```
If you remove it you'll have the same issue.

The problem is when Celestia loads models it loads _medres_ texture _only_. If you override texture in ssc files then this overridden texture is used instead, and for it Celestia uses `MultiResTexture` class to resolve to a proper resolution.

375gnu:
"obscure" in this context means that i don't understand how this change can interfere with texture loading, it affects only shaders.

LukeCEL:
I'm recompiling Celestia right now, but how do I get /tmp/celestia-shaders.log?

375gnu:
shader log says that no geometry was loaded at all

hv-work:
@LukeCEL  could you also provide error console output?

LukeCEL:
> error console output

What do you mean by this?

hv-work:
@LukeCEL press `~` (`Shift`+`)

LukeCEL:
@375gnu Oh my god... I tested this again with my build (commit 26aa3d35) and I realized that I had turned off showing nebulae in the "Preferences" menu! I am sorry for confusing everybody here.

So I have tested the latest commit (26aa3d3) and everything seems to work. (Except when the texture is in not in the "medres" folder, but that's already covered in #649.)

Issue #467, opened 2019-10-21
Spelling error


SevenSpheres:
The 1.7 preferences dialog says "Spacecrafts". It should be "Spacecraft".

SevenSpheres:
**375gnu**, this is not a bug, it is a simple typo.

375gnu:
@SevenSpheres why do you think it should be in singular here? All other objects are in plural as well.

SevenSpheres:
The plural of spacecraft is the same, like with aircraft. Spacecrafts is not a word.

SevenSpheres:
Some related things that don't need their own issues:
- "Sync Orbit" was changed to "Synch Orbit". This is fine, but why the change?
- The console warning about deprecated parameters is missing a line break after "Use GeomAlbedo instead."

hv-work:
>     * "Sync Orbit" was changed to "Synch Orbit". This is fine, but why the change?

@SevenSpheres where did you find this?

SevenSpheres:
> where did you find this?

It's in the right-click menu for any SSC object.

375gnu:
Okay, found it. I wouldn't say "it was changed". Qt interface is based on neglected KDE interface which had "Synch" from the very beginning. Windows is not the Center of the World :)

SevenSpheres:
> Qt interface is based on neglected KDE interface which had "Synch" from the very beginning.

Okay, I wasn't aware of this. Also, is @hv-work an alternate account of @375gnu?

375gnu:
yes, that's me

375gnu:
Yes, we have fixed it. Thanks for pointing out.

RacerBG:
My God, I'm an astronomy lover since 20 years and only now I learned that the plural of spacecraft is "spacecraft"... :smile:

PR #645, opened 2020-03-05
Add newline in solarsys.cpp

SevenSpheres:
As mentioned in #467


hv-work:
As you are changing this, could you also add '\n' to the message on the line 824:
```
            fmt::fprintf(cerr, "Incorrect BondAlbedo value: %lf", t);
```

SevenSpheres:
@hv-work Done.

PR #651, opened 2020-03-06
Common auto-indexing for AstroObject.

pirogronian:


src/celengine/astroobj.h
```diff
@@ -21,7 +21,15 @@ class AstroObject
     AstroCatalog::IndexNumber getIndex() const { return m_mainIndexNumber; }
     void setIndex(AstroCatalog::IndexNumber);
 
+    static constexpr AstroCatalog::IndexNumber MaxAutoIndex = UINT32_MAX - 1;
+    static AstroCatalog::IndexNumber autoIndex(bool decrement = true);
```
375gnu: instead of adding function with flag, introduce 2 functions:
`getNextIndex` and `getCurrentIndex`

they should be `inline`

src/celengine/stardb.cpp
```diff
@@ -1360,6 +1352,7 @@ bool StarDatabase::load(istream& in, const fs::path& resourcePath)
                     startPos = next;
                 }
             }
+            AstroObject::autoIndex();
```
375gnu: this is unacceptable, you call the method just for its side effects

pirogronian: Ok. So, separate method just for update intetnal counter?

375gnu: > Ok. So, separate method just for update intetnal counter?

nope. you have a function which returns counter and then decrements it. why not to use it in 2 cases above?

src/celengine/astroobj.h
375gnu: maybe getNextIndex is not a correct name, maybe better getCurrentIndexAndUpdate

src/celengine/stardb.cpp
pirogronian: To save number of used indexes in case if creation of an object would fail. Maybe not a usual case, but anyway...

375gnu: u have 4 billions of indexes...

pirogronian: Now, but who knows the future? And there is possible memory saving, if I implement index map with "slabs". I added "Wip" for a reason...

src/celengine/astroobj.h
```diff
@@ -21,7 +21,22 @@ class AstroObject
...
+    static constexpr AstroCatalog::IndexNumber MaxAutoIndex = UINT32_MAX - 1;
+    static AstroCatalog::IndexNumber getNewAutoIndex() { return m_autoIndex--; }
+    static AstroCatalog::IndexNumber getCurrentAutoIndex() { return m_autoIndex; }
+    static void recoverAutoIndex()
+    {
+        if (m_autoIndex < MaxAutoIndex)
+            ++m_autoIndex;
+    }
+    AstroCatalog::IndexNumber setNewAutoIndex();
+    static void resetAutoIndex(AstroCatalog::IndexNumber i = MaxAutoIndex) { m_autoIndex = i; }
```
375gnu: `reset` means `return to initial state`

src/celengine/astroobj.h
```diff
@@ -21,7 +21,22 @@ class AstroObject
     AstroCatalog::IndexNumber getIndex() const { return m_mainIndexNumber; }
     void setIndex(AstroCatalog::IndexNumber);
 
+//  Auto Indexing stuff
+    static constexpr AstroCatalog::IndexNumber MaxAutoIndex = UINT32_MAX - 1;
+    static AstroCatalog::IndexNumber getNewAutoIndex() { return m_autoIndex--; }
+    static AstroCatalog::IndexNumber getCurrentAutoIndex() { return m_autoIndex; }
```
375gnu: why `getNewAutoIndex` and `getCurrentAutoIndex` return the same value?

pirogronian: Hmm..  Indeed. I wanted avoid long names like `getAutoIndexAndUpdate` but apparently used misleading names.  Feel a bit tired and cant find a good ones...

src/celengine/astroobj.h
pirogronian: I though it's good enaugh and dont interferes with `setAutoIndex`. Again, I've trouble with finding very accurate and concrete names.

src/celengine/astroobj.h
375gnu: i'd prefer long names

pirogronian:
Actually these code is used (in star & dso databases), but just doesn't give much benefits for now. But yes, I everything other is ok, I can close it in favour of #653 , which is based on it.

375gnu:
yes, other looks ok, please add it as part of #653

PR #653, opened 2020-03-10
Common main index handling.

pirogronian:
Currently, Celestia has separate indexing mechanism for stars, dsos and other bodies. This strongly limits possible use of catalogues consisting objects from various categories (and duplicates infrastructure). Moreover, current impelemntation of object index, single sorted vector, have performance limitation in case of large number of objects.

So, this PR provides unified main index infrastructure, enabling cross-indexing for all types of objects. Furthermore, single sorted vector is replaced by array of vectors, which should greately reduce lookup time (and load time on case of really large number of objects, for example > 10000000). 

I think I found reasonable implementation. At least tests with big `stars.dat` from CO showed no considerable increase in startup time and memory usage. Should be fine until `IndexNumber` remains 32 bit... Later some sort of smart hash table will be needed.


src/celengine/astroobj.cpp
```diff
@@ -5,18 +5,61 @@
 #include "astroobj.h"
 #include "category.h"
 
+AstroObject::AstroObject(AstroObject &o)
+{
+    setIndex(o.getIndex());
```
hv-work: you don't need wrapper functions here
just `m_index = o.m_index`

src/celengine/astroobj.cpp
```diff
@@ -5,18 +5,61 @@
...
+        for (auto *cat : *cats)
+            addToCategory(cat);
+}
+
+AstroObject::AstroObject(AstroObject &&o)
+{
+    setIndex(o.getIndex());
+
+    auto cats = o.getCategories();
+    if (cats != nullptr)
```
hv-work: and this block is redundant, all you need is `m_categories = o.m_categories; o.m_categories = nullptr;`

src/celengine/astroobj.cpp
```diff
@@ -5,18 +5,61 @@
...
+        for (auto *cat : *cats)
+            addToCategory(cat);
+}
+
+AstroObject::AstroObject(AstroObject &&o)
+{
+    setIndex(o.getIndex());
```
hv-work: o.m_index = 0;

src/celengine/dsooctree.cpp
```diff
@@ -43,7 +43,7 @@ double dsoAbsoluteMagnitudeDecayFunction(const double excludingFactor)
 
 
 template <>
-DynamicDSOOctree* DynamicDSOOctree::getChild(DeepSkyObject* const & _obj, const PointType& cellCenterPos)
+DynamicDSOOctree* DynamicDSOOctree::getChild(DeepSkyObject* & _obj, const PointType& cellCenterPos)
```
hv-work: reference to pointer?

pirogronian: Well, the only thing I did was to remove `const`. For all the rest, ask original author, please :-)

src/celengine/astroobj.cpp
pirogronian: Not exactly. Categories would't know the object is moved. This would lead to inconsistence.

src/celengine/astroobj.cpp
pirogronian: Oh yes, we need. See at implementation of `setIndex`.

src/celengine/astroobj.cpp
pirogronian: Why "0"?? And see my above comment.

pirogronian:
@hv-work
Btw, I requested review of #651 . This one is unfinished, but thanks anyway :-)

src/celengine/astroobj.cpp
375gnu: in that case why not to remove reference to old object and add one to new object?

375gnu:
i actually don't understand what is going here.

could you provide a problem description and how your pr fixes this problem?

pirogronian:
General idea or technical details?
I put general idea to the PR description. Several index-related methods in AstroObject is a result of seeking performance gain during data loading, mainly to avoid unnecessary lookup for a duplicate index, if we are sure there is no such a one. Maybe describe every method in doxygen-like style?

375gnu:
> Maybe describe every method in doxygen-like style?

This would be great.

src/celengine/octree.h
```diff
@@ -300,9 +300,9 @@ inline void DynamicOctree<OBJ, PREC>::rebuildAndSort(StaticOctree<OBJ, PREC>*& _
     OBJ* _firstObject = _sortedObjects;
 
     if (_objects != nullptr)
-        for (typename ObjectList::const_iterator iter = _objects->begin(); iter != _objects->end(); ++iter)
+        for (typename ObjectList::iterator iter = _objects->begin(); iter != _objects->end(); ++iter)
```
375gnu, line 303: could you rewrite this using C++11 style?

src/celengine/octree.h
```diff
@@ -300,9 +300,9 @@ inline void DynamicOctree<OBJ, PREC>::rebuildAndSort(StaticOctree<OBJ, PREC>*& _
...
         {
-            *_sortedObjects++ = **iter;
+            new (_sortedObjects++) OBJ(std::move(**iter));
```
375gnu, line 305: and move `_sortedObjects++` out of `new`, because of UB you might have here either `_sortedObjects` or `_sortedObjects + 1` if i correctly understand all this postfix `++` crap.

src/celengine/star.cpp
```diff
@@ -945,6 +945,13 @@ StarDetails::setInfoURL(const string& _infoURL)
 }
 
 
+Star::Star(Star &&s) : AstroObject(std::move(s))
+{
+    setPosition(s.getPosition());
+    setAbsoluteMagnitude(s.getAbsoluteMagnitude());
+    setDetails(s.getDetails());
```
375gnu, line 952: please use direct access to private members

src/celengine/star.h
```diff
@@ -241,8 +241,12 @@ class Star : public AstroObject
     Star() = default;
     virtual ~Star();
 
+    Star(Star &&s);
+
```
375gnu, line 245: add `Star(const Star&)` and appropriate operators `operator=`

src/celengine/stardb.cpp
```diff
@@ -691,28 +680,17 @@ bool StarDatabase::loadBinary(istream& in)
 
         nStars++;
     }
+    size_t i = 0;
+    while(i < nStars)
```
375gnu, line 684: please use `for` loop

src/celengine/stardb.cpp
```diff
@@ -754,7 +732,7 @@ void StarDatabase::finish()
 static void stcError(const Tokenizer& tok,
                      const string& msg)
 {
-    fmt::fprintf(cerr,  _("Error in .stc file (line %i): %s\n"), tok.getLineNumber(), msg);
+    fmt::fprintf(cout,  _("Error in .stc file (line %i): %s\n"), tok.getLineNumber(), msg);
```
375gnu, line 735: in final version please revert this change

src/celengine/stardb.cpp
```diff
@@ -980,7 +958,7 @@ bool StarDatabase::createStar(Star* star,
 
                 if (!hasBarycenter)
                 {
-                    fmt::fprintf(cerr, _("Barycenter %s does not exist.\n"), barycenterName);
+                    fmt::fprintf(cout, _("Barycenter %s does not exist.\n"), barycenterName);
```
375gnu, line 961: as above

src/celengine/stardb.cpp
```diff
@@ -1313,12 +1283,13 @@ bool StarDatabase::load(istream& in, const fs::path& resourcePath)
         bool ok = false;
         if (isNewStar && disposition == DataDisposition::Modify)
         {
-            clog << "Modify requested for nonexistent star.\n";
+            cout << "Modify requested for new star.\n";
```
375gnu, line 1286: as above, revert cout to clog

src/celutil/arraymap.h
```diff
@@ -0,0 +1,193 @@
...
+        return k >> (sizeof(k) * 8 - ArrayKeyLen);
+    }
+    size_t size() const { return ArraySize; }
+protected:
+    V m_array[ArraySize];
+    size_t m_used { 0 };
+public:
+    ArrayMap()
+    {
+        for(size_t i = 0; i < ArraySize; i++)
```
375gnu, line 26: use `std::fill` instead

src/celutil/packedrangeset.h
```diff
@@ -0,0 +1,280 @@
...
+            if (isSorted() && k < getKey(m_vector[i]))
+                return -1;
+        }
+        return -1;
+    }
+    bool insert(const V &v)
+    {
+        if (getSize() > 0 && isWithinRange(getKey(v)))
+        {
+            for (typename Container::iterator it = m_vector.begin(); it != m_vector.end(); it++)
```
375gnu, line 139: `auto` ?

src/celengine/star.h
pirogronian: Should I? AstroObject doesn't have copy constructor either. I think objects shouldn't be copy-able.

src/celengine/stardb.cpp
pirogronian: I was sure I already reverted it... :open_mouth:

src/celengine/octree.h
pirogronian: I got
> src/celengine/dsodb.cpp:385:46:   required from here
> src/celengine/octree.h:303:9: error: no matching function for call to ‘begin(std::vector<DeepSkyObject**, std::allocator<DeepSkyObject**> >*&)’

src/celengine/star.cpp
pirogronian: Why You prefer direct access? Using accessor seems to be more stable.

src/celengine/star.h
375gnu: in this case delete the explicitly:
```
Star(const Star&) = delete;
```
I'm sorry but C++ requires this.

src/celengine/star.cpp
375gnu: outside of the class - yes. but not here. first, this is internal class implementation. second it must set moved object's properties to "null" (nullptr for pointers, etc), you should not duplicate or add validation layers here, just take there, put here and make destructor called on the moved object essentially "nop".

src/celengine/octree.h
375gnu: for `for (const auto &o : *_objects)`?

src/celengine/dsooctree.cpp
375gnu: if you change something, cleanup please

src/celengine/star.cpp
pirogronian: Ok, essentially understood.

src/celengine/octree.h
pirogronian: Ah, forgot that `_objects` is a pointer...

src/celengine/star.h
pirogronian: Ok. But why not to delete it in `AstroObject` only? Should I delete it in every derived class?

src/celengine/astroobj.cpp
pirogronian: Hm... It could be a bit more efficient, but modification of elements inside set will change its order, hence will make it decoherent. It could have a point for map or vector, but is it worth effort?

src/celengine/astroobj.cpp
pirogronian: Ok, you mean InvalidIndex.

src/celengine/dsooctree.cpp
pirogronian: General implementation is done for references, while specialisation for DSOs uses pointers. It wouldn't be cleanup, but rewrite.

src/celengine/star.h
375gnu: derived classes should derive all constructors, and deleted too

375gnu:
Memory usage is 188MB vs 158MB. The growth is too large. But I haven't tested with CO yet.

pirogronian:
I'm surprised, as minimal memory usage growth should be 32MB :-). But it's constant growth, independent of objects number. If it's inacceptable anyway, I can tune container further.

pirogronian:
@375gnu
I'm waiting for your final review. Should I wait a bit more or you are just waiting until I improve minimal memory usage?

375gnu:
With CO stars.dat I have 454MB vs 363MB. If I increase limiting magnitude to 12 then it lifts to 470MB and even after returning magnitude to 8 it doesn't decrease. I'm afraid you need to optimize it.

pirogronian:
It's unbelivable, as my code allocates memory only for loading data at startup and has nothing to do with rendering. Moreover, I'm unable to reproduce this runtime memory usage increase.

pirogronian:
Adjusted a bit. But in practise memory efficiency still strongly depend on numbers distribution.

375gnu:
167MB with the default stars.dat with growth to 177 (max magnitude + position on Milky way).
447MB with CO's stras.dat with growth to 459

src/celutil/hashmap.h
```diff
@@ -0,0 +1,225 @@
...
+    bool m_hasZero { false };
+public:
+    static V invalidValue();
+    size_t size() const { return m_size + 1; }
+    size_t used() const { return m_hasZero ? m_used + 1 : m_used; }
+    const K *keyData() const { return m_keys; }
+    const V *valData() const { return m_vals; }
+    size_t hashedKey(K k) const
+    {
+        return (k ^ k >> (sizeof(k) * 4) * 0x9E3779B97F4A7C15L >> m_shift);
```
375gnu, line 30: where this magic came from?

pirogronian: Adapted from https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/LongMap.java
When in single function, gave wrong results.
Anyway, I'm again out of developing. Not sure, if I'll return again. You can forget this PR, if you have better things to do.

375gnu: that's sad

375gnu:
closing as no activities

PR #657, opened 2020-03-13
Use libepoxy instead of GLEW

375gnu:
Unlike GLEW libepoxy supports OpenGL ES and better handles cases with
functions provided by different extensions, e.g. the same functions
provided by vendor/EXT/ARB/core.

### DIFF
--- a/src/celengine/glsupport.cpp
+++ b/src/celengine/glsupport.cpp
@@ -1,0 +1,57 @@
+#include "glsupport.h"
+
+namespace celestia
+{
+namespace gl
+{
+
+bool ARB_shader_texture_lod         = false;
+bool ARB_vertex_array_object        = false;
+bool EXT_framebuffer_object         = false;
+bool EXT_texture_compression_s3tc   = false;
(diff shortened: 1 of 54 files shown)


eyvallah:
good to see it coming along!

375gnu:
blocked by #659

375gnu:
GLEW can be used still, at least I want to keep it for a while until I found a way to use it with mingw-w64 crosscompiler.

src/celengine/glsupport.cpp
```diff
@@ -0,0 +1,57 @@
...
+namespace gl
+{
+
+bool ARB_shader_texture_lod         = false;
+bool ARB_vertex_array_object        = false;
+bool EXT_framebuffer_object         = false;
+bool EXT_texture_compression_s3tc   = false;
+bool EXT_texture_filter_anisotropic = false;
+
+namespace
```
pirogronian: Why empty namespace?

375gnu: to make function local to the file. it's a c++ way instead of making functions `static`

PR #658, opened 2020-03-13
Update

Anthony-B-Russo10:
Added Markerz's repository for an Android port of the program.


375gnu:
i don't think we really need to add this now

eyvallah:
too much still needed to be worked on, so now is not a good time. if it finally matures, it should be part of CelestiaProject org

Anthony-B-Russo10:
This will be re-added once Markerz is ready for it.

PR #477, opened 2019-10-26
Update star textures

SevenSpheres:
Now they match the Blackbody D65 star colors. Most textures are by FarGetaNik in Project Echoes. The Y dwarf texture is by Edasich, the white dwarf texture is by Frank Gregorio, and the neutron star texture is by Cham. Closes #113.

PR #660, opened 2020-03-14
Update celestia.cfg and README

### LINKED ISSUE #113: Fix colours of star spectral classes
Reported by users

SevenSpheres:
For new star textures in CelestiaContent. Previously #477. Closes #113, once the corresponding PR in CelestiaContent is merged as well. The main repository still has the old data files, making things very confusing...

### DIFF
--- a/celestia.cfg
+++ b/celestia.cfg
@@ -82,34 +82,34 @@ StarTextures
 {
 	# This texture will be used for any spectral type not listed
 	# in this block.
-	Default "astar.jpg"
+	Default "star-type-a.*"
 
-	O "bstar.*"
-	B "bstar.*"
-	A "astar.*"
-	F "astar.*"
-	G "gstar.*"
-	K "gstar.*"
-	M "mstar.*"
+	O "star-type-o.*"
+	B "star-type-b.*"
+	A "star-type-a.*"
+	F "star-type-f.*"
+	G "star-type-g.*"
+	K "star-type-k.*"
+	M "star-type-m.*"
 
 	# carbon stars
-	C "mstar.*"
-	R "mstar.*"  # former subclass of carbon star
-	N "mstar.*"  # former subclass of carbon star
-	S "mstar.*"  # roughly between M and C
+	C "star-type-m.*"
+	R "star-type-m.*"  # former subclass of carbon star
+	N "star-type-m.*"  # former subclass of carbon star
+	S "star-type-m.*"  # roughly between M and C
 
 	# Wolf-Rayet stars
-	WC "bstar.*"
-	WN "bstar.*"
+	WC "star-type-o.*"
+	WN "star-type-o.*"
 
 	# brown dwarfs
-	L "browndwarf.*"
-	T "browndwarf.*"
-	Y "browndwarf.*"
+	L "star-type-l.*"
+	T "star-type-t.*"
+	Y "star-type-t.*"
 
 	# stellar remnants
-	WD "astar.jpg"
-	NeutronStar "astar.jpg"
+	WD "star-type-d.*"
+	NeutronStar "star-type-d.*"
 }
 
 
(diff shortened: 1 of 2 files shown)


SevenSpheres:
Linking [CelestiaContent#19](https://github.com/CelestiaProject/CelestiaContent/pull/19)

celestia.cfg
```diff
@@ -82,34 +82,34 @@ StarTextures
...
-	O "bstar.*"
-	B "bstar.*"
-	A "astar.*"
-	F "astar.*"
-	G "gstar.*"
-	K "gstar.*"
-	M "mstar.*"
+	O "star-type-o.*"
+	B "star-type-b.*"
+	A "star-type-a.*"
```
LukeCEL: Why are the star textures renamed? Wouldn't it make more sense to keep the names from earlier, for the sake of compatibility?

celestia.cfg
```diff
@@ -82,34 +82,34 @@ StarTextures
...
-	Y "browndwarf.*"
+	L "star-type-l.*"
+	T "star-type-t.*"
+	Y "star-type-t.*"
 
 	# stellar remnants
-	WD "astar.jpg"
-	NeutronStar "astar.jpg"
+	WD "star-type-d.*"
+	NeutronStar "NoTexture.*"
```
LukeCEL: What does this mean? There is no file called `NoTexture` in the other PR.

SevenSpheres:
> readme should be updated in celestiacontent instead, could you copy it there?

Why isn't it there already? In [CelestiaContent#4](https://github.com/CelestiaProject/CelestiaContent/issues/4) you wrote:
> here - data, extras*, models, textures, all other in the main repo.

celestia.cfg
SevenSpheres: Exactly, it's supposed to be a nonexistent texture. This is necessary because the neutron star texture is also used for black holes.

celestia.cfg
SevenSpheres: I don't know, those are FarGetaNik's names for them. The last release of Celestia Origin renamed their star textures in the same way. I don't think any of the old star textures are referenced in data files or addons, so there wouldn't be compatibility problems (note that I did keep the old browndwarf texture, for the sake of compatibility).

375gnu: you can rename them

celestia.cfg
375gnu: bad idea. it's better to open an issue to fix BD texture

375gnu:
> Why isn't it there already? In [CelestiaContent#4](https://github.com/CelestiaProject/CelestiaContent/issues/4) you wrote:
> 
> > here - data, extras*, models, textures, all other in the main repo.

i'm not perfect and do mistakes.

celestia.cfg
SevenSpheres: Give me a good reason why.

celestia.cfg
SevenSpheres: #662

celestia.cfg
LukeCEL: Compatibility - I'm pretty sure there are older add-ons on the Motherlode that use "mstar.*", "gstar.*", etc.

SevenSpheres: Like, way back when binary stars had to be defined as planets?

I prefer all the textures beginning with "star", but if everyone disagrees I'll rename them.

LukeCEL: Maybe, but I'm also talking about add-ons like [these](http://www.celestiamotherlode.net/catalog/show_addon_details.php?addon_id=1254) where they use the original naming format.

SevenSpheres: *sigh*...I really don't feel like doing this right now.

celestia.cfg
SevenSpheres: Black hole texture issue resolved, neutron star texture changed to one that exists.

PR #666, opened 2020-03-17
Move splash.png to splash directory

SevenSpheres:
> [splash.png is beeing searched in another location - in the splash/ subfolder. This was changed to be consistent with other interfaces.](https://celestia.space/forum/viewtopic.php?p=148492#p148492)

Then it needs to actually be there.


SevenSpheres:
#666 😈

375gnu:
no need, this is handled by cmake
https://github.com/CelestiaProject/Celestia/blob/master/CMakeLists.txt#L335

SevenSpheres:
Some of us (e.g. me) are using the files from GitHub + the binaries from Appveyor. The files should be in the right place *in the repository.*

SevenSpheres:
> in this case this pr should change all relevant files

*sigh...*

Like what?

375gnu:
cmake files

PR #682, opened 2020-04-01
Initial implementation of self shadowing

375gnu:
Known bugs and limitations:
  * Only the 1st light source casts shadows.
  * Thin objects have incorrect shadows, see ISS.

Disabled by default. Can be enabled by setting `ShadowMapSize` option in `celestia.cfg`/`~/.celestia.cfg`. I don't think that setting higher value that 4096 increases quality. I suggest 2048 for older hardware and 4096 for newer. If hardware allows then it's better enable MSAA: `AntialiasingSamples 1`.

For gl4es builds it's better to use `-DGL_ONLY_SHADOWS=0`.


375gnu:
> This line should be added to celestia.cfg (commented out) so users know it's an option.

i don't want to have all possible experimental options in celestia.cfg

PR #690, opened 2020-04-11
Update INSTALL.md, ENABLE_DATA set to ON by default

eyvallah:
Some changes the macosx build instruction and add ENABLE_DATA to supported parameters.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -274,23 +274,23 @@ Install Homebrew
 Install required packages:
 
 ```
-brew install cmake cspice fmt gettext libpng lua qt5 jpeg eigen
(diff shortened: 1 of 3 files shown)


INSTALL.md
```diff
@@ -274,23 +274,17 @@ Install Homebrew
 Install required packages:
 
 ```
-brew install cmake cspice fmt gettext libpng lua qt5 jpeg eigen
+brew install cmake glew fmt gettext libpng lua qt5 jpeg eigen
```
375gnu, line 277: why did you remove cspice?

eyvallah: i think it is optional, i'll list it in optional later

eyvallah: done

PR #701, opened 2020-04-20
Add Android/iOS/Mac string for i18n

eyvallah:


eyvallah:
Do I also need to update other .po files? @375gnu

375gnu:
Could we keep separately while new UIs are not added? The problem is that when we update them, all new strings will be removed because their source files are missing.

eyvallah:
like in a separate folder po3? @375gnu

375gnu:
Actually I was thinking about custom branch. But a `po3` folder will be ok as well. Also please rename it to `celestia_ui.pot` that way we can have a separate file on transifex for translators flexibility.

eyvallah:
I guess I'll do both, a separate branch (or just maybe in my fork, since the part of code are not very ready), and in a separate folder po3(but this leads me to think about existing shared items in celestia.pot...)

pirogronian:
Sorry, in fact I reviewed only last commit. Too tired to work...

eyvallah:
@pirogronian  it's ok, this is a draft. I should close it for now.

PR #719, opened 2020-05-05
Drop GLEW and use GL names provided by core/libepoxy

375gnu:


munix9:
don't forget https://github.com/CelestiaProject/Celestia/blob/master/debian/control
```
- libglew-dev
+ libepoxy-dev
```

PR #729, opened 2020-05-06
More precise solar radius value

SevenSpheres:
Consistent with the following papers:
https://arxiv.org/abs/0711.2392
https://arxiv.org/abs/1510.07674

### DIFF
--- a/src/celengine/astro.h
+++ b/src/celengine/astro.h
@@ -40,7 +40,7 @@
 
 #define EARTH_RADIUS   6378.14
 #define JUPITER_RADIUS 71492.0
-#define SOLAR_RADIUS   696000.0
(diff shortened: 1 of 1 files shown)


375gnu:
I haven't seen a consensus in the discord discussion.

LukeCEL:
According to the IAU resolution, that the _nominal_ radius is to be 695700 km. This value is not equal to what we know is the Sun's radius, but it doesn't matter, because the nominal radius is what astronomers use as a unit.

But changing the value might cause unintended effects, so I'm not sure about committing this change, at least for now.

By the way, the value 696000 is used elsewhere, such as in src/celestia/celestiacore.cpp and src/celestia/kde/selectionpopup.cpp.

PR #733, opened 2020-05-07
[tools] Implement spherical option for new star database format in startextdump

### LINKED ISSUE #623: startextdump.cpp doesn't output spherical coordinates
The tool startextdump.cpp, in src/tools/stardb, has a command-line argument `--spherical` (or `-s`), which makes it output a stars.txt with X, Y, and Z instead of the desired RA, Dec, and Distance.
 
Andrew Tribick explained it better in [this reply](https://celestia.space/forum/viewtopic.php?p=148169#p148169). Basically the `DumpStarDatabase` function doesn't know about the `useSphericalCoords` variable.

LukeCEL:
The sig figs for the spherical format are the same as outputted from buildstardb.pl.

Fixes #623

### DIFF
--- a/src/tools/stardb/startextdump.cpp
+++ b/src/tools/stardb/startextdump.cpp
@@ -187,7 +187,7 @@ bool DumpOldStarDatabase(istream& in, ostream& out, ofstream& hdOut,
         {
             Eigen::Vector3d pos = astro::equatorialToCelestialCart((double) RA, (double) dec, distance);
             float absMag = (float) (appMag / 256.0 + 5 -
-                                    5 * log10(distance / 3.26));
+                                    5 * log10(distance / LY_PER_PARSEC));
             out << (float) pos.x() << ' ' <<
                    (float) pos.y() << ' ' <<
                    (float) pos.z() << ' ';
@@ -206,7 +206,7 @@ bool DumpOldStarDatabase(istream& in, ostream& out, ofstream& hdOut,
 }
 
 
-bool DumpStarDatabase(istream& in, ostream& out)
+bool DumpStarDatabase(istream& in, ostream& out, bool spherical)
 {
     char header[8];
     in.read(header, sizeof header);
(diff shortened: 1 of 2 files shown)


src/tools/stardb/startextdump.cpp
```diff
@@ -250,9 +250,30 @@ bool DumpStarDatabase(istream& in, ostream& out)
...
-        out << ((float) absMag / 256.0f) << ' ';
+        
+        if (spherical)
+        {
+            Eigen::Vector3d eclipticpos = {(double) x, (double) y, (double) z};
+            Eigen::Vector3d pos = astro::eclipticToEquatorial(eclipticpos);
+            double distance = sqrt(x * x + y * y + z * z);
+            double phi = -acos(pos.y() / distance) * 180 / PI;
+            double theta = atan2(pos.z(), -pos.x()) * 180 / PI;
+            double ra = theta + 180;
```
375gnu, line 261: if `θ = RA + 180°` then `RA = θ - 180`

src/tools/stardb/startextdump.cpp
```diff
@@ -250,9 +250,30 @@ bool DumpStarDatabase(istream& in, ostream& out)
...
-        out << x << ' ' << y << ' ' << z << ' ';
-        out << setprecision(4);
-        out << ((float) absMag / 256.0f) << ' ';
+        
+        if (spherical)
+        {
+            Eigen::Vector3d eclipticpos = {(double) x, (double) y, (double) z};
+            Eigen::Vector3d pos = astro::eclipticToEquatorial(eclipticpos);
+            double distance = sqrt(x * x + y * y + z * z);
+            double phi = -acos(pos.y() / distance) * 180 / PI;
```
375gnu, line 259: Why `-` before `acos`? If `y' = d cosφ` then `φ = acos y'/d`

src/tools/stardb/startextdump.cpp
LukeCEL: I used `RA = θ + 180` instead of `RA = θ - 180` because atan2 outputs angles in the interval [-π, π], not [0, 2π] like you (and I) might expect. To me it makes much more sense just to add 180 degrees instead of subtract 180 degrees, even though they're equivalent.

If you want though, I could make this more clear by doing `ra = theta - 180 + 360;`

src/tools/stardb/startextdump.cpp
LukeCEL: Well, the declination is always between -90° and 90°. When converting to φ you subtract 90° (https://en.wikibooks.org/wiki/Celestia/Binary_Star_File) so φ should always be between -180° and 0°. Problem is that acos outputs angles in the interval [0, π]. To make it work for [-π, 0] I added the negative sign.

src/tools/stardb/startextdump.cpp
375gnu: yes, it's better to use `- 180 + 360` and leave a comment "atan2 return result in [-pi; pi] instead of [0, 2pi], so we add 360 to fix this"

src/tools/stardb/startextdump.cpp
375gnu: please leave then a comment why you use a minus

LukeCEL:
@SevenSpheres didn't know how to compile, so I removed the request for review.

SevenSpheres:
> @SevenSpheres didn't know how to compile, so I removed the request for review.

I mean, I can't use cmake here, and I was able to install GCC before but it wasn't working, so...

PR #737, opened 2020-05-12
[sdl] Initial implementation of a basic SDL interface

375gnu:
now it's required only for gl es testing. it's as simple as glut ui, maybe later we can extend it.


eyvallah:
after including <celutil/gettext.h> it passed build on my machine

PR #750, opened 2020-05-22
Optimize sphere rendering using degenerate triangles

375gnu:
Total time to render a 640x480 frame drops from 6.8 ms (6.2 the best case)
to 2.7 ms (2.8 the worst case) (tested on Intel UHD Graphics 620)

FYI [how it works](https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/TechniquesforWorkingwithVertexData/TechniquesforWorkingwithVertexData.html), Figure 8-6

@eyvallah could you create a test build for mobile phones before we merge this?


375gnu:
Just realized that with in "the real world" speedup won't be so impressive because RenderDoc adds additional GL calls for its own needs, and so the more draw calls you have the bigger overhead, but on my phone I have +5 fps (Earth: 32->37, Sun: 35->40).

PR #762, opened 2020-06-07
Use system font on Mac/iOS/Android

eyvallah:
Mac/iOS uses private API while Android requires Android 10.

Windows provides similar function https://docs.microsoft.com/en-us/windows/win32/api/dwrite_2/nn-dwrite_2-idwritefontfallback

### DIFF
--- a/src/celttf/truetypefont.cpp
+++ b/src/celttf/truetypefont.cpp
@@ -25,6 +25,30 @@
 #include <fstream>
 #endif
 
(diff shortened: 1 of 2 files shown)


375gnu:
I suppose it would be better to keep platform-specific code in separate files.

eyvallah:
Closing, I think we can instead just add methods that change font after initialization. Getting system font should be implemented in the frontend code since they vary greatly.

PR #769, opened 2020-06-10
OpenGL ES 2.0 initial support

375gnu:
* Makes GL2.1 code compatible with ES as well
 * Not pure ES 2.0, requires GL_OES_texture_border_clamp presence
 * Celx `gl` module and `font:render` doesn't work in ES mode

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,11 @@ option(FAST_MATH      "Build with unsafe fast-math compiller option (Default: of
 option(ENABLE_TTF     "Use TrueType fonts instead of TXF (Default: off)" OFF)
 option(ENABLE_TESTS   "Enable unit tests? (Default: off)" OFF)
 option(ENABLE_DATA    "Install data from content submodule? (Default: on)" ON)
+option(ENABLE_GLES    "Build for OpenGL ES 2.0 instead of OpenGL 2.1 (Default: off)" OFF)
+
+if(ENABLE_GLES)
+  add_definitions(-DGL_ES)
+endif()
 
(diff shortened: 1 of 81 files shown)


eyvallah:
I think this PR is all good now.

Though I tested with ANGLE on Android and iOS. I think ANGLE does not support struct in shaders well, had to comment out `#define USE_GLSL_STRUCTS` in shadermanager.cpp.
Another thing is with ANGLE based on Metal, there is some render issue, I guess it is due to that GL_MULTISAMPLE cannot be enabled or disabled in code?

375gnu:
i added one more commit to make `gl_Position` `invariant`, our planet rendering should benefit of this.

> In this section, variance refers to the possibility of getting different values from the same expression indifferent programs.  For example, say two vertex shaders, in different programs, each set gl_Position withthe same expression in both shaders, and the input values into that expression are the same when bothshaders run.  It is possible, due to independent compilation of the two shaders, that the values assigned togl_Position are not exactly the same when the two shaders run.  In this example, this can cause problemswith alignment of geometry in a multi-pass algorithm.

375gnu:
> Though I tested with ANGLE on Android and iOS. I think ANGLE does not support struct in shaders well, had to comment out `#define USE_GLSL_STRUCTS` in shadermanager.cpp.

I think we can add a CMAKE option to control this. And it would be great if you file a bug report for ANGLE.

> Another thing is with ANGLE based on Metal about multisample, there is some render issue, I guess it is due to that GL_MULTISAMPLE cannot be enabled or disabled in code?

GL_MULTISAMPLE is not a part of core GL ES 2.0, so currently its usage is disabled in our code.

PR #776, opened 2020-06-19
Fix warnings reported by clang/gcc

375gnu:


eyvallah:
seems MSVC is not happy 🤨

PR #780, opened 2020-06-21
Fix misaligned allocation of Console and HighPrecision_VertexBuffer

375gnu:
This should fix reported abnormal terminations on startup and on orbits enabled


eyvallah:
looks like we can specify where to store the PDB files by setting the CMAKE_PDB_OUTPUT_DIRECTORY value, the pdb file will be ${CMAKE_PDB_OUTPUT_DIRECTORY}/Debug/celestia-win.pdb

PR #782, opened 2020-06-21
Check path empty before check is_directory

eyvallah:
Android returns ENOENT when path is empty

### DIFF
--- a/src/celcompat/fs.cpp
+++ b/src/celcompat/fs.cpp
@@ -260,7 +260,7 @@ recursive_directory_iterator& recursive_directory_iterator::operator++()
         return *this;
 
     auto& p = m_iter->path();
-    if (m_pending && is_directory(p))
+    if (m_pending && (p.empty() || is_directory(p)))
     {
         m_dirs->m_dirIters.emplace_back(m_iter);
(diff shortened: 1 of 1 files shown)


375gnu:
not only android:

both libc++ (using llvm 10)
```
#include <filesystem>
#include <iostream>

int main()
{
        std::cout << (std::filesystem::is_directory("") ? "dir\n" : "no\n");
        return 0;
}
```

and glibc:
```
#include <stdio.h>
#include <sys/stat.h>

int is_dir(const char path[]);
int main()
{
    puts(is_dir("") ? "is dir" : "not a dir");
    return 0;
}
int is_dir(const char path[])
{
    struct stat buf;
    if (stat(path, &buf) != 0)
    {
        return 0;
    }
    return S_ISDIR(buf.st_mode);
}
```

have the same behaviour.

i suppose this is correct, so the check for empty path should be added where you observe the issue

eyvallah:
image

the call stack  looks like this

375gnu:
```
if (m_pending && (p.empty() || is_directory(p))
```
?

eyvallah:
looks like so, I'll update it

PR #787, opened 2020-06-23
Replace is_directory checking method

eyvallah:
https://appcenter.ms/orgs/CelestiaProject/apps/Celestia-2/crashes/errors/946044734u/overview

### DIFF
--- a/src/celengine/virtualtex.cpp
+++ b/src/celengine/virtualtex.cpp
@@ -261,7 +261,8 @@ void VirtualTexture::populateTileTree()
     for (int i = 0; i < MaxResolutionLevels; i++)
     {
         fs::path path = tilePath / fmt::sprintf("level%d", i);
-        if (fs::is_directory(path))
+        std::error_code ec;
+        if (fs::is_directory(path, ec))
         {
             maxLevel = i + baseSplit;
             int uLimit = 2 << maxLevel;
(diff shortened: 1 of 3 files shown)


375gnu:
This is POSIX implementation for `fs::exists`:
```
    struct stat buf;
    if (stat(p.c_str(), &buf) == 0)
        return true;

    if (errno == ENOENT)
        return false;

    ec = std::error_code(errno, std::system_category());
    return false;
```

and `fs::is_directory`:
```
    struct stat buf;
    if (stat(p.c_str(), &buf) != 0)
    {
        ec = std::error_code(errno, std::system_category());
        return false;
    }
    return S_ISDIR(buf.st_mode);
```

As you can see they are implemented the same way, and `fs::exists` is redundant for `fs::is_directory`.

Are you able to reproduce this crash yourself?

eyvallah:
@375gnu yes I can repro

the exception is thrown here
```
bool is_directory(const path& p)
{
    std::error_code ec;
    bool r = is_directory(p, ec);
    if (ec)
        throw filesystem_error(ec, "celfs::is_directory error");
    return r;
}
```

375gnu:
In this case, instead of using `fs::exists(dir) && fs::is_directory(dir)` better use the following:

```
std::error_code ec;
if (fs::is_directory(dir, ec))
...
```

Issue #677, opened 2020-03-25
[1.7.0] Attempting to lock orbit frame can crash Celestia


SevenSpheres:
**Describe the bug**
Typing `Shift+;` to lock orbit frame sometimes causes Celestia to crash, but only sometimes. This occurs only in 1.7.0, not in 1.6.x.

**To Reproduce**
1. Type `Shift+;`
2. Celestia crashes, if not try again until it does

**Expected behavior**
Celestia does not crash.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Frontend: both Qt and native Win
 - Version: 1.7.0, latest build as of today

**Additional context**
Reported by **Nrco0e**/**N3cronium** on Discord and confirmed by me. Not reproducible by @LukeCEL on Mac.

375gnu:
I'm unable to reproduce. Any chance you run this under debugger?

SevenSpheres:
> Any chance you run this under debugger?

I have to assume you mean Visual Studio's debugger. In that case, no, since [when I try to compile Celestia VS gives me errors](https://celestia.space/forum/viewtopic.php?f=4&t=20269).

SevenSpheres:
seems to be fixed by #793

PR #793, opened 2020-07-05
Fix crash due to improper initialization of TwoVectorFrame

levinli303:
use shared_ptr directly with new operator to ensure the Eigen allocator for alignment is called

### DIFF
--- a/src/celengine/observer.cpp
+++ b/src/celengine/observer.cpp
@@ -1522,25 +1522,25 @@ ObserverFrame::createFrame(CoordinateSystem _coordSys,
 
     case PhaseLock:
     {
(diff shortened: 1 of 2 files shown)


levinli303:
tested on firebase test lab with Moto G Play (4th Gen) XT1607, API Level 23

levinli303:
possibly closes #677

levinli303:
> just the other day i read about problems with make_shared. while it fixes some problems of c++11/14 it adds it's own, the most important is that it doesn't use overridden `new`. i think we should replace all make_shared

yeah, maybe when you got the time, there are not many I think.

PR #795, opened 2020-07-07
Internal blend/depth state management

375gnu:
several hundreds calls changing GL state less

still a lot to do to optimize them

### DIFF
--- a/src/celengine/console.h
+++ b/src/celengine/console.h
@@ -42,14 +42,16 @@ class ConsoleStreamBuf : public std::streambuf
     unsigned int decodeShift{ 0 };
 };
 
-
+class Renderer;
 class Console : public std::ostream
 {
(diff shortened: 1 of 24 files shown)


src/celengine/console.h
```diff
@@ -42,14 +42,16 @@ class ConsoleStreamBuf : public std::streambuf
     unsigned int decodeShift{ 0 };
 };
 
-
+class Renderer;
 class Console : public std::ostream
 {
  public:
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
levinli303, line 49: does not look like it contains Eigen structs here?

375gnu: I has: `Eigen::Matrix4f mpv;` private member.

PR #796, opened 2020-07-08
Shared library libcelestia.so/celestia.dll

375gnu:
not tested on windows yet

@munix9 this will affect your RPM & AppImage packaging, so I won't merge this until you're ready.


debian/rules
```diff
@@ -30,3 +30,8 @@ override_dh_install:
 	mv debian/celestia-common/usr/share/celestia/celestia.cfg \
 	   debian/celestia-common/etc/celestia.cfg
 	ln -s /etc/celestia.cfg debian/celestia-common/usr/share/celestia/
+	rm debian/celestia-common/usr/share/celestia/fonts/*
+	ln -s /usr/share/fonts/truetype/dejavu/DejaVuSans.ttf \
+	      debian/celestia-common/usr/share/celestia/fonts/
+	ln -s /usr/share/fonts/truetype/dejavu/DejaVuSansMono-Bold.ttf \
```
munix9, line 36: in celestia.cfg it reads `DejaVuSans-Bold.ttf`
https://github.com/CelestiaProject/Celestia/blob/master/celestia.cfg#L226

debian/control
```diff
@@ -94,9 +103,16 @@ Description: real-time visual space simulation (GLUT frontend)
  .
  This package contains a minimal frontend for Celestia.
 
+Package: libcelestia
```
munix9, line 106: libcelestia is installed like a system library?
so under /usr/lib/ or /usr/lib64/ ?
does debian/ubuntu require installation instructions such as debian/libcelestia.install ?
the names would be something like
`libcelestia.so.1 -> libcelestia.so.1.7.0`

munix9:
I don't know exactly what the defaults are for building packages for libraries under debian/ubuntu, here's how to do it for openSUSE/rpm:

cmake VERSION and SOVERSION could be
```
VERSION ${PROJECT_VERSION}
SOVERSION 1
```

This creates
```
/usr/lib64/libcelestia.so.1 -> libcelestia.so.1.7.0
/usr/lib64/libcelestia.so.1.7.0
```

The package name would be `libcelestia1`
The package manager must ensure that ldconfig is called after installing/uninstalling.
Currently this is done with rpm e.g. using
```
%post -n lib%{name}%{_soname} -p /sbin/ldconfig
%postun -n lib%{name}%{_soname} -p /sbin/ldconfig
```
in the spec file.

for debian see also
https://www.debian.org/doc/debian-policy/ch-sharedlibs.html
https://pusling.com/blog/?p=352

debian/rules
375gnu: sharp eye :)

I copy-pasted wrong name :)

debian/control
375gnu: > libcelestia is installed like a system library?
> so under /usr/lib/ or /usr/lib64/ ?

yes, it sounds like reasonable location to me.

> does debian/ubuntu require installation instructions such as debian/libcelestia.install ?

I forgot to add it to git. fixed.

> libcelestia.so.1 -> libcelestia.so.1.7.0

SONAME should be updated on breaking API changes, so we'll have  `libcelestia.so.1.7 -> libcelestia.so.1.7.0`

375gnu:
> The package manager must ensure that ldconfig is called after installing/uninstalling.

dpkg handles this automatically

375gnu:
> The package name would be `libcelestia1`

as 1.8 definitely will be incompatible with 1.7 this should be `libcelestia1.7`

munix9:
> > The package name would be `libcelestia1`
> 
> as 1.8 definitely will be incompatible with 1.7 this should be `libcelestia1.7`

Okay, that makes sense.

When you're ready to merge the pr, you can do that.
I have already made some adjustments to the rpm/appimage configuration.

375gnu:
gnu/linux & windows sides look good, macos is built. @levinli303 waiting for your input

munix9:
some build errors on obs:

Debian_10
https://build.opensuse.org/public/build/home:munix9:unstable/Debian_10/x86_64/celestia/_log
`find: 'chrpath': No such file or directory`

Ubuntu_18.04
https://build.opensuse.org/public/build/home:munix9:unstable/Ubuntu_18.04/x86_64/celestia/_log
```
[  126s] CMake Error at src/tools/cmod/CMakeLists.txt:25 (target_link_libraries):
[  126s]   Target "celmodel" of type OBJECT_LIBRARY may not be linked into another
[  126s]   target.  One may link only to STATIC or SHARED libraries, or to executables
[  126s]   with the ENABLE_EXPORTS property set.
[  126s] Call Stack (most recent call first):
[  126s]   src/tools/cmod/3dstocmod/CMakeLists.txt:1 (build_cmod_tool)
...
```

Ubuntu_20.04
https://build.opensuse.org/public/build/home:munix9:unstable/Ubuntu_20.04/x86_64/celestia/_log
`find: 'chrpath': No such file or directory`

openSUSE_Leap_15.1
https://build.opensuse.org/public/build/home:munix9:unstable/openSUSE_Leap_15.1/x86_64/celestia/_log
```
[   77s] CMake Error at src/tools/cmod/CMakeLists.txt:25 (target_link_libraries):
[   77s]   Target "celmodel" of type OBJECT_LIBRARY may not be linked into another
[   77s]   target.  One may link only to STATIC or SHARED libraries, or to executables
[   77s]   with the ENABLE_EXPORTS property set.
[   77s] Call Stack (most recent call first):
[   77s]   src/tools/cmod/3dstocmod/CMakeLists.txt:1 (build_cmod_tool)
...
```

For Ubuntu_18.04 and openSUSE_Leap_15.1 this could be due to the older cmake version:
- cmake-3.10.2-1ubuntu2
- cmake-3.10.2-lp151.4.1

375gnu:
I made #797 with fixes

PR #797, opened 2020-07-12
Fix build failures reported by munix9

375gnu:


375gnu:
@levinli303 what about this pr?

Issue #599, opened 2020-02-07
Simulate extinction


SevenSpheres:
Suggestion from **Fafers_br**, might as well post it here:

> Was visual extinction (Av) implemented in Celestia 1.7.0? At least as an optional variable in a stc file?

> Yeah. There could be an optional variable "Extinction" in the stc codes. When ommited, all the Celestia's calculations concerning the star would continue as they are today. But if one declares it's value, the Av would then enter in the calculations, thus resulting in more realistic values for the AbsMag and the estimated radius of the star using the black body approximation. But Celestia would have to assume isotropic and homogeneous extinction though (still better than no extinction at all, IMHO).

375gnu:
Would be good to have math behind extinction posted or linked here.

SevenSpheres:
The extinction value is the number of magnitudes by which the apparent magnitude is decreased. So the AbsMag calculation would use AppMag + Extinction.

SevenSpheres:
[Here](https://celestia.space/forum/viewtopic.php?f=4&t=20247) is a post by **Fafers_br** about this on the forum.

SevenSpheres:
It looks like extinction wasn't implemented correctly. This code:
```
1010293161 "Schulte 16:Cyg OB2 16:LS III +41 33" {
InfoURL "http://simbad.u-strasbg.fr/simbad/sim-id?Ident=schulte+16"
   RA  308.1607096
   Dec  41.4204975
   Distance  4940 #Gaia DR2+Bailer-Jones
   SpectralType "O7.5V"
   AppMag 10.84
   Extinction 4.5
}
```
results in this:
![schulte16](https://user-images.githubusercontent.com/56284207/88485431-d8c00a00-cf32-11ea-9883-ffad3d0ff0bc.png)
which is no different than if `AbsMag -4.56` had been entered in the stc file directly. The `AppMag` is supposed to remain the same.

375gnu:
if i understand correctly, then extinction is caused by dust clouds between a star and a viewer, so any calculation for any other position except the Earth won't be 100% correct.

375gnu:
@SevenSpheres  please see https://github.com/CelestiaProject/Celestia/pull/819

PR #819, opened 2020-07-26
Fix extinction

375gnu:
Closes: #599 

extinction is distance dependent. i.e. we'll have `app = absToApp(abs, distance) + extinction * distance`. but in stc file it will be as is.


375gnu:
@CelestiaProject/devs

Issue #745, opened 2020-05-19
Allow toggling visibility for stars


SevenSpheres:
The celx `setvisible` function only works for ssc and dsc objects. It would be nice if it worked for stars as well. Maybe also supporting `Visible`, `Beginning`, and `Ending` parameters in stc files.

PR #835, opened 2020-09-08
Allow toggling visibility for stars #745

lrm25:
For toggling star visibility.  I don't think that I need to check getDetails for null, since it looks like the star won't be loaded from the binary in src/celengine/stardb.cpp without details, but I can change that if necessary, or anything else.

### DIFF
--- a/src/celscript/lua/celx_object.cpp
+++ b/src/celscript/lua/celx_object.cpp
@@ -159,6 +159,10 @@ static int object_setvisible(lua_State* l)
     {
         sel->deepsky()->setVisible(visible);
     }
+    else if (sel->star() != nullptr)
+    {
+        sel->star()->getDetails()->setVisibility(visible);
+    }
 
     return 0;
 }


375gnu:
Welcome back @lrm25 !

Unfortunately your change doesn't work as you may expect because Celestia treats invisible stars as barycenters.

You can use the following script to test:
```
sel = celestia:find("Sol")

sel:setvisible(false)

wait(5)

sel:setvisible(true)
```

lrm25:
@375gnu I plugged your script in and the sun did disappear for 5 seconds, then reappear.  So, to clarify, do you want me to add another non-visible state for stars where they are not visible and won't be mistaken for barycenters?

375gnu:
@lrm25 this will be simplest solution, but I think first we need to discuss why we need this feature. cc: @SevenSpheres

SevenSpheres:
> this will be simplest solution, but I think first we need to discuss why we need this feature. cc: @SevenSpheres

For scripts/Lua applications that toggle the visibility of objects. What I specifically had in mind was "replacing" the Sun with different evolutionary stages (e.g. red giant), but it could be used for other things as well.

I can't test this PR since the Appveyor build failed; I could try to compile it but probably in a few days.

375gnu:
Actually "replacing" requires total overhaul of Celestia. But we can implement the simplest solution to visually hide stars using scripts. @lrm25 you are right, this will require an additional `boolean` value for stars. @levinli303 what do you think, is it worth to implement this?

SevenSpheres:
> Actually "replacing" requires total overhaul of Celestia.

That's why I put it in quotes...I mean like what's already been done with Lua for ssc objects, like "replacing" Earth with a cutaway model of its interior structure.

levinli303:
sorry for the late reply, been really busy lately.

@375gnu I don't have a strong opinion about this. generally I think being able to hide a star and show a start would be good in scripts, and I think we already support it for other objects.

375gnu:
closing as no activities

PR #842, opened 2020-09-26
Support DXTc software decompressing

levinli303:


levinli303:
@375gnu I just copied the code from gl4es, it seems to work, probably some cleaning needed.

375gnu (approved):
let's go with this for now

PR #843, opened 2020-09-27
Fix AppVeyor CI

levinli303:


SevenSpheres (commented):
You should probably rebase this PR so there's only one commit, but otherwise it looks good

SevenSpheres (approved):

PR #845, opened 2020-10-02
Separate modelview and projection

levinli303:
by looking at the dome version code, we need to separate projection matrix and model view matrix in shaders.

### DIFF
--- a/src/celengine/globular.cpp
+++ b/src/celengine/globular.cpp
@@ -467,8 +467,7 @@ void Globular::renderGlobularPointSprites(
     tidalProg->use();
     centerTex[ic]->bind();
 
-    Matrix4f mvp = (*m.projection) * (*m.modelview);
-    tidalProg->mat4Param("MVPMatrix") = mvp;
+    tidalProg->setMVPMatrices(*m.projection, *m.modelview);
 
     Matrix3f viewMat = viewerOrientation.conjugate().toRotationMatrix();
     tidalProg->vec4Param("color")       = Vector4f(Rr, Gg, Bb, min(2 * brightness * pixelWeight, 1.0f));
@@ -492,8 +491,9 @@ void Globular::renderGlobularPointSprites(
     globProg->use();
 
     globularTex->bind();
-    globProg->mat4Param("MVPMatrix") = mvp;
-    globProg->mat4Param("ModelViewMatrix") = vecgl::translate(*m.modelview, offset);
+    globProg->setMVPMatrices(*m.projection, *m.modelview);
+    // TODO: model view matrix should not be reset here
+    globProg->ModelViewMatrix = vecgl::translate(*m.modelview, offset);
     Matrix3f mx = Scaling(form->scale) * getOrientation().toRotationMatrix() * Scaling(tidalSize);
     globProg->mat3Param("m")            = mx;
     globProg->vec3Param("offset")       = offset;
--- a/src/celengine/rendcontext.cpp
+++ b/src/celengine/rendcontext.cpp
@@ -558,7 +558,7 @@ GLSL_RenderContext::makeCurrent(const Material& m)
         return;
 
     prog->use();
-    prog->MVPMatrix = (*projectionMatrix) * (*modelViewMatrix);
+    prog->setMVPMatrices(*projectionMatrix, *modelViewMatrix);
 
     for (unsigned int i = 0; i < nTextures; i++)
     {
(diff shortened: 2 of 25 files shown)


375gnu (changes requested):
from performance pov it's better not to multiply the same matrices for all vertices. could you instead add a Projection and ModelView uniform variables to shaders and set them but still prefer MVPMatrix?

src/celengine/rendcontext.cpp
```diff
@@ -772,9 +773,8 @@ GLSLUnlit_RenderContext::makeCurrent(const Material& m)
         return;
 
     prog->use();
-    prog->MVPMatrix = (*projectionMatrix) * (*modelViewMatrix);
-    if (usePointSize)
-        prog->ModelViewMatrix = *modelViewMatrix;
```
levinli303, line 777: @375gnu  also was not sure about this, where ModelViewMatrix might not be set

src/celengine/globular.cpp
```diff
@@ -492,7 +492,8 @@ void Globular::renderGlobularPointSprites(
     globProg->use();
 
     globularTex->bind();
-    globProg->mat4Param("MVPMatrix") = mvp;
+    globProg->ProjectionMatrix = *m.projection;
+    globProg->ModelViewMatrix = *m.modelview;
     globProg->mat4Param("ModelViewMatrix") = vecgl::translate(*m.modelview, offset);
```
levinli303, line 497: and here... MVPMatrix is not a multiply of Projection * ModelView

src/celengine/rendcontext.cpp
375gnu: in the current code we don't need `ModelViewMatrix` every time, but to simplify your changes you can set it always.

src/celengine/globular.cpp
375gnu: yes, i remember i stuck with this when was converting Celestia to GL ES, but i left this for future fix. i need to check the original code. but again, currently you can set all 3 matrices independently and we will fix this later.

levinli303:
done, changed to use MVP again.

src/celengine/shadermanager.cpp
```diff
@@ -3415,13 +3417,18 @@ CelestiaGLProgram::attribIndex(const std::string& paramName) const
     return glGetAttribLocation(program->getID(), paramName.c_str());
 }
 
+void
+CelestiaGLProgram::initCommonParameters()
+{
+    ModelView = mat4Param("ModelView");
+    Projection = mat4Param("Projection");
```
375gnu, line 3424: could you rename these variables to `ModelViewMatrix` and `ProjectionMatrix`?

src/celttf/truetypefont.h
```diff
@@ -27,7 +27,7 @@ class TextureFont
     TextureFont& operator=(const TextureFont&) = delete;
     TextureFont& operator=(TextureFont&&) = delete;
 
-    void setMVPMatrix(const Eigen::Matrix4f& m);
+    void setMVPMatrices(const Matrices &mvp);
```
375gnu, line 30: not sure whether `Matrices` is good here, i see there is a lot places where you use `{&m, &p}` maybe `Matrix4f &m, Matrix4f &p` is better?

375gnu (changes requested):

src/celengine/shadermanager.cpp
levinli303: done.

src/celttf/truetypefont.h
levinli303: done, prefer projection then model view.

src/celengine/render.cpp
```diff
@@ -5334,7 +5336,7 @@ bool Renderer::captureFrame(int x, int y, int w, int h, Renderer::PixelFormat fo
     return glGetError() == GL_NO_ERROR;
 }
 
-void Renderer::drawRectangle(const Rect &r, const Matrix4f &mvp)
+void Renderer::drawRectangle(const Rect &r, const Matrices &mvp)
```
375gnu, line 5339: please use 2 separate Eigen::Matrix4f instead of Matrices here as well

src/celengine/shadermanager.h
```diff
@@ -127,6 +127,11 @@ class ShaderProperties
     friend class ShaderManager;
 };
 
+struct Matrices
```
375gnu, line 130: please return this to render.h

375gnu (changes requested):

src/celengine/render.cpp
levinli303: fixed.

src/celengine/shadermanager.h
levinli303: fixed

375gnu (approved):

PR #846, opened 2020-10-06
Add view port effect for post processing

levinli303:
When we need to apply a viewport effect, we first render to a texture, then render the texture with effect on screen. This will be needed for fisheye warping.

Configure in celestia.cfg, two effects available:
1. Passthrough effect, just draw the same thing from the texture to the screen. `ViewportEffect "passthrough"`
2. Warp mesh effect, warping according to the warp mesh file provided. `ViewportEffect "warpmesh" WarpMeshFile "some warp mesh file name in the warp folder"`

The warp mesh effect and file format is from http://paulbourke.net/dataformats/meshwarp/ which is basically a mapping between vertex coordinates and texture coordinates.

### DIFF
--- a/src/celengine/mapmanager.cpp
+++ b/src/celengine/mapmanager.cpp
@@ -1,0 +1,196 @@
+//
+// mapmanager.cpp
+//
+// Copyright © 2020 Celestia Development Team. All rights reserved.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include <celutil/debug.h>
+#include <fstream>
+#include "mapmanager.h"
+
+using namespace std;
+
+WarpMesh::WarpMesh(int nx, int ny, float *data) :
+    nx(nx),
+    ny(ny),
+    data(data)
+{
+}
+
+WarpMesh::~WarpMesh()
+{
+    delete data;
+}
+
+void WarpMesh::scopedDataForRendering(const function<void (float*, int)>& f) const
+{
+    int step = 5 * 6;
+    int size = (nx - 1) * (ny - 1) * step * sizeof(float);
+    float *renderingData = new float[size];
+    for (int y = 0; y < ny - 1; y += 1)
+    {
+        for (int x = 0; x < nx - 1; x += 1)
+        {
+            float *destination = &renderingData[(y * (nx - 1) + x) * step];
+            float *source = &data[(y * nx + x) * 5];
+            // Top left triangle
+            memcpy(destination, source + 5 * nx, sizeof(float) * 5);
+            memcpy(destination + 5, source, sizeof(float) * 5);
+            memcpy(destination + 10, source + 5, sizeof(float) * 5);
+            // Bottom right triangle
+            memcpy(destination + 15, source + 5 * nx, sizeof(float) * 5);
+            memcpy(destination + 20, source + 5, sizeof(float) * 5);
+            memcpy(destination + 25, source + 5 * nx + 5, sizeof(float) * 5);
+        }
+    }
+    f(renderingData, size);
+    delete[] renderingData;
+}
+
+int WarpMesh::count() const
+{
+    return 6 * (nx - 1) * (ny - 1);
+}
+
+bool WarpMesh::mapVertex(float x, float y, float* u, float* v) const
+{
+    float minX = data[0];
+    float minY = data[1];
+    float maxX = data[(nx * ny - 1) * 5];
+    float maxY = data[(nx * ny - 1) * 5 + 1];
+
+    float stepX = (maxX - minX) / (nx - 1);
+    float stepY = (maxY - minY) / (ny - 1);
+
+    float locX = (x - minX) / stepX;
+    float locY = (y - minY) / stepY;
+    int floX = floorf(locX);
+    int floY = floorf(locY);
+    locX -= floX;
+    locY -= floY;
+
+    if (floX < 0 || floX >= nx - 1 || floY < 0 || floY >= ny - 1)
+        return false;
+
+    float p1x = data[(floY * nx + floX) * 5 + 2];
+    float p1y = data[(floY * nx + floX) * 5 + 3];
+    float p2x = data[(floY * nx + floX + 1) * 5 + 2];
+    float p2y = data[(floY * nx + floX + 1) * 5 + 3];
+    float p3x = data[(floY * nx + floX + nx) * 5 + 2];
+    float p3y = data[(floY * nx + floX + nx) * 5 + 3];
+    float p4x = data[(floY * nx + floX + nx + 1) * 5 + 2];
+    float p4y = data[(floY * nx + floX + nx + 1) * 5 + 3];
+
+    if (locX + locY <= 1)
+    {
+        // the top left part triangle
+        *u = p1x + locX * (p2x - p1x) + locY * (p3x - p1x);
+        *v = p1y + locX * (p2y - p1y) + locY * (p3y - p1y);
+    }
+    else
+    {
+        // the bottom right triangle
+        locX -= 1;
+        locY -= 1;
+        *u = p4x + locX * (p4x - p3x) + locY * (p4x - p2x);
+        *v = p4y + locX * (p4y - p3y) + locY * (p4y - p2y);
+    }
+    // Texture coordinate is [0, 1], normalize to [-1, 1]
+    *u = (*u) * 2 - 1;
+    *v = (*v) * 2 - 1;
+    return true;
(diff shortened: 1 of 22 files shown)


levinli303:
Need to be tested, so far it works ok on Mac and iOS.  Used [warp.zip](https://github.com/CelestiaProject/Celestia/files/5331887/warp.zip) for testing the warpmesh.

375gnu:
i would like not to have GL calls in any celestia/* files, could you move them to celengine/ introducing new helper classes if necessary. another problem is oldfbo in FBO class, this violates encapsulation rules.

levinli303:
@375gnu  maybe we can just move the glClear(Color) to inside ViewportEffect? The others are not direct GL calls, except for generating FBOs which I can't seem to find a trivial way to move to celengine since it is tied to view.

375gnu:
> maybe we can just move the glClear(Color) to inside ViewportEffect?

yes, this sounds reasonable

levinli303:
> > maybe we can just move the glClear(Color) to inside ViewportEffect?
> 
> yes, this sounds reasonable

this PR should be ready now.

375gnu:
@levinli303 i'm still isn't happy with FBO unbinding. Generally we need to know if there is any FBO bound only in `renderGeometry_GLSL` . We can save currently bound FBO id before calling `renderGeometryShadow_GLSL` and restore it after. In this case `FramebufferObject` will remain the same.

levinli303:
I think this is how it is done here. I added this here because OpenGLES on iOS's default FBO to render to is not 0.

src/celengine/mapmanager.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+    // Texture coordinate is [0, 1], normalize to [-1, 1]
+    *u = (*u) * 2 - 1;
+    *v = (*v) * 2 - 1;
+    return true;
+}
+
+static WarpMeshManager* warpMeshManager = nullptr;
+
+extern WarpMeshManager* GetWarpMeshManager()
```
375gnu, line 106: please remove `extern` and move `static WarpMeshManager* warpMeshManager = nullptr;` into the function body

src/celengine/mapmanager.cpp
```diff
@@ -0,0 +1,195 @@
...
+        warpMeshManager = new WarpMeshManager("warp");
+    return warpMeshManager;
+}
+
+static string resolveWildcard(const string& filename)
+{
+    string base(filename, 0, filename.length() - 1);
+
+    string mapfile = base + "map";
+    {
```
375gnu, line 118: no need for braces here

src/celengine/mapmanager.cpp
```diff
@@ -0,0 +1,195 @@
...
+        ifstream in(mapfile);
+        if (in.good())
+            return mapfile;
+    }
+
+    return "";
```
375gnu, line 124: please use `return {}` here

src/celengine/mapmanager.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    return filename;
+}
+
+
+WarpMesh* WarpMeshInfo::load(const fs::path& name)
+{
+#define MESHTYPE_RECT 2
+    ifstream f(name.c_str());
```
375gnu, line 150: please change to `name.string()`

src/celengine/mapmanager.h
```diff
@@ -0,0 +1,61 @@
...
+#include <string>
+#include <map>
+#include <functional>
+#include <celutil/resmanager.h>
+
+// File format for data used to warp an image, for
+// detail, see http://paulbourke.net/dataformats/meshwarp/
+class WarpMesh
+{
+public:
```
375gnu, line 23: please add a space before `public` & `private`

src/celengine/mapmanager.h
```diff
@@ -0,0 +1,61 @@
...
+};
+
+inline bool operator<(const WarpMeshInfo& wi0, const WarpMeshInfo& wi1)
+{
+    return wi0.source < wi1.source;
+}
+
+typedef ResourceManager<WarpMeshInfo> WarpMeshManager;
+
+extern WarpMeshManager* GetWarpMeshManager();
```
375gnu, line 58: no need for `extern`

src/celengine/viewporteffect.cpp
```diff
@@ -0,0 +1,156 @@
...
+        return false;
+
+    glClearColor(0.0f, 0.0f, 0.0f, 1.0f);
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+    return true;
+}
+
+bool ViewportEffect::distortXY(float &x, float &y) { return true; };
+
+VOViewportEffect::VOViewportEffect(GLenum bufferType, GLsizeiptr bufferSize, GLenum streamType) : vo(bufferType, bufferSize, streamType)
```
375gnu, line 35: please move `vo(...)` to the next line

src/celengine/viewporteffect.cpp
```diff
@@ -0,0 +1,156 @@
...
+    prog->use();
+    configureShader(prog, width, height);
+    glBindTexture(GL_TEXTURE_2D, fbo->colorTexture());
+    draw(vo);
+    glBindTexture(GL_TEXTURE_2D, 0);
+    vo.unbind();
+    return true;
+}
+
+PassthroughViewportEffect::PassthroughViewportEffect() : VOViewportEffect(GL_ARRAY_BUFFER, 0, GL_STATIC_DRAW)
```
375gnu, line 58: please move `VOViewportEffect(...)` to  the next line

src/celengine/viewporteffect.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+    prog->samplerParam("tex") = 0;
+}
+
+void PassthroughViewportEffect::draw(celgl::VertexObject& vo)
+{
+    vo.draw(GL_TRIANGLES, 6);
+}
+
+WarpMeshViewportEffect::WarpMeshViewportEffect(WarpMesh *mesh) : VOViewportEffect(GL_ARRAY_BUFFER, 0, GL_STATIC_DRAW), mesh(mesh) {}
```
375gnu, line 103: as above

src/celengine/mapmanager.h
```diff
@@ -0,0 +1,61 @@
...
+// mapmanager.h
+//
+// Copyright © 2020 Celestia Development Team. All rights reserved.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#ifndef _MAPMANAGER_H_
```
375gnu, line 11: please use `#pragma once`

src/celengine/viewporteffect.h
```diff
@@ -0,0 +1,80 @@
...
+};
+
+class VOViewportEffect : public ViewportEffect
+{
+public:
+    VOViewportEffect(GLenum bufferType, GLsizeiptr bufferSize, GLenum streamType);
+    ~VOViewportEffect() override = default;
+
+    bool render(Renderer*, FramebufferObject*, int width, int height) override;
+    virtual std::string getShaderName() = 0;
```
375gnu, line 43: this method should be `const`

src/celengine/viewporteffect.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+    if (!fbo->unbind(oldFboId))
+        return false;
+
+    glClearColor(0.0f, 0.0f, 0.0f, 1.0f);
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+    return true;
+}
+
+bool ViewportEffect::distortXY(float &x, float &y) { return true; };
```
375gnu, line 33: please make multiline

src/celengine/viewporteffect.h
```diff
@@ -0,0 +1,80 @@
...
+    virtual bool preprocess(Renderer*, FramebufferObject*);
+    virtual bool prerender(Renderer*, FramebufferObject*);
+    virtual bool render(Renderer*, FramebufferObject*, int width, int height) = 0;
+    virtual bool distortXY(float& x, float& y);
+
+private:
+    GLint oldFboId;
+};
+
+class VOViewportEffect : public ViewportEffect
```
375gnu, line 36: do we really need yet another virtual subclass?

375gnu (changes requested):
please add a description about warp algorithm used, at least a link to documentation

i'm still unhappy with FBO usage in CelestiaCore and FBO switching in unbind(), but at least for now we can live with this and will refactor this code later

375gnu (approved):
let's roll

PR #849, opened 2020-10-09
Restore sun halo

### LINKED ISSUE #541: Sun halo in different starfield rendering modes
Both in 1.6 and 1.7 versions. Depending on starfield's style a sun halo either present or not:
No halo with scaled disks:
disks
Halo present with points and fuzzy points:
fuzzy
points

1.7 had the same behaviour before #533. Now it has less bright halo in «points» and «fuzzy points» modes.

Quite interesting is that Alexell's version didn't have this bug. UPD: when vp&fp programs doesn't present, so older strarfied with quads is in use.

levinli303:
Just copied some of the code back. Fix #541  ?

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -1827,6 +1827,54 @@ void renderPoint(const Renderer &renderer,
 #endif
 }
 
+void renderLargePoint(const Renderer &renderer,
+                      const Vector3f &position,
+                      const Color &color,
+                      float size,
+                      float pixelSize,
+                      const Matrix3f &cameraOrientation,
+                      const Matrices &mvp)
+{
+    auto *prog = renderer.getShaderManager().getShader("largestar");
+    if (prog == nullptr)
+        return;
+
+    prog->use();
+    prog->samplerParam("starTex") = 0;
+    prog->setMVPMatrices(*mvp.projection, *mvp.modelview);
+    prog->vec4Param("color") = color.toVector4();
+
+    float distanceAdjust = pixelSize * position.norm() * 0.5f;
+    size *= distanceAdjust;
+
+    auto m = renderer.getCameraOrientation().conjugate().toRotationMatrix();
+    Vector3f v0 = position + (cameraOrientation * Vector3f(-1, -1, 0) * size);
+    Vector3f v1 = position + (cameraOrientation * Vector3f( 1, -1, 0) * size);
+    Vector3f v2 = position + (cameraOrientation * Vector3f( 1,  1, 0) * size);
+    Vector3f v3 = position + (cameraOrientation * Vector3f(-1,  1, 0) * size);
+
+    float quadVertices[] = {
+        // positions            // texCoords
+        v0.x(), v0.y(), v0.z(), 0.0f, 0.0f,
+        v1.x(), v1.y(), v1.z(), 1.0f, 0.0f,
+        v2.x(), v2.y(), v2.z(), 1.0f, 1.0f,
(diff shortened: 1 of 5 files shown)


375gnu:
old method used quads instead of points so because of that additional math was required. and here we still have the old bug with halo disappearing when a sun goes away from the screen

levinli303:
yeah, maybe we should take the halo size into account when we decide if a star should be visible too. and by the way, we might need to draw quad when the point size is too large. for example, some ANGLE implementation limits point size to 64.

375gnu:
i tried to restore the original code but without success

SevenSpheres:
Tested now but I see this has been marked as draft now - anyway for me this only shows a bright point at a star's center, it doesn't restore the "halo" on the whole star. (Screenshot has FarGetaNik's gstar texture and not the default)

sol849

levinli303:
@375gnu  Updated to draw quads when point size > 64, or should we always draw quad in this method? This does not fix the issue where the halo disappearing when a sun goes away from the screen, but maybe we can still have this fix in?

Simulator Screen Shot - iPhone 6 - 2020-10-24 at 18 03 50

levinli303:
@SevenSpheres can you try this https://ci.appveyor.com/project/375gnu/celestia/builds/35932620/artifacts and see if it fixes your problem. you also need to download new shaders (largestar*) from https://github.com/levinli303/Celestia/tree/restore-sun-halo/shaders

SevenSpheres:
> @SevenSpheres can you try this https://ci.appveyor.com/project/375gnu/celestia/builds/35932620/artifacts and see if it fixes your problem. you also need to download new shaders (largestar*) from https://github.com/levinli303/Celestia/tree/restore-sun-halo/shaders

Yes, that works!

src/celengine/render.cpp
```diff
@@ -140,6 +140,8 @@ static const float MaxAsterismLinesConstDist   = 600.0f;
 static const float MaxAsterismLabelsDist = 20.0f;
 static const float MaxAsterismLinesDist  = 6.52e4f;
 
+static const float MaxPointSize  = 64.0f;
```
375gnu, line 143: don't hardcode this. use
```
GLint pointSizeRange[2];
glGetIntegerv(GL_SMOOTH_POINT_SIZE_RANGE, pointSizeRange);
MaxPointSize = pointSizeRange[1];
```
somewhere in `Renderer::Renderer()`

375gnu (changes requested):

src/celengine/render.cpp
levinli303: maybe we can put it in gl::init?

levinli303: and should we use GL_ALIASED_POINT_SIZE_RANGE for gles?

375gnu: > maybe we can put it in gl::init?

sounds like a good idea



> and should we use GL_ALIASED_POINT_SIZE_RANGE for gles?

yes

levinli303: done

levinli303:
https://ci.appveyor.com/project/375gnu/celestia/builds/35948238/artifacts @SevenSpheres  maybe you can test again?

I tested on macOS and iOS it works well. Also works on macCatalyst with ANGLE (max point size is 64)

SevenSpheres (approved):

LukeCEL:
Is it possible to apply the sun halo effect in "Scaled Disk" mode?

375gnu (approved):

PR #851, opened 2020-10-10
Support fisheye projection

levinli303:
Basically adding the same things to every shader, is there a better approach?  Values that depend on FOVs also needs to be worked on.

### DIFF
--- a/src/celengine/observer.cpp
+++ b/src/celengine/observer.cpp
@@ -1352,6 +1352,21 @@ Vector3f Observer::getPickRay(float x, float y) const
     return pickDirection.normalized();
 }
 
+Vector3f Observer::getPickRayFisheye(float x, float y) const
+{
+    float r = hypot(x, y);
+    float phi = float(PI) * r;
+    float sin_phi = sin(phi);
+    float theta = atan2(y, x);
+    float newX = sin_phi * cos(theta);
+    float newY = sin_phi * sin(theta);
+    float newZ = cos(phi);
+    Vector3f pickDirection = Vector3f(newX, newY, -newZ);
+    pickDirection.normalize();
+
+    return pickDirection;
+}
+
 
 // Internal method to update the position and orientation of the observer in
 // universal coordinates.
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -72,6 +72,28 @@ uniform mat4 MVPMatrix;
 invariant gl_Position;
 )glsl";
 
+static const char *VPFunction =
+    "#ifdef FISHEYE\n"
+    "void set_vp(vec4 in_Position) {\n"
+    "    float PID2 = 1.570796326794896619231322;\n"
+    "    vec4 inPos = ModelViewMatrix * in_Position;\n"
+    "    float l = length(inPos.xy);\n"
+    "    if (l != 0.0)\n{\n"
+    "        float phi = atan(l, -inPos.z);\n"
+    "        float lensR = phi / PID2;\n"
+    "        inPos.xy *= (lensR / l);\n"
+    "    }\n"
+    "    gl_Position = ProjectionMatrix * inPos;\n"
+    "}\n"
+    "#else\n"
+    "void set_vp(vec4 in_Position) {\n"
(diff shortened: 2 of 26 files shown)


375gnu:
i think this should look as following:
1. fisheye function is in a separate file, e.g. `shaders/fisheye.h`
2. after `#version ...` we add `#define FISHEYE` if fish eye projection is required
3. then we add `#ifdef FISHEYE\n#include <fisheye.h>\n#endif`
4. in the main body `if (fisheye) { ... } else { ... }` is replaced with `#ifdef FISHEYE\nfisheye(inPos)\n#else\n...\n#endif`
5. in `shaderproperties.cpp` in `bool operator<(const ShaderProperties& p0, const ShaderProperties& p1)` add `if (p0.fisheye != p1.fisheye) return int(p0.fisheye) < int(p1.fisheye);`

levinli303:
@375gnu  can we use include in GLSL?

375gnu:
Forgot to say that glsl doesn't support `#include` so the usual way is to check if a file contains it and replace with actual content.

levinli303:
Finished a dumb version where fisheye is set ShaderManager. for shaderproperties  we could have a property like that, but for shaders from files, I haven't found a viable way

src/celestia/celestiacore.cpp
```diff
@@ -79,8 +79,8 @@ static const float RotationBraking = 10.0f;
 static const float RotationDecay = 2.0f;
 static const double MaximumTimeRate = 1.0e15;
 static const double MinimumTimeRate = 1.0e-15;
-static const float stdFOV = degToRad(45.0f);
-static const float MaximumFOV = degToRad(120.0f);
+static const float stdFOV = degToRad(60.0f);
```
375gnu, line 82: shouldn't this be controlled by fish eye mode?

src/celestia/celestiacore.cpp
```diff
@@ -2284,8 +2284,13 @@ void CelestiaCore::setFOVFromZoom()
     for (const auto v : views)
         if (v->type == View::ViewWindow)
         {
-            double fov = 2 * atan(height * v->height / (screenDpi / 25.4) / 2. / distanceToScreen) / v->zoom;
-            v->observer->setFOV((float) fov);
+            if (renderer->getProjectionMode() == Renderer::FisheyeMode)
+                v->observer->setFOV((float) MaximumFOV);
```
375gnu, line 2288: MaximumFOV is already float

src/celengine/shadermanager.cpp
```diff
@@ -68,10 +68,33 @@ static const char* VertexHeader = R"glsl(
 uniform mat4 ModelViewMatrix;
 uniform mat4 ProjectionMatrix;
 uniform mat4 MVPMatrix;
+uniform bool fisheye;
```
375gnu, line 71: this uniform is not used anywhere

src/celengine/shadermanager.cpp
```diff
@@ -3383,6 +3438,11 @@ CelestiaGLProgram::intParam(const string& paramName)
     return IntegerShaderParameter(program->getID(), paramName.c_str());
 }
 
+BooleanShaderParameter
```
375gnu, line 3441: BooleanShaderParameter is not used anywhere

375gnu (changes requested):

levinli303:
should have fixed all the issues. also disabled the star glare rendering on fisheye for now

src/celengine/observer.cpp
```diff
@@ -1352,6 +1352,21 @@ Vector3f Observer::getPickRay(float x, float y) const
     return pickDirection.normalized();
 }
 
+Vector3f Observer::getPickRayFisheye(float x, float y) const
+{
+    float r = sqrt(x * x + y * y);
```
375gnu, line 1357: you can replace with `hypot` here

375gnu (changes requested):

src/celengine/observer.cpp
levinli303: fixed

levinli303:
let's merge this

375gnu (approved):
let's merge

PR #868, opened 2020-11-12
Use GL_RGBA to read pixels on GLES

levinli303:
https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glReadPixels.xhtml

### DIFF
--- a/src/celengine/render.h
+++ b/src/celengine/render.h
@@ -222,6 +222,7 @@ class Renderer
     // Currently we map them 1:1 to GL
     enum class PixelFormat
     {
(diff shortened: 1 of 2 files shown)


375gnu:
docs say RGB is also supported. what's wrong with it?

levinli303:
@375gnu  GL_RGBA/GL_UNSIGNED_BYTE is guaranteed to be supported on GLES2 core. As I have tested, using GL_RGB on iOS produces GL_INVALID_ENUM

375gnu (approved):

PR #874, opened 2020-11-17
Fix GTK building with gettext

levinli303:
@munix9


munix9:
Local build on openSUSE_Tumbleweed is fine now.
Thanks.

375gnu (approved):

PR #880, opened 2020-11-29
Fix star rendering in PointStar mode

levinli303:
Forgot to set MVP matrices in this case.


375gnu (approved):

levinli303:
let's see if we can just use #882

PR #881, opened 2020-11-29
Fix bookmark organizer

levinli303:
incorrect handling for TVM_GETITEMW.


375gnu:
we should test this first, but i don't have access to windows vm currently

375gnu (approved):

PR #882, opened 2020-11-29
Always use gl_PointSize in shaders to set point size

levinli303:
On iOS, we have to set gl_PointSize in order to draw points, otherwise it will simply not display. In non-sprite model drawing and point star drawing, the shader did not set gl_PointSize. This should also fix the point rendering on HiDPI screens.

### DIFF
--- a/shaders/largestar_vert.glsl
+++ b/shaders/largestar_vert.glsl
@@ -1,4 +1,4 @@
-attribute vec3 in_Position;
+attribute vec2 in_Position;
 attribute vec2 in_TexCoord0;
 
 uniform float pointWidth;
@@ -11,6 +11,6 @@ void main(void)
 {
     texCoord = in_TexCoord0.st;
     set_vp(vec4(center, 1.0));
-    vec2 transformed = vec2((in_TexCoord0.x - 0.5) * pointWidth, (in_TexCoord0.y - 0.5) * pointHeight);
+    vec2 transformed = vec2(in_Position.x * pointWidth, in_Position.y * pointHeight);
     gl_Position.xy += transformed * gl_Position.w;
 }
--- a/src/celengine/modelgeometry.cpp
+++ b/src/celengine/modelgeometry.cpp
@@ -136,6 +136,7 @@ ModelGeometry::render(RenderContext& rc, double /* t */)
         for (unsigned int groupIndex = 0; groupIndex < mesh->getGroupCount(); ++groupIndex)
         {
             const Mesh::PrimitiveGroup* group = mesh->getGroup(groupIndex);
+            rc.updateShader(mesh->getVertexDescription(), group->prim);
 
             // Set up the material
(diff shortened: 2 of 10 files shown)


375gnu:
Hi @levinli303 , I think your PR is quite overengineered: you can simply reuse `PointStarVertexBuffer::startSprites()` but 
on https://github.com/CelestiaProject/Celestia/blob/master/src/celengine/pointstarvertexbuffer.cpp#L52 replace the following:
```
    glEnableVertexAttribArray(CelestiaGLProgram::PointSizeAttributeIndex);
    glVertexAttribPointer(CelestiaGLProgram::PointSizeAttributeIndex,
                          1, GL_FLOAT, GL_FALSE,
                          stride, &vertices[0].size);
```
with
```
  if (useSprites)
  {
    glEnableVertexAttribArray(CelestiaGLProgram::PointSizeAttributeIndex);
    glVertexAttribPointer(CelestiaGLProgram::PointSizeAttributeIndex,
                            1, GL_FLOAT, GL_FALSE,
                            stride, &vertices[0].size);
  }
  else
  {
    glVertexAttrib1f(CelestiaGLProgram::PointSizeAttributeIndex, requiredPointSize);
  }
```
Of course `glEnable(GL_VERTEX_PROGRAM_POINT_SIZE);` will be required still for desktop GL.

levinli303:
@375gnu not really, we the shaders generated with shaderproperties we have to also set pointScale to make it work. here we just want to display a point with a fixed size. but I modified the PR to remove enablePoint.

levinli303:
@375gnu  this change should be smaller now, just adding a staticPointSize in shader, and separate setVertexArrays in rend context into setVertexArrays and updateShader.

375gnu:
I think you can simplify it further. Instead of adding conditional variables to scale all dots I'd suggest adding a scale factor and multiply all dot sizes by it. afair we even have some code for this.

levinli303:
@375gnu , I removed staticPointSize and used glVertexAttrib1f to pass 1.0 for these points, and use the existing pointScale fort scale factor.

levinli303:
also fixed the bug where glare is not drawn because by calling 
```
    starRenderer.glareVertexBuffer->startSprites();
    if (starStyle == PointStars)
        starRenderer.starVertexBuffer->startBasicPoints();
    else
        starRenderer.starVertexBuffer->startSprites();
```
only one of the star renderer's shaders is used...

levinli303:
@375gnu can you review it again?

375gnu:
Hi @levinli303, i'm still not sure that adding a new condition is a good idea. why not to scale all points unconditionally?

levinli303:
@375gnu I think the problem is that we need to adjust point size and fade according to vertex position in the other case. As the methods I can think of, a flag needs to be passed in.

375gnu (approved):
okay, let's go with this

PR #887, opened 2020-12-03
Set up CI with Azure Pipelines

levinli303:


levinli303:
@375gnu  this is ready now, but I just created matrix for Linux/Mac, didn't create separate builds for Lua version, think it is unnecessary.

azure-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+                        libtheora-dev \
+                        libjpeg-dev \
+                        libpng-dev \
+                        libglu1-mesa-dev \
+                        qtbase5-dev \
+                        qtbase5-dev-tools \
+                        libqt5opengl5-dev \
+                        libfreetype6-dev \
+                        libsdl2-dev \
+                        liblua5.3-dev
```
375gnu, line 33: Could you replace `liblua5.3-dev` with `libluajit-5.1-dev`? In this case we'll be tests both 5.1 (on GNU/Linux) and 5.3 (on macOS).

375gnu (changes requested):

375gnu (approved):

PR #894, opened 2020-12-05
Look up illumination color from color table

### LINKED ISSUE #892: Reflected starlight color
(Not sure if this is a bug report or a feature request...)

This has been frequently mentioned on Discord but never here - the color of reflected starlight on planets and other objects is always the classic colors rather than the Blackbody D65 colors, regardless of the star color setting. (In 1.7.0 this is only the case when "Tinted illumination" is checked, otherwise everything is rendered in white light.)

**Expected behavior/requested feature:**

Reflected starlight on ssc objects should match the Blackbody D65 colors when the star color is set to Blackbody D65 (and "Tinted illumination" is checked).

levinli303:
Should close #892  ?

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -1404,6 +1404,7 @@ setupLightSources(const vector<const Star*>& nearStars,
                   const UniversalCoord& observerPos,
                   double t,
                   vector<LightSource>& lightSources,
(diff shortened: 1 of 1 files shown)


375gnu:
Result is not the same. I think we should first ask somebody with good knowledge of astronomy.

PR #895, opened 2020-12-09
Draw line with triangles

levinli303:


levinli303:
@375gnu it is not finished, but I think you can maybe take a look first. just replacing GL_LINES with GL_TRIANGLES, and GL_LINE_STRIP/LOOP with GL_TRIANGLE_STRIP and fill the buffer with more vertices info.

Seems straightforward, the only things is that GL_LINE_STRIP/LOOP cannot map directly to GL_TRIANGLE_STRIP with double the vertices. Like for drawing a rectangle, I still need to use GL_TRIANGLES because of the line joins, so I went ahead with GL_TRIANGLE_STRIP in only orbits and grids where we usually draw ellipses. does it sound good?

src/celengine/asterismrenderer.cpp
```diff
@@ -114,17 +115,18 @@ GLfloat* AsterismRenderer::prepare()
...
-            ptr += 3;
-            for (unsigned i = 1; i < chain.size() - 1; i++)
+            for (unsigned i = 1; i < chain.size(); i++)
             {
-                memcpy(ptr,     chain[i].data(), 3 * sizeof(float));
-                memcpy(ptr + 3, chain[i].data(), 3 * sizeof(float));
-                ptr += 6;
+                Eigen::Vector3f prev = chain[i - 1];
+                Eigen::Vector3f cur = chain[i];
+                data.push_back({prev, cur, -0.5});
```
375gnu, line 122: it's better to use `emplace_back` here. `data.emplace_back(prev, cur, -0.5);` it works as "placement **new**".

src/celengine/render.cpp
```diff
@@ -701,6 +701,31 @@ void Renderer::setScreenDpi(int _dpi)
...
 }
 
+float Renderer::getScaleFactor() const
+{
+    return screenDpi / 96.0f;
+}
+
+float Renderer::getPointWidth() const
+{
+    return 2.0 / windowWidth * getScaleFactor();
```
375gnu, line 711: don't forget about `f` suffix to remove unneeded type conversion

src/celengine/asterismrenderer.cpp
levinli303: it seems we need to create a constructor for the struct for this purpose? otherwise it complains no matching method.

375gnu: Yes, just checked our current usage of the method. `data.emplace_back(Eigen::Vector3f(prev, cur, -0.5));` and so on

levinli303: fixed.

levinli303:
basically converted everything except the ones in rendcontext.cpp (I believe it is used for models, well anyway I don't think lines would look good on models, are there models using it? the asteroid models using PointList look very bad from a far.)

levinli303:
yeah indeed looks bad, there should be a way to fix it.

I only tested on HiDPI with MSAA, it looked good there
Screen Shot 2020-12-12 at 11 45 41 PM

levinli303:
Screen Shot 2020-12-13 at 12 22 12 AM

@375gnu  weird, there seemed to be overlapping line segments from the data and it might be the cause? this is rendering constellation boundary

375gnu:
Unfortunately in lodpi env without msaa they look bad. And when you change line's tilt then the number of triangles it's tessellated into is changing so you see the rippling on the line.

Regarding your question - i tested it with the default data and even when only the single type of lines is show (ecliptic or a grid) I still see those overlapping-like patterns.

Can we have this code hidden behind ifdefs so only platforms that require it have it enabled? I assume only some apple products and emscripten need this.

levinli303:
lets try to use celgl::maxLineWidth instead.

375gnu:
let me check this

375gnu:
ecliptic line has an incorrect position, upon application start when you see the earth it should be horisontal

375gnu:
ecliptic line has an incorrect position, upon application start when you see the earth it should be horizontal

levinli303:
> ecliptic line has an incorrect position, upon application start when you see the earth it should be horizontal

fixed.

375gnu (approved):

Issue #544, opened 2019-11-29
[post 1.7] Improve galaxy and globular rendering


SevenSpheres:
Note: this is also something for after the 1.7 release.

See #524, and celestia.Sci

LukeCEL:
Maybe we should split this into several issues, since there are several problems with galaxy and globular rendering right now.

SevenSpheres:
If you can list specific problems I can make a checklist and/or additional issues.

LukeCEL:
GLOBULAR CLUSTERS:
- [ ] Individual sprites don't look very much like the stars in Celestia—they look too dim and dull.
- [ ] Some sprites look green, which shouldn't happen. It seems to depend on the KingConcentration parameter, and it's most obvious when KingConcentration ≈ 0.90. For an example, take a look at NGC 6712:
![ngc6712](https://user-images.githubusercontent.com/45546434/69891594-93780980-12cb-11ea-8dc7-b0ccbce8dd0c.png)
- [ ] There needs to be some way to make globular clusters elongated: #543 

I can add more later.

LukeCEL:
- [ ] Globular clusters are too bright? #524 That being said, I can't seem to reproduce this on my build—they look fine to me.

GALAXIES:
- [ ] Elliptical galaxies don't look very similar to what they should look like. They should have a smooth, featureless profile and have a yellowish tone.
- [ ] Lenticular galaxies as well—they're quite diverse in appearance, but in general, they don't look as blue as spiral galaxies.
- [ ] Irregular galaxies need some way to make them elongated, sort of like E1-6 galaxies in 1.6.1. This applies specifically to galaxies like Wolf-Lundmark-Melotte.
- [ ] Specific to 1.7: Changing the light gain turns on and off sprites, so there's a weird twinkling lights effect.

SevenSpheres:
@LukeCEL I put your checklist in the OP. @375gnu should I split these out into separate issues or leave them here?

375gnu:
Let's have them all together for a while.

> Some sprites look green, which shouldn't happen.

t00fri had an argue with @claurel regarding this issue you can try to find it on the forum.

LukeCEL:
> t00fri had an argue with @claurel regarding this issue, you can try to find it on the forum. t00fri said this is correct colour.

Yeah, I saw that argument, but I couldn't really tell what he was saying. I think I might ask him on the Celestial Matters forums, but even the CM forums are having problems:
[...]

PR #897, opened 2020-12-14
Fix galaxy flickering on zooming in

levinli303:
with short, the range is from -32,768 to 32,767. we should use unsigned short instead.

mentioned by @LukeCEL in #544

### DIFF
--- a/src/celengine/galaxy.cpp
+++ b/src/celengine/galaxy.cpp
@@ -292,7 +292,7 @@ struct GalaxyVertex
     EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
     Vector4f position;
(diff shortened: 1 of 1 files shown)


LukeCEL:
Unfortunately when I tested it, the flickering effect is still there.

https://user-images.githubusercontent.com/45546434/103020888-46d72000-4517-11eb-957c-40404d11dde8.mov

levinli303:
@LukeCEL I just pushed new changes, should have been fixed now.

LukeCEL:
Yes, it appears to be fixed now!

375gnu (approved):
i'm dumb ass

PR #906, opened 2021-01-04
Remove thirdparty submodules

levinli303:
just use the system ones.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,10 +1,3 @@
-[submodule "thirdparty/Eigen"]
-	path = thirdparty/Eigen
-	url = https://github.com/eigenteam/eigen-git-mirror
-	branch = branches/3.3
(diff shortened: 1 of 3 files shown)


levinli303:
uh, it seems cspice is not offered in apt as a package.

munix9:
fmt also seems to be mandatory
By not providing "Findfmt.cmake" in CMAKE_MODULE_PATH this project has ...

levinli303:
> fmt also seems to be mandatory
> By not providing "Findfmt.cmake" in CMAKE_MODULE_PATH this project has ...

fixed now, install libfmt-dev on Ubuntu

375gnu:
this is too early. let's keep them for 6-12 month

375gnu:
let's go!

375gnu (approved):

PR #907, opened 2021-01-05
Fix point sprite not scaled in models

levinli303:
well...


375gnu (approved):
though i did test this

Issue #911, opened 2021-01-14
Debian_10 build failed on OBS: dpkg-source: error: can't build with source format '3.0 (native)'


munix9:
**Describe the bug**
Debian_10 x86_64 v1.7 master build on OBS failed:
`dpkg-source: error: can't build with source format '3.0 (native)': native package version may not have a revision`
Ubuntu builds are fine.

**To Reproduce**
See https://build.opensuse.org/package/live_build_log/home:munix9:unstable/celestia/Debian_10/x86_64

**Expected behavior**
Debian_10 build should not fail.
Maybe a problem with latest updates for Debian_10 and the definition in
https://github.com/CelestiaProject/Celestia/blob/master/debian/source/format
(should be 3.0 (quilt) ?!).

**Screenshots**

**Desktop (please complete the following information):**
Debian_10 build on OBS.

**Additional context**

375gnu:
Hi @munix9 ,

This is a problem on OBS side:
```
[  180s] Compressed celestia-1.7.0~git20210114+5360a165e.tar to celestia-1.7.0~git20210114+5360a165e.tar.xz
[  181s] /usr/lib/obs/service/set_version:33: RuntimeWarning: install 'packaging' to improve python package versions
[  181s]   RuntimeWarning)
[  181s] Found files matching debian.*, running debian transformer...
[  181s] release: (385.1), release (DEB) (385.1)
[  181s] ** Started: debtransform --release 385.1 //usr/src/packages/SOURCES //usr/src/packages/SOURCES/celestia.dsc //usr/src/packages/SOURCES.DEB
[  181s] No DEBTRANSFORM-TAR line in the .dsc file.
[  181s] Attempting automatic discovery of a suitable source archive.
[  181s] Source archive chosen for transformation: celestia-1.7.0~git20210114+5360a165e.tar.xz
[  181s] Transforming into source package 1.0 (non-native) format
[  181s] Added Debian revision to Version field, which is now "1.7.0~git20210114+5360a165e-1".
[  181s] converting //usr/src/packages/SOURCES/celestia-1.7.0~git20210114+5360a165e.tar.xz to celestia-1.7.0~git20210114+5360a165e.tar.gz
[  183s] Modifying dsc Version field to "1.7.0~git20210114+5360a165e-1"
[...]

PR #912, opened 2021-01-15
Update debian/source/format

munix9:
Replace `native` with `quilt` in `debian/source/format` #911


375gnu:
this is a workaround for OBS only, but it will break other builds because this package has native format.

PR #913, opened 2021-02-05
Add proper temperatures for WC and WN stars

### LINKED ISSUE #901: Proper WR temperatures
As you may know, I am obsessed with Wolf-Rayet stars. I'm a bit frustrated whenever I play around with WRs and see that they're so much cooler than they should be. For example, HIP 5100 (WR 2) in Celestia has a temperature of 50,000 K, while in reality that temperature is over 140,000 K. 

I would like the Wolf-Rayet temperatures to be raised, to reflect their real-life temperatures. Note: This is for the hydrogen-free Wolf-Rayet stars. For the hydrogen-rich ones (WNh stars) I would like a separate class to be created that has temperatures that reflect the WNh stars.

Sometimes, I use the hottest white dwarf (D0) which has a temperature of 100,000 K, to depict hot Wolf-Rayet stars. This works to an extent, but not very well for WRs between 60kK and 100kK, and ones between 110kK and 130kK.

Possible temperatures for WRs:

static float tempWN[10] =
{
    155000, 150000, 140000, 115000, 89000, 75000, 65000, 56000, 50000, 45000
};

GurrenLagannTSS:
Partially implements and closes #901, adding WNh as a completely new class is a bit of an hassle for me.

### DIFF
--- a/src/celengine/star.cpp
+++ b/src/celengine/star.cpp
@@ -105,16 +105,15 @@ static float tempM[3][10] =
     { 3650, 3550, 3450, 3200, 2980, 2800, 2600, 2600, 2600, 2600 },
 };
 
-// Wolf-Rayet temperatures. From Lang's Astrophysical Data: Planets and
-// Stars.
+// Wolf-Rayet temperatures.
 static float tempWN[10] =
 {
-    50000, 50000, 50000, 50000, 47000, 43000, 39000, 32000, 29000, 29000
+    185000, 158000, 140000, 115000, 89000, 75000, 65000, 56000, 50000, 45000
 };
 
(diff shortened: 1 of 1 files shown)


375gnu:
1) All commits should be squashed into a single one
2) Reference to the sources should be left in comments so we can validate temperatures and so on

Trolligi112477-ww:
Sources: https://arxiv.org/pdf/1904.04687.pdf
https://arxiv.org/pdf/1807.04293.pdf
Wikipedia
Logic

oh wait... maybe the WN5-WN6 temps should be lowered

AstroChara:
Hang on, what's the current issue with this thing?

PR #915, opened 2021-02-12
Update star data

SevenSpheres:
Update of temperatures and bolometric corrections for all main-sequence stars and brown dwarfs. Also updated temperature for the Sun.

Might as well update *everything*. Not done yet. Source used is [Mamajek's table](https://www.pas.rochester.edu/~emamajek/EEM_dwarf_UBVIJHK_colors_Teff.txt).

### DIFF
--- a/src/celengine/star.cpp
+++ b/src/celengine/star.cpp
@@ -1,6 +1,6 @@
 // star.cpp
 //
-// Copyright (C) 2001-2019, the Celestia Development Team
+// Copyright (C) 2001-2021, the Celestia Development Team
 // Original version by Chris Laurel <claurel@gmail.com>
 //
 // This program is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


SevenSpheres:
Was leaving as draft because the brown dwarf bolometric corrections aren't perfect, but it should be good to merge

src/celengine/star.h
```diff
@@ -229,8 +229,9 @@ StarDetails::getEllipsoidSemiAxes() const
 bool
 StarDetails::hasCorona() const
 {
-    // Y dwarves and T dwarves subclasses 5-9 don't have a corona
-    return spectralType[0] != 'Y' && (spectralType[0] != 'T' || spectralType[1] < '5');
+    // Don't render corona for brown dwarfs that are not hot enough to glow.
+    // See https://en.wikipedia.org/wiki/Draper_point
+    return temperature >= 798;
```
375gnu, line 234: i don't think this law can be applied here

375gnu (changes requested):

src/celengine/star.h
SevenSpheres: It's better than an arbitrary cutoff at T5.

375gnu: Sun's corona has temperature of millions of K

375gnu (changes requested):

src/celengine/star.h
Askaniy: > The absence of X-rays from LP 944-20 during the non flaring period is also a significant result. It sets the lowest observational limit on steady X-ray power produced by a brown dwarf star, and shows that million degree Celsius upper atmospheres, or coronas, cease to exist as the surface temperature of a brown dwarf cools below about 2500 degrees Celsius and becomes electrically neutral.

https://chandra.harvard.edu/xray_sources/browndwarf_fg.html

375gnu: @SevenSpheres do you have any updates for this pr?

SevenSpheres: Removed that commit, will open a separate issue later

Askaniy (approved):

375gnu (approved):

LukeCEL (approved):

PR #920, opened 2021-02-18
Call flush in end and only bind buffer once

levinli303:
Should fix this crash:
https://appcenter.ms/orgs/CelestiaProject/apps/Celestia-1/crashes/errors/1946065538u/overview


375gnu (approved):

levinli303:
I've released a 1.2.5 beta (on Play) to test if the crash has been fixed, will merge when it gets confirmed.

PR #923, opened 2021-03-01
Fix crash due to use of uninitialized memory

levinli303:
This might be the real cause of issues mentioned in https://github.com/CelestiaProject/Celestia/pull/922 I can reliably reproduce by turning on boundary rendering


375gnu (approved):

375gnu:
@levinli303 should we then revert your previous attempts to fix the issue?

levinli303:
I think not, those should be fixed as well.

PR #924, opened 2021-03-02
C++17-like features plus fixes

375gnu:
This PR provides some C++17-compatible libraries:
 * `string_view`, very useful as non-owning, non-allocation "view" of strings, both std::string and char[]. Safer replacement for `const char*`.
 * `charconv`, which provides `from_chars` function to convert numbers from strings. `to_chars` are not implemented yet.

One small utility library `array_view`. Like `string_view` but for arrays. With it we can have a single function which accepts C-style arrays, `std::array` and `std::vector`.

Then, use `std::tie` to compare structures. With `-O2` generated code looks the same but the sources are more readable and less error prune.

Fix for `LuaState::loadScript()`. This method accepts a string with a Lua script, but when transition to `fs` was made the string was replaced with `fs::path`.

Fix for `RenderFlags`. `ShowPlanetRings` was added instead of removed `ShowStarsAsPoints`, so compatibility with 1.6 version was lost. Let's restore it.

Old Celestia versions were able to blacklist GL extensions in `celestia.cfg`. When transition to `GLEW` was made this feature was lost. Let's restore it.

And at last fix warnings shown by `MinGW`, I use it to cross-compile windows backend.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -390,6 +390,9 @@ if((NOT HAVE_FILESYSTEM) AND HAVE_EXPERIMENTAL_FILESYSTEM)
   endif()
 endif()
 
+try_compile(HAVE_STRING_VIEW ${CMAKE_BINARY_DIR} "${CMAKE_SOURCE_DIR}/checks/cxxsv.cpp")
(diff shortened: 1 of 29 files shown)


levinli303 (approved):
i should definitely learn some modern cpp 😄

PR #928, opened 2021-03-09
Port garbled text fix from 1.6.x

levinli303:


levinli303:
Hmm looks like it does not work...

PR #929, opened 2021-03-12
Various fix

levinli303:
1. Make dds_decompress .cpp instead of .c
2. Use string_view.data instead of .begin and .end (MSVC throws error using .begin, .end)
3. Fix for lua 5.4


375gnu (approved):

CMakeLists.txt
```diff
@@ -111,7 +111,8 @@ if(MSVC)
   # bigobj: generate more object sections than allowed by default
   # Promoted to errors
   # C4316: object allocated on the heap may not be aligned 16
-  add_compile_options("/wd4244" "/wd4267" "/wd4503" "/wd4800" "/wd4455" "/wd5030" "/bigobj" "/we4316")
+  # /Zc:__cplusplus: correctly set __cplusplus macro value
+  add_compile_options("/wd4244" "/wd4267" "/wd4503" "/wd4800" "/wd4455" "/wd5030" "/bigobj" "/we4316" "/Zc:__cplusplus")
```
375gnu, line 115: please note that this option is supported by VC++ 2019 only

levinli303:
Also changed g_vertices to a pointer instead since other wise we need to use emplace_back to replace this, and we need to clear the vector all the time.
```
            g_vertices[vertex++] = { p + v0, { 0, 0, color, alpha } };
            g_vertices[vertex++] = { p + v1, { 1, 0, color, alpha } };
            g_vertices[vertex++] = { p + v2, { 1, 1, color, alpha } };
            g_vertices[vertex++] = { p + v3, { 0, 1, color, alpha } };
```

CMakeLists.txt
levinli303: added a check if(MSVC_VERSION GREATER_EQUAL 1914)

375gnu (approved):

PR #930, opened 2021-03-14
Some more platform dependent fixes

levinli303:
1. Move UI related code in celutil/winutil to celestia/win32
2. Add cctype include for toupper
3. Add some utils for apple platform and undef functions for portable builds

### DIFF
--- a/src/celutil/appleutils.h
+++ b/src/celutil/appleutils.h
@@ -1,0 +1,17 @@
+// appleutils.h
+//
+// Copyright (C) 2021-present, Celestia Development Team.
+//
+// Miscellaneous useful Apple platform functions.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#pragma once
+
+#include <string>
+
+std::string AppleHomeDirectory();
+std::string AppleApplicationSupportDirectory();
--- a/src/celutil/appleutils.h
+++ b/src/celutil/appleutils.h
@@ -11,7 +11,7 @@
 
 #pragma once
 
-#include <string>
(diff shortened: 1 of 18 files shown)


src/celutil/appleutils.h
```diff
@@ -0,0 +1,17 @@
...
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#pragma once
+
+#include <string>
+
+std::string AppleHomeDirectory();
```
375gnu, line 16: could you use fs::path here instead?

375gnu (changes requested):

src/celutil/utf8.cpp
```diff
@@ -961,8 +961,8 @@ static size_t greekChunkLength(const std::string& str)
         npos = true;
     }
 
-    if (sp != 0 && isdigit(str[sp - 1]))
-        while(sp != 0 && isdigit(str[sp - 1])) sp--;
+    if (sp != 0 && str[sp - 1] >= 0 && isdigit(str[sp - 1]))
```
375gnu, line 964: you are doing something strange here.

375gnu (changes requested):

src/celutil/utf8.cpp
levinli303: I ran into a crash here with MSVC (debug). passing a negative value to isdigit aborts the application.

375gnu: then the correct way it to cast to uchar:
```
if (sp != 0 && isdigit(static_cast<unsigned char>(str[sp - 1])))
    while(sp != 0 && isdigit(static_cast<unsigned char>(str[sp - 1]))) sp--;
```

375gnu (approved):

PR #932, opened 2021-03-25
Some small fixes

levinli303:
1. Use quoted include for lua.hpp
2. Fix compilation with std filesystem

### DIFF
--- a/src/celengine/skygrid.cpp
+++ b/src/celengine/skygrid.cpp
@@ -569,9 +569,7 @@ SkyGrid::render(Renderer& renderer,
 
     double arcStep = (maxTheta - minTheta) / (double) ARC_SUBDIVISIONS;
     double theta0 = minTheta;
-
-    vector<LineStripEnd> buffer;
-    buffer.reserve(2 * (ARC_SUBDIVISIONS + 2));
+    static vector<LineStripEnd> buffer(2 * (ARC_SUBDIVISIONS + 2), { Vector3f::Identity(), 1.0f });
     glEnableVertexAttribArray(CelestiaGLProgram::VertexCoordAttributeIndex);
     if (lineAsTriangles)
     {
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -1880,7 +1880,8 @@ void CelestiaCore::start(double t)
         // using the KdeAlerter in runScript would create an infinite loop,
         // break it here by resetting config->initScriptFile:
         fs::path filename(config->initScriptFile);
-        config->initScriptFile = {};
+        // Don't use {} as it will throw error C2593 on MSVC
+        config->initScriptFile = fs::path();
         runScript(filename);
     }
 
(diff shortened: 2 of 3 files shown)


src/celestia/celestiacore.cpp
```diff
@@ -1880,7 +1880,7 @@ void CelestiaCore::start(double t)
         // using the KdeAlerter in runScript would create an infinite loop,
         // break it here by resetting config->initScriptFile:
         fs::path filename(config->initScriptFile);
-        config->initScriptFile = {};
+        config->initScriptFile = "";
```
375gnu, line 1883: what errors do you have with `{}`?

levinli303: celestiacore.cpp(1902,35): error C2593: 'operator =' is ambiguous
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.28.29910\include\filesystem(737,15): message : could be 'std::filesystem::path &std::filesystem::path::operator =(std::filesystem::path::string_type &&) noexcept'
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.28.29910\include\filesystem(701,15): message : or       'std::filesystem::path &std::filesystem::path::operator =(std::filesystem::path &&) noexcept'

375gnu: could you leave a comment then so we won't change it back again? something like `don't change "" to {} or MSVC2019 throw error C2593`.
and probably better to use `fs::path()` instead of `""`.

src/celestia/celestiacore.cpp
```diff
@@ -1899,7 +1899,7 @@ void CelestiaCore::setStartURL(const string &url)
     if (!url.substr(0, 4).compare("cel:"))
     {
         startURL = url;
-        config->initScriptFile = {};
+        config->initScriptFile = "";
```
375gnu, line 1902: as above

375gnu (changes requested):

src/celestia/celestiacore.cpp
levinli303: fixed

375gnu (approved):

levinli303:
@375gnu  I pushed another commit, fix the last place I spotted using vector.reserve + assign by index.

src/celengine/skygrid.cpp
```diff
@@ -569,9 +569,7 @@ SkyGrid::render(Renderer& renderer,
 
     double arcStep = (maxTheta - minTheta) / (double) ARC_SUBDIVISIONS;
     double theta0 = minTheta;
-
-    vector<LineStripEnd> buffer;
-    buffer.reserve(2 * (ARC_SUBDIVISIONS + 2));
+    static vector<LineStripEnd> buffer(2 * (ARC_SUBDIVISIONS + 2), { Vector3f::Identity(), 1.0f });
```
375gnu, line 572: what's the problem with the current code apart that it's slower?

levinli303: The current one uses .reserve and then assigns value like [0] [1] directly. This causes undefined behavior (crashes app built with MSVC debug)

375gnu: okay. clear. so it's better to use memory chunk allocated with `new`, but let's live with this

375gnu (approved):

PR #939, opened 2021-04-03
Image refactoring

375gnu:
move image-related files to own dirs. split different formats from image.cpp. and similar changes for imagecapture.cpp


375gnu:
actually i'm in doubt whether it's a good idea to move image.cpp/image.h to celimage. probably it's better to keep them in celengine and separate only formats support.

375gnu:
restored image.* in celengine/, @levinli303 please review

levinli303 (approved):

PR #984, opened 2021-05-09
Make theora and gtk usable on Mac

levinli303:


levinli303:
Screen Shot 2021-05-09 at 2 15 25 PM

Although scaling is incorrect.

PR #986, opened 2021-05-09
Update CMakeLists.txt

levinli303:
Require CMake 3.13 to build


375gnu:
I'd rather prefer to rewrite gtk/CMakeLists.txt to be compatible with older systems

375gnu:
superseded by #988

PR #996, opened 2021-05-13
i18n for star and dsos in code

levinli303:


375gnu (approved):
i haven't tested but looks good

PR #1005, opened 2021-05-14
Support HiDPI on QT front end

levinli303:


src/celestia/qt/qtglwidget.cpp
```diff
@@ -117,6 +117,8 @@ static GLContext::GLRenderPath getBestAvailableRenderPath(const GLContext& /*glc
 
 void CelestiaGlWidget::initializeGL()
 {
+    appCore->setScreenDpi(logicalDpiY() * devicePixelRatioF());
```
levinli303, line 120: is it better to use a fixed value 96 here? on Mac `logicalDpiY()` produces 72, which still makes everything looks smaller than it should be.

375gnu: i have no idea as i don't have any highdpi devices to test

levinli303:
Looks good on Windows,  the one in the background

Screen Shot 2021-05-14 at 8 12 42 PM

PR #1012, opened 2021-05-22
Namespacing Marker* and Rect

levinli303:
I had trouble compiling with these two when using a different configuration on Mac.


375gnu (changes requested):
please celestia namespace

375gnu (approved):

PR #1019, opened 2021-05-24
Remove asset files duplicated in CelestiaContent

ajtribick:
Removes the asset files moved into CelestiaContent, and updates the relevant paths in the celestia.cfg file, the various resource managers, and the Inno Setup Script.

Depends on https://github.com/CelestiaProject/CelestiaContent/pull/42 otherwise the install location on Linux/Mac won't work.


ajtribick:
To make this easier to review: the first commit removes the data, extras, extras-standard, models and textures directories. The second commit contains the file path changes in the source code and celestia.cfg file. This should help managing the number of changed files.

375gnu:
@ajtribick what's wrong with this pr?

ajtribick:
I thought I had a better solution until I realised that symbolic links really don't play nicely with Windows. Have raised a simpler PR that only does the removal of the assets, in case people have strong opinions on the directory structure of the installed application.

PR #1021, opened 2021-05-25
Add back support for URL version 4

levinli303:
Since many mobile created URLs are using v4, probably we should keep it for a while.


375gnu (approved):
but we should define eol time for it

375gnu (changes requested):
wait a bit, something is missing here

levinli303:
yeah it seems the 27th bit was not handled, just corrected it.

375gnu (approved):

Issue #1020, opened 2021-05-24
cotire is unsupported since December 2019


ajtribick:
Ran into this while investigating build breaks using Visual Studio Code on Linux (that defaults to using ninja for its cmake builds), which turned out to be related to the use of cotire. According to the README at https://github.com/sakra/cotire/

> The functionality provided by cotire has been superseded by features added to CMake 3.16. Support for pre-compiling and unity builds is now built into CMake. Thus, there will not be any further updates or support for this project.

It would be worth considering removing cotire from the build process.

375gnu:
I actually do this in my WIP branch which uses ffmpeg to save captured video.

Issue #83, opened 2018-07-24
Captured ogg video is broken


375gnu:
```[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
V: 00:00:30 / 00:00:35 (99%)
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
V: 00:00:30 / 00:00:35 (99%)
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
V: 00:00:30 / 00:00:35 (99%)
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
V: 00:00:31 / 00:00:35 (99%)
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
[ffmpeg/demuxer] ogg: Broken file, non-keyframe not correctly marked.
V: 00:00:34 / 00:00:35 (99%)
```

levinli303:
I feel that maybe we can remove video capturing from Celestia, thus no need for ogg/theora dependency, most modern systems (I could be wrong) have built in apps/functionality for screen/window recording. @CelestiaProject/devs

375gnu:
i have an oldstanding branch with ffmpeg used instead of libtheora/windows-native. i hope one day i can finish it.

Issue #82, opened 2018-07-20
Video capture crashes with 64-bit on Windows


GottfriedSp:
Video capture crashes with 64-bit on Windows when stop capture with F12

375gnu:
Please provide more info (unless you are going to fix it yourself).

GottfriedSp:
Windows 32 works fine.
First starting video capture with SHIFT+F10
F11 start capture
F12 record end:
 - crash in CelestiaCore::recordEnd()
     `if (movieCapture != NULL)` << here
backtrace:
	ntdll.dll!0000000077b6b1f0()	Unknown
 	ntdll.dll!0000000077bf9778()	Unknown
 	kernel32.dll!00000000778f59cd()	Unknown
 	ntdll.dll!0000000077b5383d()	Unknown

backtrace with symbols:
>	ntdll.dll!DbgBreakPoint()	Unknown
 	ntdll.dll!DbgUiRemoteBreakin()	Unknown
 	kernel32.dll!BaseThreadInitThunk()	Unknown
 	ntdll.dll!RtlUserThreadStart()	Unknown

It crashes before the CelestiaCore::keyDown(int key, int modifiers) is called.
For Windows it uses class AVICapture

I don't know where this crash is comming from

375gnu:
Please build it with debug information enabled (the simplest way is to add `-g` of MSVC equivalent to CXXFLAGS) then run it in debugger. In case of gdb this will be

```
gdb ./Celestia_QT
run
bt
```

That you'll have a better stacktrace.

375gnu:
sounds like variable/array/pointer size mismatch

375gnu:
Current build doesn't crash.

PR #1022, opened 2021-05-26
Use ffmpeg to capture video

375gnu:
Closes: #1020 #82 #83


375gnu:
appveyor is annoying me. will need to use azure pipelines for windows as well

levinli303:
I tested with QT, it showed a directory picker panel instead of a save panel.

375gnu:
@levinli303 this is unrelated bug existed in original code, but i pushed a commit to address it. could you verify if it works for you?

levinli303 (approved):

PR #1024, opened 2021-05-27
Remove duplicated assets

ajtribick:
Removes the data, extras, extras-standard, models and textures directories, as these are now in the CelestiaContent repository. This removes the part of the previous pull requests that would have changed the install structure. I had looked into replacing these directories with symbolic links but that does not play nicely with Windows, where symbolic link creation requires admin privileges.


375gnu (approved):

levinli303 (approved):

levinli303:
looks good, we can remove the flag ENABLE_DATA now.

375gnu:
> looks good, we can remove the flag ENABLE_DATA now.

no, ENABLE_DATA is for submodule.

PR #1026, opened 2021-05-30
Fix compatibility with QT6

levinli303:
It seems we would also have to use QOpenGLWidget on QT6, but I just left it out here.

### DIFF
--- a/src/celestia/qt/qtglwidget.cpp
+++ b/src/celestia/qt/qtglwidget.cpp
@@ -24,6 +24,7 @@
 #include <QPaintDevice>
 #include <QMouseEvent>
 #include <QSettings>
+#include <QMessageBox>
 
 #ifndef DEBUG
 #  define G_DISABLE_ASSERT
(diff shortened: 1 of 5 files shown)


src/celestia/qt/qtglwidget.cpp
```diff
@@ -442,18 +450,18 @@ void CelestiaGlWidget::keyPressEvent( QKeyEvent* e )
...
-            if ((e->text() != 0) && (e->text() != ""))
+            if (e->text() != "")
             {
                 QString input = e->text();
 #ifdef __APPLE__
                 // Taken from the macOS project
                 if (input.length() == 1)
                 {
-                    QChar c = input.at(0);
+                    ushort c = input.at(0).unicode();
```
375gnu, line 460: ushort isn't a standard type

src/celestia/qt/qtglwidget.cpp
```diff
@@ -442,18 +450,18 @@ void CelestiaGlWidget::keyPressEvent( QKeyEvent* e )
     default:
         if (!handleSpecialKey(e, true))
         {
-            if ((e->text() != 0) && (e->text() != ""))
+            if (e->text() != "")
```
375gnu, line 453: ! `.isEmpty()`?

375gnu (changes requested):

375gnu:
QOpenGLWidget is buggy in fullscreen. at least in 5.x

375gnu (approved):

PR #1086, opened 2021-06-13
Remove alpha channel for DXT1 textures

levinli303:
To align with the behaviors of loading DXT1 textures with gl methods. Takin the easy approach here...

### DIFF
--- a/src/celimage/dds.cpp
+++ b/src/celimage/dds.cpp
@@ -239,7 +239,7 @@ Image* LoadDDSImage(const fs::path& filename)
     {
         if (!gl::EXT_texture_compression_s3tc)
         {
-            // DXTc texture not supported, decompress DXTc to RGBA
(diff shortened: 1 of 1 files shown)


375gnu:
@levinli303 could you explain purpose of the change? i'm not a specialist in DXT textures

levinli303:
Since in some places when hasAlpha is true on a texture, different render configuration is applied. the gl loaded dxt1 texture' returns false for hasAlpha. while after being transformed to RGBA, the hasAlpha value is true.

levinli303:
Before
Simulator Screen Shot - iPhone 12 - 2021-06-14 at 09 37 39

After
Simulator Screen Shot - iPhone 12 - 2021-06-14 at 09 36 33

375gnu:
@levinli303 could you then leave a comment with a reference to this PR in the body on your change?

levinli303:
done

375gnu (approved):

PR #1094, opened 2021-06-22
[ci] build ffmpeg with x264

375gnu:


levinli303 (approved):

375gnu:
either doesn't work or uneeded

PR #1096, opened 2021-06-26
Do not rely on clamp to border on ring textures

levinli303:


375gnu:
i'll test it first

375gnu:
@levinli303 could you provide info where you noticed issues without working clamp_to_border? i want to check if other approaches work?

levinli303:
You can check Uranus/Neptune with ring shadow on.... when clamp_to_border is not available.

375gnu (changes requested):
@levinli303 please amend your change to be applied only if GL_ES & GL_OES_texture_border_clamp doesn't present?

Because the proposed approach is visually worth (jugged edges) then the original. And older HW doesn't likes branching, so let's don't make worth where we can have a better result.

levinli303:
My Android device has GL_EXT_texture_border_clamp, but not GL_OES_texture_border_clamp, so I made it detect either GL_EXT_texture_border_clamp or GL_OES_texture_border_clamp

375gnu (approved):

levinli303:
added a "\n", I tested again and I did noticed that on iOS, sometimes some pixels around the ring shadow edge are flickering, wonder why it is there...

levinli303:
I'll merge first, it is better than a non-functional one.

PR #1099, opened 2021-06-26
make it compile with fmt v8

joequant:
This includes a patch that allows compile with fmt v8


375gnu (approved):

375gnu:
thank you for the contribution!

PR #1100, opened 2021-07-04
common command line parser

375gnu:
@levinli303 what do you think about this? my attempt is to remove copy-pasted command line parsing and keep the its common part to celestiacore. it's tested with qt/gtk/win. not finished yet. i want to hear your feedback. it also contains some tiny unrelated changes i will file them as separate prs.


levinli303:
since my frontends do not really accept command line arguments, im fine with the changes as long as we still expose interfaces to load from specific .cfg and extra dirs.

Also I wonder if we can move away from current directory to load from, would it make sense for all files to be loaded according to an absolute path?

375gnu:
> Also I wonder if we can move away from current directory to load from, would it make sense for all files to be loaded according to an absolute path?

We can use absolute paths in many (or even all) places of celestia.cfg. Or do you mean something else?

levinli303:
I'm saying now (on my frontends), I need to call chdir (or similar APIs) to change working directory so the relative path in data files in cfg can be correctly loaded. is it possible to avoid it. (maybe store a value of the root path for relative paths to load from?)

375gnu:
chdir approach is much simpler, in this pr i want to move it to CelestiaCore

375gnu:
no much sense with only 2 active ui

PR #1107, opened 2021-07-22
Refine the URL of the Spice submodule

podsvirov:


375gnu:
it works fine without `.git`. did you have any situation when it didn't work?

podsvirov:
Yes, it works, but this is not the recommended address for code cloning.
I propose this for unification.
For an example, look at the address of the `content` submodule in the `.gitmodules` file.
Why are these differences?

375gnu (approved):

PR #1114, opened 2021-09-05
lunar-lambert & specular

375gnu:


src/celengine/shadermanager.h
```diff
@@ -75,14 +75,14 @@ class ShaderProperties
      DiffuseModel          = 0,
```
levinli303, line 75: should start from 1?

375gnu: yeah, this would be better

src/celengine/shadermanager.cpp
```diff
@@ -277,14 +277,7 @@ ShaderProperties::hasSharedTextureCoords() const
...
 {
-    switch (lightModel)
-    {
-    case SpecularModel:
-    case PerPixelSpecularModel:
-        return true;
-    default:
-        return false;
-    }
+    return lightModel == SpecularModel || (lightModel & PerPixelSpecularModel) != 0;
```
levinli303, line 280: (lightModel & SpecularModel) != 0 ？

levinli303 (approved):

PR #1120, opened 2021-09-27
Translate well known star and dsos

levinli303:


levinli303:
@375gnu  this is a rough version with translations added to the end of .po files, it should work without the submodule update.

375gnu (approved):

PR #1121, opened 2021-09-27
Add support for imperial units

levinli303:
Had quite a few reviews from Play asking for supporting imperial units. These units are added but I'm not sure if they are the common ones.

mi, ft for length (km, m)
mi/s, ft/s for speed (km/s, m/s)
slug/ft^3 for density (1000 kg/m^3)
lb for mass (kg)

@SevenSpheres

### DIFF
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -86,7 +86,10 @@ static const float MaximumFOVFisheye = degToRad(179.99f);
 static const float MinimumFOV = degToRad(0.001f);
 static float KeyRotationAccel = degToRad(120.0f);
 static float MouseRotationSensitivity = degToRad(1.0f);
-
+static const double OneMiInKm = 1.609344;
+static const double OneFtInKm = 0.0003048;
+static const double OneLbInKg = 0.45359237;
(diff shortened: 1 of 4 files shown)


SevenSpheres (dismissed):
> Units of slugs are not in common use. Pounds are (mis)used instead.
> Units of yards/sec are never used. Feet/sec are used instead. [I see this was already changed though]

(I'm assuming imperial units are optional and not displayed by default, correct?)

SevenSpheres:
> Should i use lb/ft^3 for density instead?

I guess so.

src/celestia/celestiacore.cpp
```diff
@@ -3920,6 +3969,16 @@ bool CelestiaCore::initSimulation(const fs::path& configFileName,
             DPRINTF(LOG_LEVEL_WARNING, "Unknown viewport effect %s\n", config->viewportEffect);
     }
 
+    if (!config->measurementSystem.empty())
+    {
+        if (config->measurementSystem == "Imperial")
```
375gnu, line 3974: too fragile, could you make it case-insensitive?

375gnu (changes requested):

375gnu (approved):

PR #1134, opened 2021-10-16
Rewrite tokenizer to skip UTF-8 BOM and use standard library float parsing utilities

### LINKED ISSUE #1133: Skip UTF-8 pseudo-BOM when parsing text files
**Describe the bug**
While Celestia seems to be able to handle UTF-8 input files, if the "BOM" is present at the start of the file (bytes 0xef 0xbb 0xbf), it will report "Bad character in stream". Some editors (particularly Windows) tend to like saving UTF-8 with the BOM, so it would be good to take account of this possibility.

**To Reproduce**
Save any of Celestia's .ssc/.stc/.dsc files with the UTF-8 BOM present then start Celestia. The file will be skipped and the error "Bad character in stream" will be reported in the log.

**Expected behavior**
UTF-8 BOM is ignored while parsing the files.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: affects all
 - Frontend: affects all
 - Version: tested on commit e7ee8ba045f7768da0af2fcfaf31b1fa64b1c18a

**Additional context**
Looks like the fix should be done in src/celengine/tokenizer.cpp, not sure how this interacts with the implementation-defined signedness of `char`.

ajtribick:
Resolves #1133

### DIFF
--- a/src/celengine/tokenizer.cpp
+++ b/src/celengine/tokenizer.cpp
@@ -60,6 +60,16 @@ Tokenizer::TokenType Tokenizer::nextToken()
         return tokenType;
     }
 
+    if (isStart)
+    {
+        isStart = false;
+        if (!skipUtf8Bom())
+        {
+            tokenType = TokenError;
+            return tokenType;
+        }
+    }
+
     textToken.clear();
     tokenValue = std::nan("");
     State state = State::Start;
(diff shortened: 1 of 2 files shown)


ajtribick:
Build failures on Windows should be fixed by #1135 - can rebase this one on top of that after it is merged.

PR #1135, opened 2021-10-17
Fixes for the Windows build

ajtribick:
- Move the compatibility classes into the cel17 namespace
- Skip generating resource translations if Perl not present
- Use fmt/ostream.h instead of fmt/printf.h for string_view

### DIFF
--- a/src/celcompat/cc.cpp
+++ b/src/celcompat/cc.cpp
@@ -11,12 +11,14 @@
 
 #include "cc.h"
 
-namespace std
+namespace celestia
+{
+namespace compat
 {
 
 auto from_chars(const char* first, const char* last, float &value,
-                std::chars_format fmt)
-    -> std::from_chars_result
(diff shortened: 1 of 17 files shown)


ajtribick:
> > ```
> > * Move the compatibility classes into the cel17 namespace
> > ```
> 
> Could you show the error you're trying to fix?

The issue I was originally solving came down to a call to `fmt::fprintf` on the `string_view` shim from sv.h. Since fmt doesn't know about this type, this failed to compile. I've now defined a `fmt:formatter` specialization and replaced the relevant `fmt::fprintf` call (in url.cpp) with the `fmt::print` overload from `fmt/ostream.h`.

As regards moving the shims into the `cel17` namespace, adding types to namespace `std` (except for explicitly-permitted template specializations) causes the program to have undefined behaviour per the C++ standard. We should not do that, so I took the opportunity to fix it.

PR #1145, opened 2021-10-29
Fix buffer overflow error in cel3ds, improve error handling

ajtribick:
I noticed a potential buffer overflow error in the `readString` function in 3dsread.cpp: if the string is 1024 characters or longer, the terminating zero byte is not written and the string constructor from `const char*` will therefore read beyond the end of the buffer.

The rest of the loader functionality was in my opinion far too trusting that the input files are not malformed. I've therefore added various error checking that was missing, including verifying that the stream is in a good state after attempting to read data, and consistency checks that the expected amount of data has been read. I also replaced the void pointer usage with templates to ensure better type safety, and switched to using `std::make_unique` rather than `new` to simplify cleanup on the error path.

So far I've checked that the Apollo 11 and Mir models still load.

### DIFF
--- a/src/cel3ds/3dsread.cpp
+++ b/src/cel3ds/3dsread.cpp
@@ -9,6 +9,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <fstream>
 #include <iostream>
 #include <string>
@@ -17,7 +18,7 @@
 #include <Eigen/Core>
 #include <fmt/ostream.h>
 
-#include "celutil/bytes.h"
+#include <celutil/bytes.h>
 #include "3dschunk.h"
 #include "3dsmodel.h"
 #include "3dsread.h"
@@ -34,8 +35,9 @@ using ProcessChunkFunc = std::int32_t (*)(std::istream &, std::uint16_t, std::in
 
 bool readInt(std::istream& in, std::int32_t& value)
 {
-    in.read(reinterpret_cast<char*>(&value), sizeof(std::int32_t));
-    if (!in.good()) { return false; }
+    char buffer[sizeof(value)];
+    if (!in.read(buffer, sizeof(buffer)).good()) { return false; }
+    std::memcpy(&value, buffer, sizeof(value));
     LE_TO_CPU_INT32(value, value);
     return true;
 }
@@ -43,8 +45,9 @@ bool readInt(std::istream& in, std::int32_t& value)
 
 bool readShort(std::istream& in, std::int16_t& value)
 {
-    in.read(reinterpret_cast<char*>(&value), sizeof(std::int16_t));
-    if (!in.good()) { return false; }
+    char buffer[sizeof(value)];
+    if (!in.read(buffer, sizeof(buffer)).good()) { return false; }
+    std::memcpy(&value, buffer, sizeof(value));
     LE_TO_CPU_INT16(value, value);
     return true;
 }
@@ -52,8 +55,9 @@ bool readShort(std::istream& in, std::int16_t& value)
 
 bool readUshort(std::istream& in, std::uint16_t& value)
 {
-    in.read(reinterpret_cast<char*>(&value), sizeof(std::uint16_t));
-    if (!in.good()) { return false; }
+    char buffer[sizeof(value)];
+    if (!in.read(buffer, sizeof(buffer)).good()) { return false; }
+    std::memcpy(&value, buffer, sizeof(value));
     LE_TO_CPU_INT16(value, value);
     return true;
 }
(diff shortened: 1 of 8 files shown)


ajtribick:
The code analysis report for the comma operator in matrix initialization is a known issue: https://trac.cppcheck.net/ticket/10518

As for the "Check buffer boundaries if used in a loop including recursive loops" - I genuinely don't know what it wants me to do here. The call to read is using the size of the buffer defined immediately beforehand.

src/cel3ds/3dsread.cpp
```diff
@@ -7,52 +7,58 @@
...
 
-#include <iomanip>
+#include <cstdint>
+#include <fstream>
+#include <iostream>
+#include <string>
 
+#include <Eigen/Core>
+
+#include "celutil/bytes.h"
```
375gnu, line 21: Hi @ajtribick! We follow the following style: stl, 3rd party libs, cel*/*.h, local *.h. And we use `<>` with cel*.

src/cel3ds/3dsread.cpp
```diff
@@ -34,35 +35,39 @@ using ProcessChunkFunc = std::int32_t (*)(std::istream &, std::uint16_t, std::in
 
 bool readInt(std::istream& in, std::int32_t& value)
 {
-    in.read(reinterpret_cast<char*>(&value), sizeof(std::int32_t));
-    if (!in.good()) { return false; }
+    char buffer[sizeof(value)];
```
375gnu, line 38: Could you point to any official documentation regarding UB in read()? Because https://en.cppreference.com/w/cpp/io/basic_istream/read uses exactly the same code as the current Celestia.

375gnu (changes requested):

src/cel3ds/3dsread.cpp
ajtribick: It's not precisely the same, but this appears to be similar to the example at 43:44 in Timur Doumler's talk "[Type Punning in Modern C++](https://www.youtube.com/watch?v=_qzMpk-22cc)" - while the `reinterpret_cast` from a `std::int32_t*` to `char*` is allowed, according to the C++ abstract machine the resultant pointer does not actually point to a char array (and there isn't a char array at this location), therefore any access to it as though it was a char array is (as far as I understand it) undefined behaviour.

src/cel3ds/3dsread.cpp
ajtribick: Ok I can change these to angle brackets (I think the order in this file matches what you've stated).

Question - should the celcompat headers (which essentially do forward compatibility for C++17 stl) be grouped with the cel headers or the stl headers?

375gnu: we put them together with other cel headers

src/cel3ds/3dsread.cpp
375gnu:
```
    std::uint32_t n;
    if(raw.read(reinterpret_cast<char*>(&n), sizeof n))
        std::cout << std::hex << std::showbase << n << '\n';
```

is this not the same?

i need a few days to check the video, unfortunately i don't have time till Monday

ajtribick: As far as I'm aware, that example is undefined behaviour because the first parameter of `read` should be a pointer to a char array, and by the rules of the C++ abstract machine, there isn't actually a char array at whatever `reinterpret_cast<char*>(&n)` is pointing to.

375gnu: @ajtribick can you remove this commit from this PR so we can go forward?

375gnu: I've just found what i was looking for: https://en.cppreference.com/w/c/language/object, "Strict aliasing".
> Given an object with effective type T1, using an lvalue expression (typically, dereferencing a pointer) of a different type T2 is undefined behavior, unless:
> * T2 is a character type (char, signed char, or unsigned char).

So our old code is absolutely correct.

375gnu: I've just found what i was looking for: https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing.

> Whenever an attempt is made to read or modify the stored value of an object of type DynamicType through a glvalue of type AliasedType, the behavior is undefined unless one of the following is true: 
> AliasedType is std::byte (since C++17), char, or unsigned char: this permits examination of the object representation of any object as an array of bytes.

So our old code is absolutely correct.

ajtribick: I'll point out that cppreference is not the standard. I fully agree that `reinterpret_cast` rules regarding `char` and `unsigned char` are *intended* to let you access the byte representation of the array but due to rules brought in in C++17 this seems to have accidentally become undefined behaviour. The proposal to fix this is http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1839r2.pdf which includes a description of the problem.

From what I'm seeing in that proposal, I *think* we're ok with `reinterpret_cast` in C++14, but I think it would be good to stay forward-compatible with the C++17 standard where the only valid way of accessing the byte representation of an object is via `memcpy`. Based on what I'm seeing on Godbolt the optimisers in the major compilers are pretty good at removing the `memcpy` from the final code.

375gnu (approved):
sometimes c++ is too stupid. but let's roll

levinli303 (approved):

375gnu:
@ajtribick we usually merge own PRs ourselves so feel free to merge this if you consider it ready.

PR #1150, opened 2021-11-06
Set up common binary IO functions

ajtribick:
Extracts little/big-endian utiilty functions that were present across several cpp files and moves them into common header files. I've also added some more error handling where I updated the files.

The file binarywrite.h is currently not used, this will be used in a follow-up cleanup of modelfile.cpp, which I will raise as a separate pull request as the fixes required there are quite extensive.

### DIFF
--- a/src/celimage/bmp.cpp
+++ b/src/celimage/bmp.cpp
@@ -11,13 +11,14 @@
 #include <cstdint>
 #include <fstream>  // ifstream
 #include <iostream> // ios
+#include <memory>
+#include <vector>
+
 #include <celengine/image.h>
-#include <celutil/bytes.h>
+#include <celutil/binaryread.h>
 #include <celutil/debug.h>
 
-using std::ifstream;
-using std::ios;
-using celestia::PixelFormat;
+namespace celutil = celestia::util;
 
 namespace
 {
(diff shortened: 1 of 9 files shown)


src/celimage/bmp.cpp
```diff
@@ -26,120 +27,100 @@ namespace
...
-
-    if (imageHeader.width <= 0 || imageHeader.height <= 0)
-        return nullptr;
-
-    // We currently don't support compressed BMPs
-    if (imageHeader.compression != 0)
-        return nullptr;
-    // We don't handle 1-, 2-, or 4-bpp images
-    if (imageHeader.bpp != 8 && imageHeader.bpp != 24 && imageHeader.bpp != 32)
+    if (!celutil::read_le(in, fileHeader.magic[0])
```
375gnu, line 56: there are places where you use read_le with explicitly defined type and here without it. also according to the current style it should be named ReadLE

375gnu (changes requested):

src/celimage/bmp.cpp
ajtribick: Thanks, fixed.

375gnu (approved):

levinli303 (approved):

PR #1155, opened 2021-11-13
Update cmod I/O

### LINKED ISSUE #1151: Upper limit on token lengths?
Currently working on a few changes that involve the tokenizer used to parse the various catalog files, one thing I've noticed is it currently allows unlimited token lengths, which would eventually exhaust memory. In order to tighten things up, I am planning to add an upper limit: if the token length is exceeded, the token will be reported as an error and parsing will bail out.

As an initial suggestion, I'm considering putting the limit to 1024 characters (which basically correspond to UTF-8 code units), perhaps with a lower limit (64) for tokens that aren't quote-delimited strings.

Any use cases you can see for supporting longer tokens?

ajtribick:
* Replace the tokenizer implementation in the celmodel code with the one used for other file types, move the tokenizer to celutil
* Use the binaryread.hpp/binarywrite.hpp routines for parsing binary CMOD files
* Enforce maximum token length limit of 1024 characters - resolves #1151 
* Disallow creation of surrogate pair characters (U+D800 to U+DFFF) via Unicode escape sequences
* Detect and reject invalid UTF-8 sequences in the tokenizer

### DIFF
--- a/src/celmodel/modelfile.cpp
+++ b/src/celmodel/modelfile.cpp
@@ -8,35 +8,110 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
+#include <cassert>
+#include <cstring>
+#include <iostream>
+
 #include <fmt/core.h>
 #include <fmt/ostream.h>
 
-#include "modelfile.h"
-#include <celutil/bytes.h>
-#include <cassert>
-#include <cmath>
-#include <cstring>
-#include <vector>
 #include <celutil/binaryread.h>
 #include <celutil/binarywrite.h>
-#include <celutil/debug.h>
 #include <celutil/tokenizer.h>
+#include "mesh.h"
+#include "model.h"
+#include "modelfile.h"
 
-
-using namespace cmod;
-using namespace std;
 namespace celutil = celestia::util;
 
 
+namespace cmod
+{
+namespace
+{
+constexpr std::size_t CEL_MODEL_HEADER_LENGTH = 16;
+constexpr const char CEL_MODEL_HEADER_ASCII[] = "#celmodel__ascii";
+constexpr const char CEL_MODEL_HEADER_BINARY[] = "#celmodel_binary";
+
 // Material default values
-static Material::Color DefaultDiffuse(0.0f, 0.0f, 0.0f);
-static Material::Color DefaultSpecular(0.0f, 0.0f, 0.0f);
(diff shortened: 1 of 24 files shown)


src/celmodel/modelfile.cpp
```diff
@@ -2127,251 +1635,332 @@ BinaryModelLoader::loadVertices(const Mesh::VertexDescription& vertexDesc,
...
-                out.write(cdata, 4);
+                result = out.write(cdata, 4).good();
                 break;
             default:
                 assert(0);
                 break;
             }
+
+            if (!result)
+            {
```
375gnu, line 1812: could you use the same style everywhere?

375gnu (changes requested):

375gnu:
did you think about replacing material::color with celutil::color? for me it's annoying to have 2 similar implementations

src/celmodel/modelfile.cpp
ajtribick: What style are you referring to here?

src/celmodel/modelfile.cpp
```diff
@@ -8,535 +8,110 @@
...
-
-        } // Switch
-
-        if (!hasError() && !newToken.isValid())
-        {
-            m_nextChar = readChar();
-        }
-    }
+ public:
+    virtual ~ModelWriter() {};
```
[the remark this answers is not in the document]
ajtribick: Leftover from moving this class out of modelfile.h. Fixed

src/celutil/utf8.h
```diff
@@ -136,4 +136,26 @@ class Greek
...
+{
+public:
+    UTF8Validator() = default;
+    ~UTF8Validator() = default;
+
+    bool check(char c);
+    bool check(unsigned char c);
+
+private:
+    enum class State {
```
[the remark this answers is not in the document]
ajtribick: Done.

ajtribick:
Yes, that's definitely something I wanted to look at... I've been doing some further tidying-up of the model-related code which I'll raise in a future PR.

src/celmodel/modelfile.cpp
375gnu: there are many places where you do `if (...) { return ...; }`. Here you use multiple lines.

ajtribick: Got it. Fixed.

375gnu (approved):

PR #1159, opened 2021-11-15
Fix for fmt 4.0 (as used by Ubuntu 18.04 LTS)

ajtribick:
Ubuntu 18.04 LTS supplies fmt 4.0, which doesn't have the core.h header.


375gnu (approved):
i'll add ubuntu18.04 to our ci

PR #1160, opened 2021-11-15
Ensure reproducibility of "random" template generation

ajtribick:
- Add an implementation of [Bob Jenkins' small fast noncryptographic random number generator](http://burtleburtle.net/bob/rand/talksmall.html), adapted to meet the [UniformRandomBitGenerator](https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator) named requirement to allow it to be used with the distributions in the `<random>` header.
- Use separate generators in galaxy, globular and Perlin noise generation to ensure these generations are independent.

The seed values are taken from the output of a random device. According to [this article by the creator of the PCG pseudorandom number generator](https://www.pcg-random.org/posts/bob-jenkins-small-prng-passes-practrand.html), this generator passes PractRand, and it's a lot smaller than the Mersenne Twister (which does not).

This ensures that the galaxy and globular cluster sprite distributions, and the shape of cms meshes is consistent between runs and across distributions.

### DIFF
--- a/src/celengine/galaxy.cpp
+++ b/src/celengine/galaxy.cpp
@@ -602,7 +602,7 @@ GalacticForm* buildGalacticForms(const fs::path& filename)
     height = img->getHeight();
     rgb    = img->getComponents();
 
-    auto& rng = getRNG();
+    celmath::Jsf32 rng(UINT32_C(0xa0030c29));
     for (int i = 0; i < width * height; i++)
     {
         value = img->getPixels()[rgb * i];
@@ -667,7 +667,7 @@ GalacticForm* buildGalacticForms(const fs::path& filename)
     // reshuffle the galaxy points randomly...except the first kmin+1 in the center!
     // the higher that number the stronger the central "glow"
 
-    shuffle(galacticPoints->begin() + kmin, galacticPoints->end(), getRNG());
+    shuffle(galacticPoints->begin() + kmin, galacticPoints->end(), celmath::Jsf32(UINT32_C(0x29bf0617)));
 
     auto* galacticForm  = new GalacticForm();
     galacticForm->blobs = galacticPoints;
@@ -726,7 +726,7 @@ void InitializeForms()
     BlobVector* irregularPoints = new BlobVector;
     irregularPoints->reserve(galaxySize);
 
-    auto& rng = getRNG();
+    celmath::Jsf32 rng(UINT32_C(0xa24af6d3));
     while (ip < galaxySize)
     {
         p = Vector3f(RealDists<float>::SignedUnit(rng),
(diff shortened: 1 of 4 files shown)


src/celengine/galaxy.cpp
```diff
@@ -602,7 +602,7 @@ GalacticForm* buildGalacticForms(const fs::path& filename)
     height = img->getHeight();
     rgb    = img->getComponents();
 
-    auto& rng = getRNG();
+    celmath::Jsf32 rng(UINT32_C(0xa0030c29));
```
375gnu, line 605: magic constants, we all like them! :)))

could you provide names for them with some descriptive comments?

src/celmath/randutils.cpp
```diff
@@ -87,8 +75,36 @@ constexpr inline float smooth(float t)
...
+constexpr inline std::uint32_t rot(std::uint32_t x, int k)
+{
+    return (x << k) | (x >> (32 - k));
+}
+
 } // end of anonymous namespace
 
+
+Jsf32::Jsf32(std::uint32_t seed) :
+    a(UINT32_C(0xf1ea5eed)), b(seed), c(seed), d(seed)
```
375gnu, line 87: please make one variable per line

src/celmath/randutils.cpp
```diff
@@ -87,8 +75,36 @@ constexpr inline float smooth(float t)
...
+{
+    for (int i = 0; i < 20; ++i)
+    {
+        (void)this->operator()();
```
375gnu, line 91: I think it's better to add a private function and call it here and in operator, cause this like looks cryptic

src/celmath/randutils.h
```diff
@@ -9,6 +10,27 @@
...
+// Small noncryptographic pseudorandom number generator by Bob Jenkins
+// http://burtleburtle.net/bob/rand/talksmall.html
+class Jsf32
+{
+public:
+    using result_type = std::uint32_t;
+    Jsf32(std::uint32_t seed);
+
+    static constexpr result_type min() { return 0; }
+    static constexpr result_type max() { return UINT32_MAX; }
```
375gnu, line 23: std::numeric_limits<result_type>::max()

375gnu: std::numeric_limits<result_type>::min()

375gnu (changes requested):

ajtribick:
Withdrawing this as it's incomplete: while it guarantees the underlying RNG is consistent between platforms and runs, the implementations of the distributions (`std::uniform_int_distribution` and `std::uniform_real_distribution`) can and do vary between implementations, so this only gets us partially there.

PR #1169, opened 2021-11-16
Tokenizer improvements, recover from invalid UTF-8 sequences

ajtribick:
* Disallow numbers ending with a sign (+, -, 3.2e+, 4.7e-)
* Disallow exponents immediately after the sign character (+e12, -e6)
* Disallow invalid string escapes (\q)
* More forgiving UTF-8 parsing: write an error to the log (once per stream). In string literals, remove any partially-written sequences and append U+FFFD REPLACEMENT CHARACTER. Then continue processing the next byte of the input string.

### DIFF
--- a/src/celutil/utf8.h
+++ b/src/celutil/utf8.h
@@ -66,6 +66,24 @@ inline int UTF8EncodedSize(wchar_t ch)
 #endif
 }
 
+constexpr inline int UTF8EncodedSizeChecked(std::uint32_t ch)
+{
+    if (ch < 0x80)
+        return 1;
+    if (ch < 0x800)
+        return 2;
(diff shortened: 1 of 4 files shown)


src/celutil/utf8.h
```diff
@@ -66,6 +66,24 @@ inline int UTF8EncodedSize(wchar_t ch)
 #endif
 }
 
+constexpr inline int UTF8EncodedSizeChecked(std::uint32_t ch)
```
ajtribick, line 69: For now I don't want to change the existing `UTF8EncodedSize` function which supports up to 6-byte "UTF-8" sequences and is used at various places around the codebase, but I intend to fix that in a later PR.

375gnu (approved):

PR #1170, opened 2021-11-16
Don't read DDS header twice, reallow triggering eof()

### LINKED ISSUE #1167: DDS textures are broken since da5f621
Commit  da5f621 introduced common routines to read binary data. But it broke DDS textures:

image

This supposed to be Earth. The texture used is from the latest Celestia Origin.

ajtribick:
Fixes #1167

I'm not particularly happy about allowing the `eof()` condition at the end as this implies we are attempting to read more data than the file contains, but it seems to be necessary at least with the Celestia Origin .dds files. I will take another look at this in future, but in the meantime it addresses the issue for people using master builds.

### DIFF
--- a/src/celimage/dds.cpp
+++ b/src/celimage/dds.cpp
@@ -144,7 +144,6 @@ Image* LoadDDSImage(const fs::path& filename)
     }
 
     char header[4];
-    in.read(header, sizeof header);
     if (!in.read(header, sizeof(header)).good()
         || header[0] != 'D' || header[1] != 'D'
         || header[2] != 'S' || header[3] != ' ')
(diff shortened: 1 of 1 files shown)


src/celimage/dds.cpp
```diff
@@ -144,7 +144,6 @@ Image* LoadDDSImage(const fs::path& filename)
     }
 
     char header[4];
-    in.read(header, sizeof header);
```
375gnu, line 147: we should be more careful when reviewing changes

375gnu (approved):

src/celimage/dds.cpp
ajtribick: Agreed. Apologies for the breakages along the way, but I do think it's important for an application that's dealing with user content to be a lot stricter about error checking the input files these days.

PR #1171, opened 2021-11-17
Height of atmosphere/cloud should be added to culling radius

levinli303:


src/celengine/body.cpp
```diff
@@ -1178,10 +1178,7 @@ void Body::recomputeCullingRadius()
         r = max(r, rings->outerRadius);
 
     if (atmosphere)
-    {
-        r = max(r, atmosphere->height);
-        r = max(r, atmosphere->cloudHeight);
-    }
+        r += max(atmosphere->height, atmosphere->cloudHeight);
```
375gnu, line 1181: this looks reasonable (if atmosphere stores height above surface) but then you should move this to line 1177 (before `if (rings)` otherwise you add atmosphere height to outer rings radius.

could you also provide actual values of getBoundingRadius(), atmosphere->height and atmosphere->cloudHeight for some bodies?

375gnu (changes requested):

src/celengine/body.cpp
levinli303: for the earth, getBoundingRadius() is 6310.xxxx, atmosphere->height is 100, cloudHeight is 7.

375gnu: okay

levinli303: moved to before the ring check

375gnu (approved):

PR #1175, opened 2021-11-18
Add default go to time/center time to cfg

levinli303:


src/celengine/observer.cpp
```diff
@@ -669,7 +669,7 @@ void Observer::computeCenterParameters(const Selection& destination,
     jparams.startInterpolation = 0;
     jparams.endInterpolation   = 1;
 
-    jparams.accelTime = 0.5;
+    jparams.accelTime = min(1.0, max(0.1, defaultAccelTime));
```
375gnu, line 672: celmath::clamp(defaultAccelTime, 0.1, 1.0)

src/celengine/observer.cpp
```diff
@@ -632,7 +632,7 @@ void Observer::computeGotoParametersGC(const Selection& destination,
     jparams.startInterpolation = min(startInter, endInter);
     jparams.endInterpolation   = max(startInter, endInter);
 
-    jparams.accelTime = 0.5;
+    jparams.accelTime = min(1.0, max(0.1, defaultAccelTime));
```
375gnu, line 635: celmath::clamp(defaultAccelTime, 0.1, 1.0)

src/celengine/observer.h
```diff
@@ -305,6 +314,13 @@ class Observer
     Eigen::Vector3d 	targetVelocity{ 0.0, 0.0, 0.0 };
     Eigen::Vector3d 	initialVelocity{ 0.0, 0.0, 0.0 };
     double          	beginAccelTime{ 0.0 };
+    double              defaultGoToTime{ 5.0 };
```
375gnu, line 317: misindent

src/celengine/observer.cpp
```diff
@@ -570,7 +570,7 @@ void Observer::computeGotoParameters(const Selection& destination,
...
```
375gnu, line 573: celmath::clamp(defaultAccelTime, 0.1, 1.0)

src/celengine/observer.cpp
```diff
@@ -1197,17 +1249,17 @@ void Observer::gotoLocation(const UniversalCoord& toPosition,
...
     journey.to = toPosition;
     journey.finalOrientation = toOrientation;
 
-    journey.startInterpolation = 0.25f;
-    journey.endInterpolation   = 0.75f;
+    journey.startInterpolation = defaultInterpolationStart;
+    journey.endInterpolation   = defaultInterpolationEnd;
 
-    journey.accelTime = 0.5;
+    journey.accelTime = min(1.0, max(0.1, defaultAccelTime));
```
375gnu, line 1262: celmath::clamp(defaultAccelTime, 0.1, 1.0)

src/celestia/celestiacore.cpp
```diff
@@ -3966,6 +3966,32 @@ bool CelestiaCore::initSimulation(const fs::path& configFileName,
...
+    {
+        sim->setDefaultInterpolation(config->defaultInterpolationStart, config->defaultInterpolationEnd);
+    }
+
+    if (config->altInterpolationStart >= 0.0 && config->altInterpolationEnd <= 1.0 && config->altInterpolationEnd >= config->altInterpolationStart)
+    {
+        sim->setAlternativeInterpolation(config->altInterpolationStart, config->altInterpolationEnd);
+    }
+
+    if (config->defaultAccelTime >= 0.1 && config->defaultAccelTime <= 1.0)
```
375gnu, line 3990: i think it's good to provide constants like k_defaultAccelTimeMin = 0.1 & k_defaultAccelTimeMax = 1.0 instead of hardcoding 0.1 & 1.0 here and there

375gnu (changes requested):
i'll check more carefully later, these are first notices

src/celengine/observer.h
levinli303: hmm is there? they look aligned to me.

src/celscript/lua/celx_observer.cpp
```diff
@@ -284,21 +287,21 @@ static int observer_gototable(lua_State* l)
...
     lua_pushstring(l, "accelTime");
     lua_gettable(l, 2);
-    jparams.accelTime = celx.safeGetNumber(3, NoErrors, "", 0.5);
+    jparams.accelTime = celx.safeGetNumber(3, NoErrors, "", o->getDefaultAccelerationTime());
     lua_settop(l, 2);
 
     jparams.duration = max(0.0, jparams.duration);
-    jparams.accelTime = min(1.0, max(0.1, jparams.accelTime));
+    jparams.accelTime = celmath::clamp(jparams.accelTime, Observer::defaultAccelTimeMin, Observer::defaultAccelTimeMax);
     jparams.startInterpolation = min(1.0, max(0.0, jparams.startInterpolation));
```
375gnu, line 305: could you update to clamp here and on the next line as well?

src/celengine/observer.h
375gnu: probably you have a TAB instead of spaces

src/celscript/lua/celx_observer.cpp
levinli303: done

src/celengine/observer.h
levinli303: ok apparently the other lines used some tabs instead. I've replaced all the tabs in this file with spaces

375gnu:
after some additional thought i suppose that instead of hacking here we can provide a lua hook instead

375gnu:
```
myluahook = {
    charentered = function(self, ch)
        if ch == 'g' or ch == 'G' then
            selection = celestia:getselection()
            observer  = celestia:getobserver()
            if selection and observer then
                observer:gotoobject(selection, 20, 0.1, 0.9)
            end
            return true
        end
        return false
    end;
}
```

375gnu:
only accel is not suported

levinli303:
> after some additional thought i suppose that instead of hacking here we can provide a lua hook instead

While it's true that lua script can be used for the purpose.this PR can also act as an effort to remove the default values littering around everywhere.

375gnu:
> While it's true that lua script can be used for the purpose.this PR can also act as an effort to remove the default values littering around everywhere.

That's a good idea not to have magic value here and there. But the current PR looks rather like Q&D fix.
`jparams.accelTime = celmath::clamp(defaultAccelTime, defaultAccelTimeMin, defaultAccelTimeMax);` - we should not validate values every time, we should validate them only once.

`jparams.duration = centerTime < 0.0 ? defaultCenterTime : centerTime;` - instead of introducing another magic value like here `void centerSelection(double centerTime = -1.0);` we can provide 2 overloads:
```
void centerSelection(double centerTime);
void centerSelection() { centerSelection(defaultCenterTime); };
```
Another bad solution is to keep copies of the same values both in `Simulation` and `Observer`.

But the main my point is that we have scripting support to override default Celestia behaviour. If something is missing, then it's better to update celx code. We should have small maintainable application core and move additional features used by minority of users to scripts. When such features proves their usefulness for all or vast majority of users then we can rewrite them in C++. This also means that features unused by mobile versions can be moved to separate files and built only with desktop targets.

@ajtribick what's you opinion on this?

ajtribick:
I definitely agree that the parameter validation should be done once during the input handling, then the adjusted/corrected parameters passed to the core (or just error out if the values are completely out of range). I do like the idea of using scripting more for the various core actions though: reduce the number of explicitly mapped keybindings in the C++ and pass unmapped ones to a hypothetical keyhandler.celx, which would be the more flexible solution that could support other use cases, e.g. people wanting to unbind the demo key.

alcomposer:
Scripting speed value is fine for me. But hardcoding feels like a bad idea. Key bindings are not needed for Celestia to function. A user can still explore with a mouse etc? At which point there is still a 'default speed' hardcoded into Celestia.

So is adding a speed field to `.cfg` dead in the water? Or could both scripting & non-hardcoded default be acceptable?

375gnu:
> Or could both scripting & non-hardcoded default be acceptable if properly implemented?

The latter.

Currently you can do the following:

1. Create a file `hook.lua` with the following content:
```
GOTO_TIME   = 20.0
ITER_START  = 0.1
ITER_FINISH = 0.9

function gotoobject(self)
    selection = celestia:getselection()
    observer  = celestia:getobserver()
    if selection and observer then
        observer:gotoobject(selection, GOTO_TIME, ITER_START, ITER_FINISH)
        return true
    end
    return false
end

keys = {
    g = gotoobject,
}

hook = {
    charentered = function(self, ch)
        fun = keys[ch]
        if fun then
            return fun(self)
        end
        return false
    end;
}

celestia:setluahook(hook)
```

2. Edit your `celestia.cfg` (I suggest editing `.celestia.cfg` in your home directory). Add the following line to it
```
LuaHook "hook.lua"
```
If you put the file not to your celestia directory then use the full path to it.

PR #1176, opened 2021-11-18
Add unit tests for the tokenizer

ajtribick:
- Added tests for the tokenizer
- Reorganized the tests so that the catch main method is only compiled once
- Fixed missing error return in the tokenizer identified by the tests


test/CMakeLists.txt
```diff
@@ -1 +1,2 @@
+add_subdirectory(common)
```
ajtribick, line 1: I moved the catch.hpp and the main file into a the common directory as I'm planning to add some integration-type tests later.

375gnu: yeah, it takes too long to build catch's main

375gnu (approved):

ajtribick:
@375gnu - since you approved, I just realized the last couple of test cases were missing the `[Tokenizer]` tag, have added those.

PR #1178, opened 2021-11-18
Add test for CMOD binary→ASCII→binary roundtrip

ajtribick:
An integration test roundtripping the iss.cmod file to ASCII and back. This *should* now work with the tokenizer updates, unless the standard library floating point parser is broken in some implementations.


375gnu (approved):

ajtribick:
My suspicion is that the Ubuntu 18.04 build failure is related to the old version of fmt supplied by the distribution: checking the version history of fmt, it looks like 6.1 or maybe 6.0 would be needed to have reliable floating point output.

Not sure what's going on with the Windows builds, locally on a 64-bit build this passes, possibly I've messed up the file paths - will need to instrument this test better to figure out precisely where it's failing. Withdrawing this for now.

ajtribick:
As discussed in #1179 - fmt versions before 6.1.0 don't produce roundtrip-capable output for floats in the default format. I have therefore added fmt 6.1.0 as a submodule in thirdparty, and an additional option in the root CMakeLists.txt that enables forcing the build to use this even if a version ≥6.1.0 is found via `find_package`.

Also updated the Windows CI to check out the submodules, otherwise copying the iss.cmod file used in the integration test will fail.

CMakeLists.txt
```diff
@@ -210,7 +211,18 @@ find_package(Eigen3 3.3 REQUIRED NO_MODULE) # -DEigen3_DIR=...
 message(STATUS "Found Eigen3 ${EIGEN3_VERSION_STRING}")
 link_libraries(Eigen3::Eigen)
 
-find_package(fmt REQUIRED)
```
375gnu, line 213: I don't think that custom option is required and so the following is enough (a bit changed old code). 
```
find_package(fmt 6.1.0 CONFIG QUIET)
if(NOT fmt_FOUND)
  message(STATUS "Using fmt submodule")
  add_subdirectory("${CMAKE_SOURCE_DIR}/thirdparty/fmt")
endif()
link_libraries(fmt::fmt)
```

munix9:
FYI: For the build on OBS I used the sources of libfmt 6.2.1 from the Kodi repo for Ubuntu_18.04 (bionic-backport):
https://launchpad.net/~team-xbmc/+archive/ubuntu/ppa/+packages
`fmtlib_6.1.2+ds-2ubuntu1~bionic.debian.tar.xz`

Because I don't want to check out the whole content directory as a submodule with currently over 270 MB size.

Looks like it will work.
There should be no dependency issues as libfmt 6.2.1 is statically included.

https://build.opensuse.org/project/monitor/home:munix9:unstable

ajtribick:
@munix9 - do your builds run the tests? At the moment, this change pulls iss.cmod out of the content submodule in order to use it for the roundtrip testing. I could add it specifically to the test directory instead, which would avoid having to pull in the submodule.

munix9:
Oh yes, please provide the file in the test directory - it doesn't make much sense to get the whole submodule just because of one file.

In the openSUSE builds the tests are done, in the Debian and Ubuntu builds (currently) not.

CMakeLists.txt
ajtribick: Done.

ajtribick:
Ok, added iss.cmod under test/data/iss/models.

375gnu:
> I just noticed that Debian 10 ships libfmt-dev-5.2.1+ds-2.
> Then I need a workaround here as well.

Backports should provide 6.2.

munix9:
> > I just noticed that Debian 10 ships libfmt-dev-5.2.1+ds-2.
> > Then I need a workaround here as well.
> 
> Backports should provide 6.2.

Yes, I have already taken that into account.

https://build.opensuse.org/package/show/home:munix9:unstable/libfmt-debian10
from
https://packages.debian.org/buster-backports/libfmt-dev

ajtribick:
@munix9 - so is it ok from your side now? If so, I will do the merge.

munix9:
> @munix9 - so is it ok from your side now? If so, I will do the merge.

Yes, go for it.

PR #1182, opened 2021-11-23
Remove inheritance from Material maps, Model destructor

ajtribick:
- Reference texture maps by ResourceHandles in cmod Materials
- Add basic path manager for integration tests and tools
- Remove virtual destructor on Model class as nothing inherits from it

As far as I can see, the only reason to use `TextureResource` rather than `ResourceHandle` was so that the tools could store paths directly in the maps. This then ends up making the main part of Celestia's material handling more complex, including some awkward "juggling" logic in renderglsl.cpp to avoid textures being destroyed too early.

Redefining maps as being a `std::vector<ResourceHandle>` rather than a `std::vector<TextureResource*>` ends up slightly simplifying the Celestia code (and reducing some allocations) at the cost of adding a bit more complexity into the tools.

I've also made the `cmod::Model` destructor non-virtual, as nothing actually inherits from it (confirmed by doing a compilation with `final`).


src/celengine/modelgeometry.cpp
```diff
@@ -228,6 +228,7 @@ ModelGeometry::loadTextures()
 }
 
 
+/*
```
375gnu, line 231: please remove this instead of commenting out

src/celmodel/model.cpp
```diff
@@ -8,10 +8,12 @@
...
-#include "model.h"
+#include <algorithm>
 #include <cassert>
 #include <functional>
-#include <algorithm>
+
+#include <celutil/reshandle.h>
+#include "model.h"
 
 using namespace cmod;
```
375gnu, line 18: as you touched this file, then please replace this with
```
namespace cmod
{
```

375gnu (changes requested):

src/celengine/modelgeometry.cpp
ajtribick: Oops, had that on my to-do list but missed it. Done.

src/celmodel/model.cpp
ajtribick: Namespace/includes cleanup done.

src/celmodel/model.cpp
```diff
@@ -9,15 +9,125 @@
...
-using namespace std;
+
+namespace cmod
+{
+namespace
+{
+bool
+operator<(const Material::Color& c0, const Material::Color& c1)
+{
+    if (c0.red() < c1.red())
```
375gnu, line 25: can be replaced with std::tie

src/celmodel/model.cpp
```diff
@@ -9,15 +9,125 @@
...
+// materials.
+bool
+operator<(const Material& m0, const Material& m1)
+{
+    // Checking opacity first and doing it backwards is deliberate. It means
+    // that after sorting, translucent materials will end up with higher
+    // material indices than opaque ones. Ultimately, after sorting
+    // mesh primitive groups by material, translucent groups will end up
+    // rendered after opaque ones.
+    if (m0.opacity < m1.opacity)
```
375gnu, line 49: another candidate for std::tie

src/celmodel/model.cpp
ajtribick: `std::tie` doesn't work here because the results of the functions are temporaries. `std::forward_as_tuple` compiles but I'm not completely sure whether this ends up producing dangling references, so I've refactored this as a friend operator bool in the material files.

375gnu (approved):

PR #1187, opened 2021-11-25
Replace C++-style comments with C-style comments in gtkegl.c

ajtribick:
Apparently this has started to cause the Ubuntu 18.04 build to fail.

### DIFF
--- a/src/celestia/gtk/gtkegl.c
+++ b/src/celestia/gtk/gtkegl.c
@@ -145,7 +145,7 @@ gtk_widget_set_egl_capability(GtkWidget *widget)
         return TRUE;
 
 #if !GTK_CHECK_VERSION(3, 10, 0)
-    // No double buffering as we might have triple buffering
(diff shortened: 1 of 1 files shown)


375gnu:
This is strange. `set(CMAKE_C_STANDARD 11)` is ignored by cmake. Let me downgrade cmake and check locally.

ajtribick:
CMake allows decay to previous standards if [C_STANDARD_REQUIRED](https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD_REQUIRED.html#prop_tgt:C_STANDARD_REQUIRED) is not set, so Ubuntu 18.04 apparently decays to C90. Other than these comments (which is a C99 feature), I'm not sure we actually need to set the C standard later than C90 though, the `set(CMAKE_C_STANDARD 11)` seems like overkill to me.

375gnu (approved):
While I'm unable to reproduce the issue, there are no reasons to use smth higher that C89/90.

PR #1189, opened 2021-11-28
Implement Logger to replace raw cerr/cout and DPRINTF usage

375gnu:
ready for review


src/celutil/logger.h
```diff
@@ -0,0 +1,124 @@
...
+#endif
+
+namespace celestia
+{
+namespace util
+{
+class Logger
+{
+ public:
+    enum Level
```
ajtribick, line 20: I think it would be better to have this as an enum class at namespace level.

ajtribick:
Nice, I definitely like the idea of having a single unambiguous way to do logging in the codebase.

375gnu:
most probably not all levels are correct, some into should be at least verbose, some warnings should be errors and vice verse.

src/celutil/logger.h
```diff
@@ -0,0 +1,123 @@
...
+}
+
+template <typename... Args> inline void
+Logger::error(const char *format, const Args&... args) const
+{
+    log(Level::Error, format, args...);
+}
+
+template <typename... Args> void
+Logger::log2(Level level, const char *format, const Args&... args) const
```
ajtribick, line 113: I'm wondering if we should take the opportunity to update the printf-style strings to the new fmt syntax (with apologies to the translators), to reduce the API surface.

src/celutil/logger.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+#ifdef _MSC_VER
+    if (level == Level::Debug && IsDebuggerPresent())
+    {
+        OutputDebugStringA(fmt::vformat(format, args).c_str());
+        return;
+    }
+#endif
+
+    auto &stream = (level <= Level::Warning || level == Level::Debug) ? std::cerr : std::clog;
```
ajtribick, line 43: It might be worth having the output streams as member variables in the logger (defaulting to clog and cerr), which would make this testable. Then we could move the code in the `#if 0` block at the end of this file into the test project.

src/celutil/logger.h
375gnu: i will do this later, i want to writer a script to update translations first

375gnu: or i can start working on the script now and defer this pr

src/celutil/logger.cpp
375gnu: yeah, good idea

src/celutil/logger.h
ajtribick: Ok no worries, would rather get this one merged first I think.

375gnu: ok, i decided to got with the proposal and updated some languages

375gnu:
@levinli303 what's your oppinion?

levinli303 (approved):

PR #1190, opened 2021-11-28
Refactoring the CMOD code

ajtribick:
Various updates to the code, mainly around clarifying resource ownership and removing explicit new/delete in favour of RAII. The vertex data is now explicitly stored as 4-byte words, which should mean it meets alignment requirements.


src/cel3ds/3dsmodel.cpp
```diff
@@ -21,230 +21,274 @@ M3DColor::M3DColor(float _red, float _green, float _blue) :
...
+M3DTriangleMesh::getSmoothingGroupCount() const
 {
     return static_cast<std::uint16_t>(smoothingGroups.size());
 }
 
-void M3DTriangleMesh::addMeshMaterialGroup(std::unique_ptr<M3DMeshMaterialGroup> matGroup)
+void
+M3DTriangleMesh::addMeshMaterialGroup(M3DMeshMaterialGroup&& matGroup)
 {
     meshMaterialGroups.push_back(std::move(matGroup));
```
375gnu, line 201: could you update this with `emplace_back(matGroup);`?

src/cel3ds/3dsmodel.cpp
```diff
@@ -21,230 +21,274 @@ M3DColor::M3DColor(float _red, float _green, float _blue) :
...
+M3DModel::getTriMeshCount() const
 {
     return triMeshes.size();
 }
 
-void M3DModel::addTriMesh(std::unique_ptr<M3DTriangleMesh> triMesh)
+void
+M3DModel::addTriMesh(M3DTriangleMesh&& triMesh)
 {
     triMeshes.push_back(std::move(triMesh));
```
375gnu, line 232: the same as above (emplace_back)

src/cel3ds/3dsmodel.cpp
```diff
@@ -21,230 +21,274 @@ M3DColor::M3DColor(float _red, float _green, float _blue) :
...
+M3DScene::getModelCount() const
 {
     return models.size();
 }
 
-void M3DScene::addModel(std::unique_ptr<M3DModel> model)
+void
+M3DScene::addModel(M3DModel&& model)
 {
     models.push_back(std::move(model));
```
375gnu, line 263: emplace_back

src/celengine/rendcontext.h
```diff
@@ -8,29 +8,37 @@
...
+
+class Atmosphere;
+class Color;
+class LightingState;
 class Renderer;
 
 class RenderContext
 {
- public:
+public:
```
375gnu, line 30: indent

src/celengine/render.h
```diff
@@ -8,19 +8,20 @@
...
-#include <celengine/selection.h>
-#include <celengine/starcolors.h>
-#include <celengine/rendcontext.h>
-#include <celengine/renderlistentry.h>
+
+#include "universe.h"
+#include "selection.h"
+#include "starcolors.h"
+#include "rendcontext.h"
+#include "renderlistentry.h"
```
375gnu, line 24: i think it's better to restore the original includes because i have an idea to separate render code from other engine parts, it will include celengine/foo.h and the change  will be smaller

src/celengine/renderglsl.cpp
```diff
@@ -205,7 +261,7 @@ void renderEllipsoid_GLSL(const RenderInfo& ri,
             nTextures++;
 
 #ifdef GL_ES
-            if (gl::OES_texture_border_clamp)
+            if (celestia::gl::OES_texture_border_clamp)
```
375gnu, line 264: please restore original

src/celengine/renderglsl.cpp
```diff
@@ -10,59 +10,114 @@
...
-#include <celmath/intersect.h>
-#include <celutil/utf8.h>
-#include <celutil/util.h>
-#include "shadowmap.h"
-
-using namespace cmod;
-using namespace Eigen;
-using namespace std;
-using namespace celmath;
-using namespace celestia;
```
375gnu, line 40: please restore `using namespace celestia;` as all code will in celestia namespace so the future change will be minimal

src/celengine/renderglsl.cpp
```diff
@@ -643,7 +700,7 @@ renderAtmosphere_GLSL(const RenderInfo& ri,
         prog->setEclipseShadowParameters(ls, radius, planetOrientation);
 #endif
 
-    prog->setMVPMatrices(*m.projection, (*m.modelview) * vecgl::scale(atmScale));
+    prog->setMVPMatrices(*m.projection, (*m.modelview) * celestia::vecgl::scale(atmScale));
```
375gnu, line 698: please restore original

src/celengine/spheremesh.cpp
```diff
@@ -251,9 +250,9 @@ void SphereMesh::fixNormals()
     {
         float* v0 = normals + (i * (nSlices + 1)) * 3;
         float* v1 = normals + ((i + 1) * (nSlices + 1) - 1) * 3;
-        Vector3f n0(v0[0], v0[1], v0[2]);
-        Vector3f n1(v0[0], v0[1], v0[2]);
-        Vector3f normal = n0 + n1;
+        Eigen::Map<Eigen::Vector3f> n0(v0);
+        Eigen::Map<Eigen::Vector3f> n1(v1);
```
375gnu, line 257: yep, looks like original code had a mistake

src/celmodel/mesh.cpp
```diff
@@ -29,13 +29,14 @@ static size_t VertexAttributeFormatSizes[Mesh::FormatMax] =
      4,  // UByte4
 };
 
+} // end unnamed namespace
 
-Mesh::VertexDescription::VertexDescription(unsigned int _stride,
+VertexDescription::VertexDescription(unsigned int _stride,
                                            unsigned int _nAttributes,
```
375gnu: broken indentation

src/celmodel/mesh.cpp
```diff
@@ -88,18 +89,17 @@ Mesh::VertexDescription::operator=(const Mesh::VertexDescription& desc)
...
         VertexAttribute& attr = attributes[i];
 
         // Validate the attribute
-        if (attr.semantic >= SemanticMax || attr.format >= FormatMax)
-            return false;
-        if (attr.offset % 4 != 0)
-            return false;
-        if (attr.offset + VertexAttributeFormatSizes[attr.format] > stride)
+        if (attr.semantic >= VertexAttributeSemantic::SemanticMax
+            || attr.format >= VertexAttributeFormat::FormatMax
```
375gnu, line 92: with `enum class` these 2 lines are always false unless somebody wants to play with casts (but if somebody wants to show their feet who we are to forbid them?)

src/celmodel/mesh.cpp
```diff
@@ -8,19 +8,19 @@
...
-using namespace Eigen;
-using namespace std;
 
+namespace cmod
+{
+namespace
+{
 
-static size_t VertexAttributeFormatSizes[Mesh::FormatMax] =
+constexpr std::size_t VertexAttributeFormatSizes[static_cast<std::size_t>(VertexAttributeFormat::FormatMax)] =
```
375gnu, line 26: if c++14 allows `constexpr` functions with `switch` inside then it's better to provide a function instead of array with ugly static_casts

src/celmodel/mesh.h
```diff
@@ -8,144 +8,151 @@
...
+    VertexAttributeFormat   format;
+    unsigned int            offset;
+};
+
 
-    struct VertexDescription
+struct VertexDescription
+{
+    VertexDescription(unsigned int _stride,
+                        unsigned int _nAttributes,
```
375gnu: indentation issue

src/celmodel/mesh.h
```diff
@@ -8,144 +8,151 @@
...
-        TriStrip   = 1,
-        TriFan     = 2,
-        LineList   = 3,
-        LineStrip  = 4,
-        PointList  = 5,
-        SpriteList = 6,
-        PrimitiveTypeMax = 7,
-        InvalidPrimitiveGroupType = -1
-    };
+class PrimitiveGroup
```
375gnu, line 142: struct

src/celmodel/mesh.h
```diff
@@ -8,144 +8,151 @@
...
+{
+public:
+    PrimitiveGroup() = default;
```
375gnu, line 145: these 2 are uneeded

src/celmodel/mesh.cpp
```diff
@@ -29,60 +32,48 @@ constexpr std::size_t VertexAttributeFormatSizes[static_cast<std::size_t>(Vertex
...
-                                           VertexAttribute* _attributes) :
-    stride(_stride),
-    nAttributes(_nAttributes),
-    attributes(nullptr)
+template<typename It>
+VertexDescription appendingAttributes(const VertexDescription& desc, It begin, It end)
 {
-    if (nAttributes != 0)
+    std::vector<VertexAttribute> allAttributes;
+    allAttributes.reserve(desc.attributes.size() + (end - begin));
```
375gnu, line 40: is (end-begin) guaranteed to return size_t?

src/celmodel/mesh.h
```diff
@@ -96,31 +96,37 @@ struct VertexAttribute
...
         return semanticMap[static_cast<std::size_t>(semantic)];
     }
 
-    VertexDescription appendingAttributes(const VertexAttribute* newAttributes, int count) const;
-
     bool validate() const;
 
-    VertexDescription& operator=(const VertexDescription&);
-
     unsigned int stride;
```
375gnu: { 0 }

src/celmodel/mesh.h
```diff
@@ -96,31 +96,37 @@ struct VertexAttribute
...
 };
 
+bool operator==(const VertexAttribute& a, const VertexAttribute& b);
+bool operator<(const VertexAttribute& a, const VertexAttribute& b);
+
 
 struct VertexDescription
 {
+    VertexDescription() :
+        stride(0),
```
375gnu: plz use c++11 syntax

src/celmodel/modelfile.cpp
```diff
@@ -342,19 +343,20 @@ AsciiModelLoader::loadMaterial()
...
 
-VertexDescription*
+VertexDescription
 AsciiModelLoader::loadVertexDescription()
 {
     if (tok.nextToken() != Tokenizer::TokenName || tok.getNameValue() != VertexDescToken)
     {
         reportError("Vertex description expected");
-        return nullptr;
+        return VertexDescription();
```
375gnu, line 347: return {}

src/celmodel/modelfile.cpp
```diff
@@ -367,8 +369,7 @@ AsciiModelLoader::loadVertexDescription()
             // TODO: Should eliminate the attribute limit, though no real vertex
             // will ever exceed it.
             reportError("Attribute limit exceeded in vertex description");
-            delete[] attributes;
-            return nullptr;
+            return VertexDescription();
```
375gnu, line 367: return {}

src/celmodel/modelfile.cpp
```diff
@@ -382,29 +383,24 @@ AsciiModelLoader::loadVertexDescription()
         if (!valid)
         {
             reportError("Invalid vertex description");
-            delete[] attributes;
-            return nullptr;
+            return VertexDescription();
```
375gnu, line 381: {}

src/celmodel/modelfile.cpp
```diff
@@ -382,29 +383,24 @@ AsciiModelLoader::loadVertexDescription()
...
         }
 
         VertexAttributeSemantic semantic = Mesh::parseVertexAttributeSemantic(semanticName);
         if (semantic == VertexAttributeSemantic::InvalidSemantic)
         {
             reportError(fmt::format("Invalid vertex attribute semantic '{}'", semanticName));
-            delete[] attributes;
-            return nullptr;
+            return VertexDescription();
```
375gnu, line 388: {}

src/celengine/meshmanager.cpp
```diff
@@ -301,26 +300,21 @@ ConvertTriangleMesh(const M3DTriangleMesh& mesh,
...
 
-            int dataOffset = (i * 3 + j) * floatsPerVertex;
-            vertexData[dataOffset + 0] = position.x();
-            vertexData[dataOffset + 1] = position.y();
-            vertexData[dataOffset + 2] = position.z();
-            vertexData[dataOffset + 3] = normal.x();
-            vertexData[dataOffset + 4] = normal.y();
-            vertexData[dataOffset + 5] = normal.z();
+            int dataOffset = (i * 3 + j) * vertexSize;
+            std::memcpy(vertexData.data() + dataOffset, position.data(), sizeof(float) * 3);
```
375gnu, line 298: won't be std::copy more correct here?

src/celengine/meshmanager.cpp
```diff
@@ -282,7 +276,7 @@ ConvertTriangleMesh(const M3DTriangleMesh& mesh,
                 if (faceNormals[i].dot(faceNormals[k]) > 0.5f)
                     v += faceNormals[k];
             }
-            vertexNormals[i * 3 + 2] = v.normalized();
+            vertexNormals.push_back(v.normalized());
```
375gnu, line 279: probably emplace_back should be here to avoid an copy

375gnu (changes requested):
i had a lot of comments but stupid github lost them all

upd: actually they are here

src/cel3ds/3dsmodel.h
```diff
@@ -10,26 +10,31 @@
...
-#include <memory>
 #include <string>
 #include <vector>
 
 #include <Eigen/Core>
 
 class M3DColor
 {
- public:
+public:
```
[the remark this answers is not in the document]
ajtribick: I see that the convention of using a 1 space indent or zero spaces is different throughout various header files. The 1-space indent tends to have the annoying effect of causing VS Code to infer a tab size of 3 characters when editing the file, and switching to zero spaces is more consistent with things like case statements in switch blocks. Your call though.

src/celengine/meshmanager.cpp
ajtribick: vertexData is a set of words of (an alias of) `std::uint32_t`, we need it to contain the bit representation of the floats but copy would try to cast the value.

src/cel3ds/3dsmodel.h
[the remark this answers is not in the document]
375gnu: i'd say we have only a few files with zero indent. i admit that 1 space indent is quite unusual style, but if we come to agreement to change it then it's better to make this with a separate PR. For now let's keep existing style.

src/cel3ds/3dsmodel.cpp
ajtribick: That would lead to this call doing a copy rather than a move (verified using a class that prints which constructor/assignment operator is called). Options that keep this as a move are `push_back(std::move(matGroup))` and `emplace_back(std::move(matGroup))`, and the `push_back` version feels more natural to me here.

src/cel3ds/3dsmodel.cpp
```diff
@@ -21,230 +21,274 @@ M3DColor::M3DColor(float _red, float _green, float _blue) :
...
-std::string M3DMaterial::getName() const
+std::string
+M3DMaterial::getName() const
 {
     return name;
 }
 
-void M3DMaterial::setName(std::string _name)
+void
+M3DMaterial::setName(std::string _name)
```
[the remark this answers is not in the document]
ajtribick: Looking at how this method is actually used, a better option would be to treat this as a sink parameter taking `std::string&&`.

src/celmodel/mesh.cpp
ajtribick: This will end up being the `difference_type` of the iterator. For `std::array` this will be `ptrdiff_t`.

src/celmodel/mesh.h
ajtribick: They were present in the class when it was nested and I knew I would need them again in the next commit, so I didn't bother removing them here. If you want I can remove them in this commit and re-add them in the next one but I'm not sure there's much point.

src/celengine/meshmanager.cpp
ajtribick: As the result of calling `normalized` is a temporary, this will call the rvalue overload of `push_back` and do a move. On the other hand, fixed size Eigen objects hold their data directly in an array member, so move and copy will end up doing the same thing.

src/celmodel/mesh.cpp
ajtribick: And we already had this as a function anyway. Removed the duplication.

375gnu (approved):

PR #1193, opened 2021-12-01
Move data translations to own file located in CelestiaContent

375gnu:
See https://github.com/CelestiaProject/CelestiaContent/pull/56 as well. Now translations for the application and for data are independent from each other.


ajtribick (approved):

375gnu:
@levinli303 what do you think about this? Please note that this affects mobile versions.

levinli303:
should be ok, I'll need to update the scripts and.change the use of domain "celestia" to "celestia-data"

375gnu:
@levinli303 please share when you're ready so i can merge

src/celengine/stardb.cpp
```diff
@@ -413,11 +416,18 @@ void StarDatabase::getStarName(const Star& star, char* nameBuffer, unsigned int
...
-            if (i18n && iter->second != _(iter->second.c_str()))
-                strncpy(nameBuffer, _(iter->second.c_str()), bufferSize);
-            else
-                strncpy(nameBuffer, iter->second.c_str(), bufferSize);
+            if (i18n)
+            {
+                const char * local = D_(iter->second.c_str());
+                if (iter->second != local)
+                {
+                    strncpy(nameBuffer, local, bufferSize);
```
levinli303, line 424: should it be strncpy(nameBuffer, local, bufferSize - 1) ?

375gnu: `strncpy` doesn't add the final '\0' if `bufferSize <= strlen(local)` so if you pass `bufferSize - 1` then `nameBuffer` is not guarantied to have final '\0'. That's why the next line is mandatory and why there is no difference whether we pass `nameBuffer` or`nameBuffer - 1` .

levinli303 (approved):

PR #1198, opened 2021-12-05
Use localized names at bottom right

levinli303:


375gnu (approved):
while it's bad practice to add 2 different changes to the same commit

PR #1202, opened 2021-12-05
Experimental Qt6 support

ajtribick:
- Switch from legacy QGLWidget to QOpenGLWidget
- [win32] Use a frameless window 1 pixel larger than the monitor instead of a fullscreen window, in order to avoid issues showing context menus

Using a frameless window with a size just larger than the actual screen solves the invisible menu issue on Windows with QOpenGLWidget, although 1 pixel of window will appear on an adjacent monitor. Making the window exactly the same size as the screen causes the menus to stop working again. For me, this does not seem to be a Linux issue (at least on Kubuntu 21.10), so I only do this workaround on Windows.

The Qt5 compatibility module is required to provide `QRegExp`: the `QRegularExpression::wildcardToRegularExpression` helper method that we'd need to use if we replace this with `QRegularExpression` is not available in Qt5.

I don't have a Mac to test with, so no idea what happens there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,7 @@ option(FAST_MATH      "Build with unsafe fast-math compiller option (Default: of
 option(ENABLE_TESTS   "Enable unit tests? (Default: off)" OFF)
 option(ENABLE_GLES    "Build for OpenGL ES 2.0 instead of OpenGL 2.1 (Default: off)" OFF)
 option(USE_GTKGLEXT   "Use libgtkglext1 for GTK2 frontend (Default: on)" ON)
+option(USE_QT6        "Use Qt6 in Qt frontend (Default: off)" OFF)
 option(USE_GTK3       "Use Gtk3 in GTK2 frontend (Default: off)" OFF)
 
 if(ENABLE_GLES)
--- a/src/celestia/qt/CMakeLists.txt
+++ b/src/celestia/qt/CMakeLists.txt
@@ -10,7 +10,11 @@ if(APPLE AND EXISTS /usr/local/opt/qt5)
   list(APPEND CMAKE_PREFIX_PATH "/usr/local/opt/qt5")
 endif()
 
-find_package(Qt5 COMPONENTS Widgets OpenGL CONFIG REQUIRED)
+if(USE_QT6)
+  find_package(Qt6 COMPONENTS Widgets OpenGLWidgets Core5Compat CONFIG REQUIRED)
+else()
+  find_package(Qt5 COMPONENTS Widgets CONFIG REQUIRED)
+endif()
 
 set(QT_SOURCES
   qtappwin.cpp
@@ -58,7 +62,11 @@ set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTOUIC ON)
 # Find includes in corresponding build directories
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
-qt5_add_resources(RC_SRC "icons.qrc")
+if(USE_QT6)
+  qt6_add_resources(RC_SRC "icons.qrc")
+else()
+  qt5_add_resources(RC_SRC "icons.qrc")
+endif()
 
 if(WIN32)
   set (RES celestia.rc)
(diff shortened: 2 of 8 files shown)


munix9:
Would it be possible to build Qt 5 and Qt 6 in parallel, e.g. with
`-DENABLE_QT5=ON -DENABLE_QT6=ON`
similar to what is already possible with Qt, GTK, GLUT and SDL?

The switch `ENABLE_QT` could be kept as backwards compatibility and treated internally as `ENABLE_QT5`.

munix9:
FYI: I have now tried something:
For openSUSE Tumbleweed, I built gtk2+gtk3+qt5+qt6 packages in one run by the following hack in the rpm-spec file:
```
...
BuildRequires:  cmake(Qt6Core5Compat)
BuildRequires:  cmake(Qt6OpenGLWidgets)
BuildRequires:  cmake(Qt6Widgets)
...
cp -a src/celestia/gtk src/celestia/gtk3
sed -e 's/USE_GTK3/USE_GTK3_ALT/g;s/USE_GTKGLEXT/USE_GTKGLEXT_ALT/g;s/celestia-gtk/celestia-gtk3/g' \
    -i src/celestia/gtk3/CMakeLists.txt
cp -a src/celestia/qt src/celestia/qt6
sed -e 's/USE_QT6/USE_QT6_ALT/g;s/celestia-qt/celestia-qt6/g' -i src/celestia/qt6/CMakeLists.txt
sed -e 's/^\(add_subdirectory(gtk)\)$/\1\nadd_subdirectory(gtk3)/' \
    -e 's/^\(add_subdirectory(qt)\)$/\1\nadd_subdirectory(qt6)/' \
    -i src/celestia/CMakeLists.txt
...
%cmake \
-DENABLE_GTK=ON -DUSE_GTK3=OFF -DUSE_GTKGLEXT=ON -DUSE_GTK3_ALT=ON -DUSE_GTKGLEXT_ALT=OFF \
-DENABLE_QT=ON -DUSE_QT6=OFF -DUSE_QT6_ALT=ON \
...
```

The patch https://patch-diff.githubusercontent.com/raw/CelestiaProject/Celestia/pull/1202.patch was also included.

This is just meant as a test to check out the possibilities.
Build log: https://build.opensuse.org/public/build/home:munix9:test/openSUSE_Tumbleweed/x86_64/celestia/_log

I have not (yet) tested the generated binaries.

ajtribick:
> Would it be possible to build Qt 5 and Qt 6 in parallel, e.g. with `-DENABLE_QT5=ON -DENABLE_QT6=ON` similar to what is already possible with Qt, GTK, GLUT and SDL?
> 
> The switch `ENABLE_QT` could be kept as backwards compatibility and treated internally as `ENABLE_QT5`.

Having separate `ENABLE_QT` and `ENABLE_QT6` was my initial strategy but I couldn't get some of the global variables that get set when running `find_package` to play nicely with each other when both were enabled. I'm sure it's possible but would rather confirm whether QOpenGLWidget-based Celestia (Qt5 and Qt6) is working properly before spending too much additional time wrestling with CMake.

munix9:
> Having separate `ENABLE_QT` and `ENABLE_QT6` was my initial strategy but I couldn't get some of the global variables that get set when running `find_package` to play nicely with each other when both were enabled. I'm sure it's possible but would rather confirm whether QOpenGLWidget-based Celestia (Qt5 and Qt6) is working properly before spending too much additional time wrestling with CMake.

This is not a problem. My approach for the rpm-spec seems to work for qt5/qt6, so I will implement that later for testing.

375gnu (approved):

levinli303:
Screen Shot 2021-12-15 at 11 04 21 PM

there is some difference between QOpenGLWidget and QGLWidget, I think.

ajtribick:
That's the kind of thing I was worried about. Not seeing that behaviour on Windows or Linux. Is that screenshot from a Qt5 or a Qt6 build?

375gnu:
https://doc.qt.io/qt-5/highdpi.html and https://stackoverflow.com/questions/39455504/qt-mainwindow-with-qopenglwidget-in-retina-display-displays-wrong-size#39467033 can be useful

ajtribick:
I've updated the scaling to take the DPI into account. From my quick tests setting the scaling to 125%, this appears to fix the scaling issue on Linux and doesn't break it on Windows. I'm hoping it also works on Mac.

375gnu (approved):

ajtribick:
@levinli303 - is the scaling issue fixed on Mac?

levinli303:
@ajtribick  I will verify this weekend

levinli303 (approved):

levinli303:
the scaling issue seems fixed.

PR #1203, opened 2021-12-05
[tests] Make catch_main shared library

375gnu:
This fixes the following error on my Ubuntu 18.04 with CMake 3.10.2:

CMake Error at cmake/TestCase.cmake:8 (target_link_libraries):
  Target "catch_main" of type OBJECT_LIBRARY may not be linked into another
  target.  One may link only to STATIC or SHARED libraries, or to executables
  with the ENABLE_EXPORTS property set.
Call Stack (most recent call first):
  test/unit/CMakeLists.txt:6 (test_case)


cmake/TestCase.cmake
```diff
@@ -4,6 +4,7 @@ macro(test_case)
   list(REMOVE_AT libs 0 0)
 
   add_executable(${trgt} "${trgt}_test.cpp")
```
ajtribick, line 6: I think this should actually be
```cmake
add_executable(${trgt} $<TARGET_OBJECTS:catch_main> "${trgt}_test.cpp")
```
and remove catch_main from `target_link_libraries`
```

375gnu: yep, that's was right way

ajtribick (approved):

PR #1241, opened 2021-12-14
Return const reference from tokenizer getStringValue

ajtribick:
- Remove `Tokenizer::getNameValue` - it does the same thing as `getStringValue`
- Return `const std::string&` from `getStringValue` instead of `std::string` - avoids unnecessary copies in many cases
- Some relatively small-scale refactoring at the areas where it is used to further reduce string copying during parsing

There are some more heavy-duty changes I want to do in addition to this in order to reduce string copying even further, but this would likely require large-scale changes (e.g. in dsodb.cpp / solarsys.cpp) that would likely conflict with the changes in #1189 - so I'm holding off on those for now.

### DIFF
--- a/src/celengine/asterism.cpp
+++ b/src/celengine/asterism.cpp
@@ -18,8 +18,8 @@
 using namespace std;
 using celestia::util::GetLogger;
 
-Asterism::Asterism(string _name) :
-    name(std::move(_name))
+Asterism::Asterism(string_view _name) :
+    name(_name)
 {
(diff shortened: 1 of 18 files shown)


375gnu:
What about using string_view instead?

ajtribick:
Now we have C++17, I've updated this to use `std::string_view`. Mostly results in similar code, except where the usage ends up being close to a C function (e.g. in scripts).

375gnu (approved):

PR #1243, opened 2021-12-15
Require C++17

### LINKED ISSUE #919: Bump required C++ version to C++17
I don't think that when we release Celestia 1.7 it will be hard to find a system with C++17 support. Ubuntu 16.04 is near it's EOL, 18.04 ships both g++7 and g++8. Debian stable has g++8. Windows users can install VS2019 or mingw-g++ 8. Also LLVM project provides builds for different systems. I don't know how it's supported on macOS/iOS and Android.

But C++17 has a lot of useful features, especially `std::filesystem` so we can drop own implementation and `std::sting_view` which can speedup text files parsing, see `celengine/hash.cpp`.

So @levinli303 what do you think?

ajtribick:
Resolves #919

- Set minimum version to C++17, CMake 3.8
- Remove the `string_view` shim, and the implementation of `filesystem` (the filesystem alias namespace is retained to allow usage of `std::experimental::filesystem` on older compilers)
- Remove the Eigen alignment code (this should be taken care of with C++17 aligned `new`)
- A few minor adjustments: collapse nested namespace blocks and multiple `using`, use inline constexpr to define header constants


ajtribick:
I guessed that would happen, not surprising to see the usual suspects: floating point `<charconv>` and `<filesystem>`. Will make a revised version of this later.

375gnu:
fp in charconv is supported since g++11. and early c++17 compilers don't have filesystem (g++7 i guess)

CMakeLists.txt
```diff
@@ -85,9 +85,9 @@ endif()
 # which are by default disabled in standard C++ mode (like std=c++11 instead of std=gnu11)
 add_definitions(-D_USE_MATH_DEFINES)
 
-# Let CMake handle setting C++11 (since 3.1)
+# Let CMake handle setting C++17 (since 3.1)
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
```
375gnu, line 90: you sh'ld bump minimal cmake version to 3.8

375gnu (approved):

ajtribick:
@munix9 - could you check if this works with your automated builds?

munix9:
> @munix9 - could you check if this works with your automated builds?

looks good (see the respective log files for details => "succeeded"-link)
https://build.opensuse.org/package/show/home:munix9:unstable/celestia

https://build.opensuse.org/package/live_build_log/home:munix9:unstable/celestia/Debian_10/x86_64
https://build.opensuse.org/package/live_build_log/home:munix9:unstable/celestia/Ubuntu_18.04/x86_64
https://build.opensuse.org/package/live_build_log/home:munix9:unstable/celestia/openSUSE_Tumbleweed/x86_64
etc...

ajtribick:
@375gnu - the `SONAR_TOKEN` environment variable did not get set for some reason - is there some configuration specific to pull requests that might cause this to fail?

PR #1245, opened 2021-12-15
Fix floating point from_chars implementation

ajtribick:
Fixes a bug with the backport of charconv: the current implementation calls `strtof` (and friends) on the source. Unfortunately this is a C function which assumes a zero-byte terminator, this does not necessarily exist in the source. Also cleans up some other missing parts of the API, and falls back to the standard library functions if charconv is present (including detecting support for floating point charconv, which was a later addition in many standard library implementations).

- Do not assume a zero-byte terminator exists in the source
- Ignore exponent if format does not contain scientific
- Require exponent if format does not contain fixed
- Add support for hexfloat parsing
- Fail if a plus sign is present
- Add bitwise operations to chars_format
- Avoid use of unnamed namespace in header file
- Detect whether charconv is present and if floating point charconv is supported


ajtribick:
Not sure how the Codacy stuff is set up, but I don't think it makes much sense to analyze the checks directory.

src/celcompat/cc.cpp
```diff
@@ -9,66 +9,254 @@
...
+    if (*ptr == '-')
+    {
+        *(buffer++) = *(ptr++);
+        if (ptr == last) { return { first, std::errc::invalid_argument }; }
+    }
+
+    if (*ptr == 'i' || *ptr == 'I')
+    {
+        if (last - ptr < 3) { return { first, std::errc::invalid_argument }; }
+        std::size_t length = std::min(static_cast<std::size_t>(8), static_cast<std::size_t>(last - ptr));
```
375gnu, line 55: magic constants?

src/celcompat/cc.cpp
```diff
@@ -9,66 +9,254 @@
...
+                if (state == State::ExponentStart) { state = State::Exponent; }
+            }
+            else
+            {
+                break;
+            }
+
+            if (buffer == buffer_end) {
+                // non-standard return code: no space to put string terminator
+                return { first, std::errc::not_enough_memory };
```
375gnu, line 146: you reserve 1 char for '\0' on line 45, so i'm not sure that you should return error when buffer == buffer_end

src/celcompat/cc.cpp
```diff
@@ -9,66 +9,254 @@
...
+
+        if (state == State::ExponentStart) { --buffer; state = State::Fraction; }
+
+        if (fmt == chars_format::scientific && state != State::Exponent)
+        {
+            return { first, std::errc::invalid_argument };
+        }
+    }
+
+    *(buffer++) = '\0';
```
375gnu, line 159: just *buffer

375gnu (changes requested):

src/celcompat/cc.cpp
ajtribick: This indicates there isn't enough space in the buffer to parse the literal. While there is space to write the terminating zero byte to avoid out-of-bounds reads, it will still give the wrong result, e.g. if some value at the 1000th place would change the rounding. Hence this is an error condition.

CMakeLists.txt
```diff
@@ -365,6 +365,16 @@ endif()
 try_compile(HAVE_STRING_VIEW ${CMAKE_BINARY_DIR} "${CMAKE_SOURCE_DIR}/checks/cxxsv.cpp")
 try_compile(HAVE_EXPERIMENTAL_STRING_VIEW ${CMAKE_BINARY_DIR} "${CMAKE_SOURCE_DIR}/checks/cxxsvexp.cpp")
 
+try_compile(HAVE_CHARCONV ${CMAKE_BINARY_DIR} "${CMAKE_SOURCE_DIR}/checks/cxxccint.cpp")
+if(HAVE_CHARCONV)
+  try_compile(HAVE_FLOAT_CHARCONV ${CMAKE_BINARY_DIR} "${CMAKE_SOURCE_DIR}/checks/cxxccfloat.cpp")
+  if(NOT(HAVE_CHARCONV))
```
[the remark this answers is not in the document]
ajtribick: Fixed. CMake syntax is terrible.

src/celcompat/cc.cpp
ajtribick: Fixed.

src/celcompat/cc.cpp
ajtribick: Fixed

src/celcompat/cc.cpp
ajtribick: Added more descriptive comments.

375gnu (approved):

PR #1248, opened 2021-12-16
[ci] enable scanning with sonarcloud

375gnu:


levinli303 (approved):

375gnu:
FYI: i've enabled CI on ARM64 machine, at least some early compile check on hw similar to ones used by android and ios. @CelestiaProject/devs

PR #1254, opened 2021-12-24
Add support for different temperature scales

levinli303:
Kelvin/Celsius/Fahrenheit

### DIFF
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -90,6 +90,47 @@ static const double OneFtInKm = 0.0003048;
 static const double OneLbInKg = 0.45359237;
 static const double OneLbPerFt3InKgPerM3 = OneLbInKg / pow(OneFtInKm * 1000.0, 3);
 
+namespace
+{
+float KelvinToCelsius(float kelvin)
+{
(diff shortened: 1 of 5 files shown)


src/celestia/celestiacore.cpp
```diff
@@ -90,6 +90,47 @@ static const double OneFtInKm = 0.0003048;
...
+
+string KelvinToStr(float value, int digits, CelestiaCore::TemperatureScale temperatureScale)
+{
+    const char* unitTemplate = "";
+
+    switch (temperatureScale)
+    {
+        case CelestiaCore::Celsius:
+            value = KelvinToCelsius(value);
+            unitTemplate = _("{} °C");
```
375gnu, line 120: do these strings need localization?

375gnu (changes requested):

src/celestia/celestiacore.cpp
RacerBG: > do these strings need localization?

Usually they don't need it. Those units are internationally recognized anyway.

levinli303: removed localization

375gnu (approved):

PR #1264, opened 2021-12-27
[ci] add a custom job for *.po files

375gnu:


375gnu:
Running full CI is resource wasteful, but we need to validate translation files.

ajtribick (approved):

PR #1274, opened 2021-12-28
Replace invalid use of memcpy in curveplot

ajtribick:
- Vertex is not trivially copyable, therefore memcpy is not allowed.
- Put local classes and functions into unnamed namespace.
- Remove `using namespace std`, `using namespace Eigen`.
- Prefer C++-style casts.
- Use C++17 std::clamp where appropriate.
- Use logger for debugging

### DIFF
--- a/src/celengine/curveplot.cpp
+++ b/src/celengine/curveplot.cpp
@@ -34,22 +34,24 @@
 #define USE_VERTEX_BUFFER 1
 #endif
 
-#include "glsupport.h"
-#include "curveplot.h"
-#include "shadermanager.h"
+#include <algorithm>
+#include <cmath>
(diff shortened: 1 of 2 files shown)


src/celengine/curveplot.cpp
```diff
@@ -311,7 +299,8 @@ class HighPrec_VertexBuffer
         if (currentStripLength > 1)
         {
             int index = currentPosition * 2;
-            memcpy(&data[index], &data[index - 4], 2 * sizeof(Vertex));
```
ajtribick, line 314: The invalid `memcpy` use is here.

375gnu (approved):

PR #1277, opened 2021-12-28
Remove celmath::clamp in favour of std::clamp

ajtribick:
More C++17 changes


levinli303 (approved):

375gnu (approved):

ajtribick:
Updated since I realised that this also enables removing the celmath dependency from celutil.

375gnu (approved):

PR #1279, opened 2021-12-29
Remove unused celengine code

ajtribick:
Removes the following files, and associated `#ifdef`ed code.

* glcontext (render path handling)
* Experimental particle system
* Experimental HDR rendering

I suspect the latter two may be things we'd want to consider implementing at some point but I'm not sure trying to do archaeology on long-abandoned code is necessarily the right way to go. In any case, it would be possible to revert the relevant commit if we decide to resurrect these at a future date.


375gnu (approved):

375gnu:
nowadays hdr can be done much easier

PR #1280, opened 2021-12-29
Refactoring galaxy and globular code

ajtribick:
A bit of refactoring of the galaxy/globular code, mainly focusing on

- Avoid `using namespace std` and friends
- Use C++-style casts
- Put local types into unnamed namespaces
- Avoid excess indirection

### DIFF
--- a/src/celengine/galaxy.cpp
+++ b/src/celengine/galaxy.cpp
@@ -8,57 +8,80 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-#include "render.h"
-#include "astro.h"
-#include "galaxy.h"
-#include "vecgl.h"
-#include "texture.h"
-#include <celmath/mathlib.h>
-#include <celmath/intersect.h>
-#include <celmath/ray.h>
-#include <celmath/randutils.h>
-#include <celutil/gettext.h>
-#include <celcompat/filesystem.h>
-#include <fstream>
 #include <algorithm>
-#include <random>
 #include <cassert>
+#include <cmath>
+#include <cstddef>
+#include <iterator>
+#include <ostream>
+#include <vector>
+
 #include <fmt/printf.h>
 
-using namespace Eigen;
-using namespace std;
-using namespace celmath;
-using namespace celestia;
+#include <celmath/ellipsoid.h>
+#include <celmath/intersect.h>
+#include <celmath/randutils.h>
+#include <celmath/ray.h>
+#include <celutil/gettext.h>
+#include <celutil/logger.h>
+#include "galaxy.h"
+#include "glsupport.h"
+#include "image.h"
+#include "pixelformat.h"
+#include "render.h"
+#include "texture.h"
+#include "vecgl.h"
 
-static int width = 128, height = 128;
-static const unsigned int GALAXY_POINTS  = 3500;
+namespace vecgl = celestia::vecgl;
 
-static bool formsInitialized = false;
+namespace
+{
 
-static GalacticForm** spiralForms     = nullptr;
-static GalacticForm** ellipticalForms = nullptr;
-static GalacticForm*  irregularForm   = nullptr;
+struct Blob
+{
+    Eigen::Vector4f position;
+    unsigned int    colorIndex;
+    float           brightness;
+};
 
-static Texture* galaxyTex = nullptr;
-static Texture* colorTex  = nullptr;
-
-static void InitializeForms();
-static GalacticForm* buildGalacticForms(const fs::path& filename);
-
-float Galaxy::lightGain  = 0.0f;
-
-bool operator< (const Blob& b1, const Blob& b2)
+bool operator<(const Blob& b1, const Blob& b2)
 {
     return (b1.position.squaredNorm() < b2.position.squaredNorm());
 }
 
-typedef vector<Blob> BlobVector;
+using BlobVector = std::vector<Blob>;
+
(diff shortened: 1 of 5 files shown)


src/celengine/galaxy.cpp
```diff
@@ -8,114 +8,357 @@
...
-    if (i >= 28) hue += 220;
+    float hue = 25.0f * std::tanh(0.0615f * (27.0f - static_cast<float>(i)));
+    if (i >= 28) hue += 220.0f;
     //convert Hue to RGB
     float r, g, b;
     DeepSkyObject::hsv2rgb(&r, &g, &b, hue, 0.20f, 1.0f);
-    pixel[0] = (unsigned char) (r * 255.99f);
-    pixel[1] = (unsigned char) (g * 255.99f);
-    pixel[2] = (unsigned char) (b * 255.99f);
+    pixel[0] = static_cast<unsigned char>(r * 255.99f);
```
375gnu, line 139: i believe here we should use `uint8_t` instead of `unsigned char`, it will be more semantically correct

src/celengine/galaxy.cpp
```diff
@@ -8,114 +8,357 @@
...
+    pixel[1] = static_cast<unsigned char>(g * 255.99f);
+    pixel[2] = static_cast<unsigned char>(b * 255.99f);
+}
+
+
```
375gnu, line 144: let's use only 1 empty line in new of heavily modified code

src/celengine/galaxy.cpp
```diff
@@ -8,114 +8,357 @@
...
+    rgb    = img->getComponents();
+
+    auto& rng = celmath::getRNG();
+    for (int i = 0; i < width * height; i++)
+    {
+        value = img->getPixels()[rgb * i];
+        if (value > 10)
+        {
+            float x, y, z, r2, yy, prob;
+            z  = std::floor(static_cast<float>(i) / static_cast<float>(width));
```
375gnu, line 195: because of all casts it's hard to read such code, lets add temporary variables to hold cast values

src/celengine/galaxy.cpp
```diff
@@ -8,114 +8,357 @@
...
+
+    for (unsigned int eform  = 0; eform <= 7; ++eform)
+    {
+        float ell = 1.0f - static_cast<float>(eform) / 8.0f;
+
+        // note the correct x,y-alignment of 'ell' scaling!!
+        // build all elliptical templates from rescaling E0
+
+        galacticForms.push_back(buildGalacticForm("models/E0.png"));
+        if (!galacticForms.back().has_value()) {
```
375gnu, line 325: please move `{` to the next line

src/celengine/galaxy.cpp
```diff
@@ -128,192 +371,115 @@ void Galaxy::setDetail(float d)
 }
 
 
-void Galaxy::setCustomTmpName(const string& tmpNameStr)
+void Galaxy::setCustomTmpName(std::string&& tmpNameStr)
```
375gnu, line 374: maybe just std::string? (or even std::string_view?).

src/celengine/galaxy.cpp
```diff
@@ -128,192 +371,115 @@ void Galaxy::setDetail(float d)
...
 {
-    if (customTmpName == nullptr)
-        customTmpName = new string(tmpNameStr);
-    else
-        *customTmpName = tmpNameStr;
+    customTmpName = std::move(tmpNameStr);
```
375gnu, line 376: this change is not equal. with it every `Galaxy` object will be 24 bytes bigger (on amd64)

375gnu (changes requested):
plz see my comments, i haven't checked everything yet

src/celengine/galaxy.cpp
ajtribick: Turns out we don't actually need this field at all, I'm going to get rid of it.

src/celengine/galaxy.cpp
ajtribick: This is essentially a sink method: I prefer to consider things like `std::string` and `std::vector` as "morally" like `std::unique_ptr`, and have the API set up to do moves rather than copy. Most of the time the calling code doesn't need the value afterwards anyway, so copying (which would happen with `const std::string&`) incurs unnecessary allocations.

Nevertheless, this method is completely unnecessary and going away.

src/celengine/globular.cpp
```diff
@@ -214,31 +452,23 @@ float Globular::getConcentration() const
 void Globular::setConcentration(const float conc)
 {
     c = conc;
-    if (!formsInitialized)
-        InitializeForms();
-
     // For saving time, account for the c dependence via 8 bins only,
 
-    form = globularForms[cSlot(conc)];
+    static const std::vector<GlobularForm> globularForms = initializeForms();
```
375gnu, line 457: afair, google style guide forbids global/static non-trivial variables, i also remember that we had application failures on exit due to global `std::string`s.

375gnu (approved):
please see my concern regarding static vector

src/celengine/globular.cpp
ajtribick: Presumably if we're following the Google style guide we should remove all use of `<filesystem>`, as they explicitly forbid that...

Will look at an alternative way to do this.

ajtribick:
Added some manager classes to avoid the potential for destructor issues.

375gnu (approved):

PR #1281, opened 2021-12-29
[ci] run sonar scanner on a self-managed runner

375gnu:
* legacy frontends won't be checked, only sdl & qt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,10 +3,10 @@ name: Celestia
 on:
   push:
     branches: [ master ]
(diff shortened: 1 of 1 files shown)


ajtribick (dismissed):

375gnu:
i'm not sure if it worth to use this runner as it takes ~80 mins to run (the machine has only 1 cpu with 1GB memory).

ajtribick:
Oh wow, didn't notice that timestamp

375gnu:
Now 9 minutes, this is much-much better

levinli303 (approved):

PR #1282, opened 2021-12-29
Update .clang-format

375gnu:
I propose to make this style mandatory for all our new code. It's based on our current style, but a few rules might be found controversial. Please leave comments, raise your concerns.


.clang-format
```diff
@@ -1,47 +1,63 @@
...
-AllowAllParametersOfDeclarationOnNextLine: false
-AllowShortFunctionsOnASingleLine: true
-AllowShortIfStatementsOnASingleLine: false
-AllowShortLoopsOnASingleLine: false
-AlwaysBreakBeforeMultilineStrings: false
-AlwaysBreakTemplateDeclarations: true
-BinPackParameters: false
-BreakBeforeBinaryOperators: false
+---
+AccessModifierOffset: '-3'
```
ajtribick, line 2: Presumably this gives us the 1 space indent for the access modifiers. I think we should switch this to -4 and reformat the code.

.clang-format
```diff
@@ -1,47 +1,63 @@
...
+AlignAfterOpenBracket: Align
```
ajtribick, line 3: I wonder if we should switch to `AlwaysBreak` - the hanging indent style ends up with everything far off to the right of the window, which can get a bit annoying.

375gnu:
@levinli303 what do you think?

.clang-format
```diff
@@ -1,47 +1,63 @@
...
+AlignEscapedNewlines: Left
+AlignOperands: 'true'
+AlignTrailingComments: 'true'
+AllowAllArgumentsOnNextLine: 'false'
+AllowAllConstructorInitializersOnNextLine: 'false'
+AllowAllParametersOfDeclarationOnNextLine: 'false'
+AllowShortBlocksOnASingleLine: 'false'
+AllowShortCaseLabelsOnASingleLine: 'false'
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: WithoutElse
```
375gnu, line 16: maybe still be better to return to `false`

.clang-format
375gnu: with `AlwaysBreak` it also looks ugly :)

levinli303 (approved):

375gnu:
@ajtribick i've updated according with your review, plz check

ajtribick (approved):
Ok, let's see how this goes...

PR #1287, opened 2021-12-29
Clamp universal position to within 2^62 microlightyears on x, y and z axes

### LINKED ISSUE #1284: Weird zooming out bug
- [ ] When you zoom out really far it does this. Please fix https://user-images.githubusercontent.com/96560141/147693196-24e9e602-4f63-40c2-b203-14a65774bb7e.mp4

ajtribick:
Resolves #1284 by stopping the observer if their coordinates would go beyond a box with invisible walls at x, y, z = ±2^62 microlightyears (~1400 gigaparsecs).


src/celengine/observer.cpp
```diff
@@ -1375,7 +1375,13 @@ Vector3f Observer::getPickRayFisheye(float x, float y) const
 // universal coordinates.
 void Observer::updateUniversal()
 {
+    UniversalCoord oldPositionUniv = positionUniv;
     positionUniv = frame->convertToUniversal(position, simTime);
+    if (positionUniv.clampRange())
+    {
+        positionUniv = oldPositionUniv;
```
375gnu, line 1382: so the new position is not used at all? in this case i propose instead of clamping provide a method `isOutOfBounds` which will only check if the value is out of bounds.
on the other hand, what if observer position is updated by a script? in this case it will not move at all. maybe it will be better to "stop"  it near the "invisible wall"?

375gnu (changes requested):

src/celengine/observer.cpp
ajtribick: Yeah, this is a bit of a remnant of my original approach which led to the observer sliding into the corner of the bounding cube, which felt really weird.

src/celengine/univcoord.h
```diff
@@ -81,9 +85,9 @@ class UniversalCoord
     Eigen::Vector3f offsetFromLy(const Eigen::Vector3f& v) const
     {
         Eigen::Vector3f vUly = v * 1.0e6f;
-        Eigen::Vector3f offsetUly((float) (x - (BigFix) vUly.x()),
-                                  (float) (y - (BigFix) vUly.y()),
-                                  (float) (z - (BigFix) vUly.z()));
+        Eigen::Vector3f offsetUly(static_cast<float>(x - static_cast<BigFix>(vUly.x())),
```
375gnu, line 88: all these static_casts make my eyes bleed. could you introduce methods `toFloat()` & `toDouble` as we have `toString()` and use the instead?
```
Eigen::Vector3f offsetUly((x - BigFix(vUly.x()).toFloaf(),...)
```

375gnu (changes requested):

src/celengine/univcoord.h
ajtribick: In this case, I can replace the `static_cast<BigFix>` with just a constructor call, it looks a bit neater now. I'd rather not duplicate functionality between the operators (used in various other places in the code base) and methods

375gnu (approved):

PR #1290, opened 2021-12-30
Move SphereMesh initialization into constructor

ajtribick:
Another attempt to convince SonarScanner that there isn't a division-by-zero at line 163 of spheremesh.cpp


src/celengine/spheremesh.cpp
```diff
@@ -31,36 +31,27 @@ constexpr const int MinSlices = 3;
...
                        DisplacementMapFunc func,
-                       void* info)
+                       void* info) :
+    nRings(std::max(_nRings, MinRings)),
+    nSlices(std::max(_nSlices, MinSlices))
 {
-    createSphere(1.0f, std::max(MinRings, _nRings), std::max(MinSlices, _nSlices));
+    nVertices = nRings * (nSlices + 1);
+    nIndices = (nRings - 1) * (nSlices + 1) * 2;
+    vertices = std::unique_ptr<float[]>(new float[nVertices * 3]);
```
375gnu, line 40: make_unique?

375gnu: you also can move these assignments to constructor initializer block

ajtribick: don't think that works with arrays

375gnu:
``` 
   // Create a unique_ptr to an array of 5 elements
    std::unique_ptr<Vec3[]> v3 = std::make_unique<Vec3[]>(5);
 ```

Comment (ajtribick) [reply src/celengine/spheremesh.cpp:40]: Ah right, it was arrays of *known size* at compile time...

**Comment by 375gnu** [comment]
It seems that sonar incorrectly found out of bounds access:
```
nVertices = nRings * (nSlices + 1);
auto *faceCounts = new int[nVertices];
...
for (i = 0; i <= nSlices; i++)
{
vertex = (nRings - 1) * (nSlices + 1) + i;
faceCounts[vertex] = nSlices;
}
```
```
nVertices = nRings * nSlices + nRings;
vertext = (nRings - 1) * (nSlices + 1) + i => nRings * nSlices - nSlices + nRings - 1 + i => nVertices - nSlices + i - 1
(nVertices - nSlices - 1) <= vertex <= (nVertices - 1)
```

ajtribick:
Maybe if I express `vertex` in terms of `nVertices`?

ajtribick:
ok that worked, now to address the other stuff

ajtribick:
If SonarScanner complains about anything in meshmanager.cpp, I'm not fixing it in this pull request, too much scope creep.

ajtribick:
Much as I'd like for this to turn into a never-ending quest to write tests for all the legacy code in Celestia, I'm not going to do that as part of this pull request.

375gnu (approved):

PR #1292, opened 2021-12-31
Backport C++20 numbers header

ajtribick:
Mathematical constants. I've used the `celestia::numbers` namespace rather than `celestia::compat` as this namespace contains the identifier `e`, and I wanted to keep that isolated from everything else.

### DIFF
--- a/src/celmath/randutils.h
+++ b/src/celmath/randutils.h
@@ -5,6 +5,7 @@
 
 #include <Eigen/Core>
 
+#include <celcompat/numbers.h>
 #include "mathlib.h"
 
 namespace celmath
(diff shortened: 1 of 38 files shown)


src/celmath/randutils.h
```diff
@@ -26,6 +27,16 @@ struct RealDists
     static std::uniform_real_distribution<T> SignedFullAngle;
 };
 
+namespace detail
+{
+template<typename T>
+std::uniform_real_distribution<T> createFullAngle()
+{
+    using celestia::numbers::pi_v;
```
ajtribick, line 35: In case we ever have a custom type with its own `pi_v`

375gnu (approved):

PR #1295, opened 2022-01-01
[qtwin] Fix crash on pasting screenshots

ajtribick:
Workaround for an issue with the Windows Qt version where Celestia crashes when a copied screenshot is pasted into another application. This appears to be because the data sent to the clipboard still contains a reference to the data in the destructed Image object, despite the Qt documentation stating that QClipboard::setImage causes the clipboard to take ownership of the data. This does not occur if the QImage manages its own buffer.

(Issue observed with Qt 5.15.2, Windows 10, x64)

### DIFF
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -4745,22 +4745,25 @@ View* CelestiaCore::getViewByObserver(const Observer *obs) const
     return nullptr;
 }
 
-Image CelestiaCore::captureImage() const
+void CelestiaCore::getCaptureInfo(std::array<int, 4>& viewport, celestia::PixelFormat& format) const
 {
-    // Get the dimensions of the current viewport
-    array<int, 4> viewport;
-    getRenderer()->getViewport(viewport);
+    renderer->getViewport(viewport);
+    format = renderer->getPreferredCaptureFormat();
+}
 
-    PixelFormat format = renderer->getPreferredCaptureFormat();
-
-    Image image(format, viewport[2], viewport[3]);
-    if (!renderer->captureFrame(viewport[0], viewport[1],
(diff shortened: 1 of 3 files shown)


src/celestia/celestiacore.cpp
```diff
@@ -4745,22 +4745,25 @@ View* CelestiaCore::getViewByObserver(const Observer *obs) const
...
-    PixelFormat format = renderer->getPreferredCaptureFormat();
+    renderer->getViewport(viewport);
+    format = renderer->getPreferredCaptureFormat();
+}
 
-    Image image(format, viewport[2], viewport[3]);
-    if (!renderer->captureFrame(viewport[0], viewport[1],
-                                viewport[2], viewport[3],
-                                format, image.getPixels()))
+bool CelestiaCore::captureImage(std::uint8_t* buffer,
```
375gnu, line 4754: std::byte?

375gnu (changes requested):

src/celestia/celestiacore.cpp
ajtribick: Using `std::byte` would introduce a bit of friction with both Celestia's `Image` class and Qt's `QImage` (which are the relevant destinations here), which both use `std::uint8_t`/`unsigned char` to store their data. It might be something to consider doing as a larger refactoring of the `Image` class though (although interacting with `QImage` would still end up needing `reinterpret_cast`).

levinli303:
works alright on Mac

src/celestia/celestiacore.cpp
375gnu: okay, let's refactor here &  Image later

375gnu (approved):

Issue #1305, opened 2022-01-04
[1.7.0] Some ringed planets are transparent


SevenSpheres:
**Describe the bug**
On Celestia 1.7.0, some ringed planets are transparent - the rings and atmosphere are visible, but the planet itself is not. This only happens when ring shadows are turned on. So far I've seen this happen with two planets - Proxima c in [my addon for the system](https://celestia.mobi/resources/item?item=41755833-971A-7F24-F769-C9332075DD00) (also reported by @DaveBowman2001), and Oltoras in @Trolligi112477-ww's [Cethenom system](https://celestia.space/forum/viewtopic.php?p=152078#p152078).

**To Reproduce**
1. Download one of the above addons
2. Go to the planet in question
3. Ensure rings and ring shadows are turned on

**Expected behavior**
The planet should be rendered normally, as in 1.6.2.

**Screenshots**
![oltorasbug](https://user-images.githubusercontent.com/56284207/148107174-a523c877-b7e3-42b0-85a2-cd0e3d5546c3.png)
![Screenshot_20220105-002228](https://user-images.githubusercontent.com/56284207/148107203-2baec8d9-d19b-49c1-96a6-6f9972868b66.png)

**Desktop (please complete the following information):**
 - OS: Windows, Android
 - Frontend: Qt, native Win, Android
 - Version: 1.7.0

PR #1306, opened 2022-01-04
Fix ring shadow shaders for planets with several suns

375gnu:
Closes: #1305

Was broken in 32baed9bc

@SevenSpheres please confirm


SevenSpheres (approved):
issue is fixed

PR #1316, opened 2022-01-08
Fix for SonarScanner out of bound memory access false positive

ajtribick:
Should resolve a SonarScanner bug (actually a false positive, as the infinity literal plus trailing zero byte is shorter than the buffer)

Draft until I figure out what actually works to convince SonarScanner that this cannot overflow.


ajtribick:
Looks like it's not complaining, so hoping it works when it's merged into the codebase.

375gnu (approved):

ajtribick:
...and SonarScanner still reports the same false positive bug on the main build. Great.

PR #1317, opened 2022-01-08
Refactor font handling code

375gnu:
* Add cache for loaded fonts
 * Replace `const std::string &` with `std::string_view` in print
 * Remove static load method
 * Reformat with clang-format
 * Return bool from celestiacore::set*Font

The last one sh'ld be a separate commit, but I lost history during various `git rebase`s.


375gnu:
cppcheck (codacy) is a false positive

375gnu:
ping @ajtribick @levinli303

levinli303 (approved):

src/celttf/truetypefont.cpp
```diff
@@ -646,69 +708,89 @@ TextureFont* TextureFont::load(const Renderer *r, const fs::path &path, int inde
...
-    auto pos = fn.rfind(',');
-    if (pos != string::npos)
+    if (auto ps = fn.rfind(','); ps != std::string::npos)
     {
-        size = (int) stof(fn.substr(pos + 1));
-        auto rest = fn.substr(0, pos);
-
-        pos = rest.rfind(',');
-        if (pos != string::npos)
+        if (auto [_, ec] = from_chars(&fn[ps + 1], &fn[fn.size()], size); ec == std::errc())
```
ajtribick, line 733: Simpler to avoid the pseudo-discard and just use
```C++
if (from_chars(&fn[ps + 1], &fn[fn.size()], size).ec == std::errc())
```

src/celttf/truetypefont.cpp
```diff
@@ -646,69 +708,89 @@ TextureFont* TextureFont::load(const Renderer *r, const fs::path &path, int inde
...
         {
-            collectionIndex = stof(rest.substr(pos + 1));
-            return rest.substr(0, pos);
+            if (auto pi = fn.rfind(',', ps - 1); pi != std::string::npos)
+            {
+                if (auto [_, ec] = from_chars(&fn[pi + 1], &fn[pi], index); ec == std::errc())
```
ajtribick, line 737:
```C++
if (from_chars(&fn[pi + 1], &fn[pi], index).ec == std::errc())
```

src/celttf/truetypefont.cpp
```diff
@@ -646,69 +708,89 @@ TextureFont* TextureFont::load(const Renderer *r, const fs::path &path, int inde
...
 // temporary while no fontconfig support
-static fs::path ParseFontName(const fs::path &filename, int &collectionIndex, int &size)
+fs::path
+ParseFontName(const fs::path &filename, int &index, int &size)
 {
     // Format with font path/collection index(if any)/font size(if any)
     auto fn = filename.string();
-    auto pos = fn.rfind(',');
-    if (pos != string::npos)
+    if (auto ps = fn.rfind(','); ps != std::string::npos)
```
ajtribick, line 731: Better to combine the conditions with `&&` to avoid the indentation

src/celttf/truetypefont.cpp
```diff
@@ -646,69 +708,89 @@ TextureFont* TextureFont::load(const Renderer *r, const fs::path &path, int inde
...
+    }
+
+    // Init FontCache
+    static FontCache *fontCache = nullptr;
+    if (fontCache == nullptr)
+        fontCache = new FontCache;
+
+    // Lookup for an existing cached font
+    std::weak_ptr<TextureFont> &font = (*fontCache)[filename];
+    if (font.expired())
```
ajtribick, line 768: I think it's better to use the result of `lock` for the test, rather than testing `expired` and doing the `lock` later

```C++
std::shared_ptr<TextureFont> ret = font.lock();
if (ret == nullptr)
{
    // do the creation
}
return ret;

File: src/celttf/truetypefont.cpp [hunk 5]
```diff
@@ -427,32 +449,30 @@ float TextureFontPrivate::render(wchar_t ch, float xoffset, float yoffset)
...
-        return;
+    if (m_fontVertices.size() < 4) return;
 
-    vector<unsigned short> indexes;
+    std::vector<unsigned short> indexes;
     indexes.reserve(m_fontVertices.size() / 4 * 6);
-    for (unsigned short index = 0;
-         index < (unsigned short) m_fontVertices.size();
-         index += 4)
+    for (unsigned short index = 0; index < (unsigned short)m_fontVertices.size(); index += 4)
```
ajtribick, line 475: Should use `static_cast` here.

ajtribick (changes requested):

src/celttf/truetypefont.cpp
375gnu: what do you mean?

```
    if (auto ps = fn.rfind(','); ps != std::string::npos &&
        from_chars(&fn[ps + 1], &fn[fn.size()], size).ec == std::errc())
    {
```
this?

src/celttf/truetypefont.cpp
375gnu: probably such way is safer

src/celttf/truetypefont.cpp
ajtribick: Never mind, forgot that this doesn't work with the bindings.

ajtribick (approved):

PR #1318, opened 2022-01-09
Enable visual styles for Win32 controls

levinli303:
https://docs.microsoft.com/en-us/windows/win32/controls/cookbook-overview

### DIFF
--- a/src/celestia/win32/CMakeLists.txt
+++ b/src/celestia/win32/CMakeLists.txt
@@ -39,6 +39,8 @@ set(WIN32_HEADERS
   winviewoptsdlg.h
 )
 
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /MANIFEST:NO")
+
 add_library(resources OBJECT res/celestia.rc)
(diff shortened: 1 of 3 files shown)


levinli303:
looks like it works.

new vs old

Screen Shot 2022-01-09 at 9 22 18 PM

SevenSpheres:
Huh, oddly enough 1.7.0-win does look like the "old" image on my end (as did 1.6.1), even though 1.6.2 looks like the "new" image.

levinli303:
@SevenSpheres did you use the one from the artifacts generated by this PR? I just copied the celestia-win.exe from the artifacts to the win1.7.0 installation and it works.

SevenSpheres:
No, I was referring to previous builds. I just saw the "new" image and thought "doesn't it already look like that?" because it does on 1.6.2.

levinli303:
> No, I was referring to previous builds. I just saw the "new" image and thought "doesn't it already look like that?" because it does on 1.6.2.

yes. We used to use this manifest file which would enable visual styles in 1.6.x. but when we switched to cmake, we stopped using the file and cmake defaults to generating a manifest that does not support visual styles.

375gnu (approved):

PR #1321, opened 2022-01-10
Unify point stars size calculation + tiny-tiny optimizations and one fix

375gnu:


375gnu:
Added code to reuse `PointStarVertexBuffer` instead of `Renderer::renderPoint`.

375gnu:
ping @levinli303 @ajtribick

ajtribick (approved):

PR #1326, opened 2022-01-17
[qt] Do not capture menu bar/toolbars when taking screenshots

ajtribick:
Restores the old behaviour that the menu bar/toolbars are not captured when using "grab screenshot" or "copy image".

There may be a similar issue with video capture but this always crashes for me (including with QGLWidget), will look into doing this in another PR.


375gnu (approved):
i hate qt :)

PR #1327, opened 2022-01-18
[win32] Convert decimal point & thousands separator to UTF-8

375gnu:


Askaniy (approved):
tested by Lepestronik (x86, Wine)

ajtribick (approved):

Issue #604, opened 2020-02-10
Autocompletion in search


SevenSpheres:
In Celestia 1.7.0, when you type a partial string into the search box and press Enter, the first object matching that string is selected. For example if I type `jup` and then press Enter, Jupiter is selected. I assume this was intended to be a new feature, but it causes a bug: if I type, say, `eps ind b`, Eps Ind Ba (the first object in the list) is selected, instead of the Eps Ind B barycenter. I think this "feature" should just be removed, and the search should work as it does in 1.6.x.

SevenSpheres:
"Wishlist"? I'm basically asking for a new "feature" to be reverted. There are some other things I'll tag as wishlist though...

SevenSpheres:
"Wishlist"? I'm basically asking for a new "feature" to be reverted.

375gnu:
for me this is not a bug

SevenSpheres:
@375gnu maybe @LukeCEL explained it more clearly in #629?

375gnu:
This was introduced in 1495531a0, we can hide it behind `#ifdef AUTO_COMPLETION`

PR #1329, opened 2022-01-20
Don't use suggestions if object searched is not found

375gnu:
Closes: #604

It seems not all users are happy with this feature so let's hide it behind `#ifdef`.


SevenSpheres:
Since opening the issue I've found that this is sometimes annoying but sometimes useful. Maybe it should be a configurable option.

ajtribick (approved):
Approving this as it does what it sets out to do. I think I actually prefer the autocomplete behaviour, although it would probably be more user-friendly to have the first item highlighted to imply this will be selected.

PR #1333, opened 2022-01-25
CMOD Fixes

yashi:
This PR has two fixes related to 3dstocmod.  Hope you like it.

### DIFF
--- a/src/tools/cmod/3dstocmod/3dstocmod.cpp
+++ b/src/tools/cmod/3dstocmod/3dstocmod.cpp
@@ -15,11 +15,13 @@
 
 #include <cel3ds/3dsread.h>
 #include <celmath/mathlib.h>
+#include <celutil/logger.h>
 
 #include "cmodops.h"
 #include "convert3ds.h"
 #include "pathmanager.h"
 
+using celestia::util::CreateLogger;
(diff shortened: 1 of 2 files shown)


src/tools/cmod/3dstocmod/3dstocmod.cpp
```diff
@@ -15,11 +15,13 @@
 
 #include <cel3ds/3dsread.h>
 #include <celmath/mathlib.h>
+#include <celutil/logger.h>
 
 #include "cmodops.h"
 #include "convert3ds.h"
 #include "pathmanager.h"
 
+using namespace celestia::util;
```
ajtribick, line 24: I think it would be better to just import `GetLogger` rather than the whole namespace here, this is similar to how this is imported in the main part of the codebase.

```C++
using celestia::util::GetLogger;
```

ajtribick (changes requested):

src/tools/cmod/3dstocmod/3dstocmod.cpp
yashi: True. Fixed.

ajtribick (approved):

PR #1335, opened 2022-01-25
SDL UI: new features & cleanup

375gnu:


src/celestia/sdl/sdlmain.cpp
```diff
@@ -515,10 +566,14 @@ int main(int argc, char **argv)
     bind_textdomain_codeset(PACKAGE, "UTF-8");
     textdomain(PACKAGE);
 
-    if (chdir(CONFIG_DATA_DIR) == -1)
+    const char *dataDir = getenv("CELESTIA_DATA_DIR");
+    if (dataDir == nullptr)
+        dataDir = CONFIG_DATA_DIR;
+
+    if (chdir(dataDir) == -1)
```
ajtribick, line 573: It would be good to change this to use the filesystem API so this can also build on Windows, which doesn't have `<unistd.h>`:

```C++
#include <system_error>
#include <celcompat/filesystem.h>
// remove #include <unistd.h>

// ...

std::error_code ec;
fs::current_path(dataDir, ec);
if (ec)
{
    FatalError( /* ... arguments ... */ );
    return 1;
}
```

PR #1338, opened 2022-01-27
Optimize mesh by merging similar primitive groups

375gnu:
This is an attempt to optimize bad models rendering. One example of such models is Hubble.

I'm not very satisfied with the line 419, but CCW rotation is hardcoded in Celestia and cmod format so we can live with it.

I'm not gonna merge this immediately, let's test carefully first.

The next steps are: to merge similar meshes but currently result is similar to abstract painting. And then to integrate https://github.com/zeux/meshoptimizer.


levinli303:
Looks alright

Screen Shot 2022-02-19 at 12 09 42 PM

ajtribick:
Manages to substantially reduce the number of groups in my W UMa binaries add-on. Models still look like they're working.

375gnu:
@ajtribick @levinli303 any objection to merge?

levinli303 (approved):

PR #1347, opened 2022-02-09
Fix broken star browser

levinli303:
StarBrowser is broken because the star position * 1.0e6f is used to compare with current position. Just copied the code from winstarbrowser.cpp

### DIFF
--- a/src/celengine/starbrowser.cpp
+++ b/src/celengine/starbrowser.cpp
@@ -17,29 +17,14 @@
 using namespace Eigen;
 using namespace std;
 
-
-// TODO: More of the functions in this module should be converted to
-// methods of the StarBrowser class.
-
 struct CloserStarPredicate
 {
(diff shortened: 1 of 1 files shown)


levinli303:
ping @375gnu @ajtribick

375gnu:
```
pos =  appSim->getObserver().getPosition().toLy().cast<float>();
where:
    Eigen::Vector3d toLy() const
    {
        return Eigen::Vector3d(static_cast<double>(x),
                               static_cast<double>(y),
                               static_cast<double>(z)) * 1.0e-6;
    }
```

so `pos` is in LY.

`Star::getPosition(t)` is `UniversalCoord::CreateLy(position)` -> `position * 1e6`, so 
`Star::getPosition()` returns LY.

So additional multiplication is not required.

375gnu (approved):

PR #1348, opened 2022-02-18
Fix lua crash in cleanup

levinli303:
Screen Shot 2022-02-18 at 10 51 57 AM


375gnu (approved):

375gnu:
it would be better to investigate why costate is nullptr

levinli303:
> it would be better to investigate why costate is nullptr

seems to be caused by bad script code.

Issue #1343, opened 2022-01-31
3ds models reading is broken


375gnu:
**Describe the bug**
For any 3ds model from the current content repository i see errors "Failed to read 3DS chunk" and nothing is shown.

ajtribick:
Works fine for me on current master (cbdd9de6) with CelestiaProject/CelestiaContent@cc0bc0b. (Windows 10, Ubuntu 21.10, tested with Cassini, Huygens, Galileo, Apollo 11)

375gnu:
I will check

375gnu:
not related to locale. here is a log file
[log.txt](https://github.com/CelestiaProject/Celestia/files/7978380/log.txt)
.

ajtribick:
Did a quick check, Mir also works fine for me. Will take another look later.

ajtribick:
Only thing that comes to mind is that we don't have a .gitattributes file in CelestiaContent. Maybe your Git has decided for some reason that .3ds files are text files and is doing line-ending conversion?

First 16 bytes of mir.3ds should be:
```
4D 4D C1 99 07 00 02 00 0A 00 00 00 03 00 00 00
```

if you've got that 0A converted to 0D 0A then that might be causing the problems.

375gnu:
not my case as i'm on unix

ajtribick:
Well, here's the mir.3ds file I have on my system (currently working), can you check with this?
[mir.zip](https://github.com/CelestiaProject/Celestia/files/7987169/mir.zip)

ajtribick:
@375gnu are you still seeing this issue, including with the attached file? I've added logging and I get the same trace, except that `in` status is good. And the test is running fine in the automated build.

375gnu:
@ajtribick your file is 100% identical to mine

ajtribick:
Ok. Well, since I can't reproduce it on any of my environments, or with the CI, I think you're going to have to debug this one.

dave-kaye:
Hello 375gnu - thank you for fixing the GTK planet rings rendering bug so quickly.

Running Linux openSUSE version Leap 15.2.  For the 3ds models of MIR, Apollo 11, Galileo and Cassini none of them will render in present master in either GTK or QT front ends.

Once again, I took my 1.6.2 GTK compiled front end only and activated it (all other Celestia files are from the compiled master tree) and all of the 3ds models rendered successfully.

[...]

PR #1363, opened 2022-04-30
Rewrite 3DS loader

ajtribick:
Attempt at fixing #1343

Both this code and the original work fine for me, so no idea if this is actually a fix.


dave-kaye:
A second test of the code shows that it does not fix the model rendering issue on some systems.  Earlier reports of it working were incorrect.

ajtribick:
@dave-kaye - sorry, that is an almost completely useless report. I need the model file that's failing (attach it to the post), the log output that occurs during the failure, and some idea of the system you are using to compile (compiler version, OS, architecture).

375gnu (approved):

375gnu:
the last commit fixes 3ds reading for me

PR #1367, opened 2022-05-04
Update GTK solar system browser

dave-kaye:
Add Dwarf Planets and Minor Moons to solar system browser to match Qt version


dave-kaye:
This will update the GTK navigation->solar system browser to include the minor moons and dwarf planets listing that is presently showing as "-" in the listing.  It is already in the Qt version. 

There is an Invisible tag in the GTK version that appears to correspond to the Qt reference point.  I did not add that since to many items that I don't know about could be connected to the Invisible tag.  Reviewer can add if desired.

375gnu (approved):

375gnu:
Thanks @dave-kaye for contribution!

PR #1380, opened 2022-07-21
miniaudio integration for playback

levinli303:
new api in celx

bool isplayingaudio(int channel)
bool playaudio(int channel, string path, optional double startTime, optional float volume, optional float pan, optional bool loop, optional bool nopause)
bool resumeaudio(int channel)
void pauseaudio(int channel)
void stopaudio(int channel)
bool seekaudio(int channel, double time)
void setaudiovolume(int channel, float volume)
void setaudiopan(int channel, float pan)
void setaudioloop(int channel, bool loop)
void setaudionopause(int channel, bool nopause)


levinli303:
compared to ffmpeg, I prefer using this library, as it is header only and easy to integrate and does not introduce more dependencies. it has support for almost all platforms (windows including uwp, apple oses, android, linux). @375gnu let me know what you think?

375gnu:
@levinli303 I'm not a big fan of introducing new dependencies but as both `ffmpeg` and `miniaudio` are optional, we can live with `miniaudio`. What I have doubt about is API incompatibility with Celestia ED, could you provide compatibility with it? https://github.com/CelestiaProject/Celestia-legacy/blob/master/src/celestia/celx.cpp#L3529 and https://github.com/CelestiaProject/Celestia-legacy/blob/master/src/celengine/command.cpp#L999

levinli303:
This should be compatible with celestia:play command now. although I was not sure why https://github.com/CelestiaProject/Celestia-legacy/blob/4fb86d9d8e9bb6e28b3af485a356e93eb91b7b9e/scripts/audio/sound_demo.cel

set loop to -1 in `play { channel 0 volume 1 pan 0 loop -1 filename "music.mp3" nopause 1 }` where apparently loop = 1 is the right way to turn on loop. and it seems in this demo loop should be on.

src/celscript/lua/celx_celestia.cpp
```diff
@@ -2177,11 +2180,292 @@ static int celestia_verbosity(lua_State* l)
...
+        Celx_DoError(l, "Second argument for celestia:setaudionopause must be a boolean");
+        lua_pushboolean(l, false);
+        return 0;
+    }
+
+    bool nopause = Celx_SafeGetBoolean(l, 3, WrongType, "Second argument for celestia:setaudionopause must be a boolean", false);
+    CelestiaCore* appCore = this_celestia(l);
+    appCore->setAudioNoPause(channel, nopause);
+#else
+    Celx_DoError(l, "Audio playback not supported");
```
375gnu, line 2465: Audio playback is not supported

src/celestia/audiosession.h
```diff
@@ -0,0 +1,62 @@
...
+    float volume() const { return _volume; }
+    float pan() const { return _pan; }
+    bool loop() const { return _loop; }
+
+    virtual void updateVolume() = 0;
+    virtual void updatePan() = 0;
+    virtual void updateLoop() = 0;
+
+ private:
+    fs::path _path;
```
375gnu, line 55: We have two styles for private variables naming: without any prefixes in the oldest code and with `m_` prefix in newer one. Could you use `m_` instead of just `_`?

src/celestia/audiosession.h
```diff
@@ -0,0 +1,62 @@
...
+const float maxAudioVolume = 1.0f;
+const float defaultAudioVolume = 1.0f;
+const int minAudioChannel = 0;
+const int defaultAudioChannel = 0;
+}
+
+class AudioSession
+{
+ public:
+    explicit AudioSession(const fs::path &path, float volume, float pan, bool loop, bool nopause);
```
375gnu, line 23: `explicit` is not required here.
```
1) Specifies that a constructor or conversion function (since C++11) or [deduction guide](https://en.cppreference.com/w/cpp/language/ctad) (since C++17) is explicit, that is, it cannot be used for [implicit conversions](https://en.cppreference.com/w/cpp/language/implicit_cast) and [copy-initialization](https://en.cppreference.com/w/cpp/language/copy_initialization).
...
A constructor with a single non-default parameter (until C++11) that is declared without the function specifier explicit is called a [converting constructor](https://en.cppreference.com/w/cpp/language/converting_constructor).
```

src/celestia/miniaudiosession.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace celestia
+{
+
+class MiniAudioSessionPrivate;
+
+class MiniAudioSession: public AudioSession
+{
+ public:
+    explicit MiniAudioSession(const fs::path &path, float volume, float pan, bool loop, bool nopause);
```
375gnu, line 14: `explicit` is not required here.

src/celestia/audiosession.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include <celcompat/filesystem.h>
+
+namespace celestia
+{
+
+namespace constants
```
375gnu, line 8: `constants` is too generic name. either add all audio code into `celestia::audio` or remove `namespace constants`

375gnu (changes requested):

375gnu:
> set loop to -1 in `play { channel 0 volume 1 pan 0 loop -1 filename "music.mp3" nopause 1 }` where apparently loop = 1 is the right way to turn on loop. and it seems in this demo loop should be on.

yes, it looks like an error in the script

src/celestia/audiosession.h
levinli303: removed

src/celestia/audiosession.h
levinli303: yea h I figure so, I added it when the initializer has only one argument, and forgot to delete it.

375gnu (approved):

Issue #1384, opened 2022-07-24
Some hiccup with miniaudio


munix9:
Debian 10 breaks the build with
```
[  364s] In file included from /usr/include/c++/8/bits/locale_conv.h:41,
[  364s]                  from /usr/include/c++/8/locale:43,
[  364s]                  from /usr/include/c++/8/bits/fs_path.h:38,
[  364s]                  from /usr/include/c++/8/filesystem:37,
[  364s]                  from ../src/celcompat/filesystem.h:6,
[  364s]                  from ../src/celutil/filetype.h:14,
[  364s]                  from ../src/celestia/celestiacore.h:18,
[  364s]                  from ../src/celestia/celestiacore.cpp:16:
[  364s] /usr/include/c++/8/bits/unique_ptr.h: In instantiation of 'void std::default_delete<_Tp>::operator()(_Tp*) const [with _Tp = celestia::MiniAudioSessionPrivate]':
[  364s] /usr/include/c++/8/bits/unique_ptr.h:274:17:   required from 'std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = celestia::MiniAudioSessionPrivate; _Dp = std::default_delete<celestia::MiniAudioSessionPrivate>]'
[  364s] ../src/celestia/miniaudiosession.h:34:59:   required from here
[  364s] /usr/include/c++/8/bits/unique_ptr.h:79:16: error: invalid application of 'sizeof' to incomplete type 'celestia::MiniAudioSessionPrivate'
[  364s]   static_assert(sizeof(_Tp)>0,
[  364s]                 ^~~~~~~~~~~
```
Probably because of `gcc 8`, see complete [log](https://build.opensuse.org/public/build/home:munix9:unstable/Debian_10/x86_64/celestia/_log).
(Debian 11 build with `gcc 10` works).

openSUSE Tumbleweed grumbles about
`[  665s] /home/abuild/rpmbuild/BUILD/celestia-1.7.0~git20220724+6088291b/src/celestia/miniaudiosession.cpp:124:1: error: control reaches end of non-void function [-Werror=return-type]`
and aborts the build.
Maybe a `default` block or a `return false;` at the [end](https://github.com/CelestiaProject/Celestia/blob/master/src/celestia/miniaudiosession.cpp#L124) of `bool MiniAudioSession::play(double startTime)` helps here?
[...]

PR #1386, opened 2022-07-26
Add a default case for play

levinli303:
#1384


levinli303:
@munix9 you would still need a newer GCC version to avoid the other issue.

375gnu (approved):

PR #1412, opened 2022-09-03
[qt] add items to filter criteria groupings in SS browser

dave-kaye:


dave-kaye:
Within the QT interface is a View -> Celestial Browser -> Solar System tab.  It has a filter for the objects consisting of types: Planets and moons, Asteroids, Spacecraft and Comets.  The interface uses checkboxes for this filter. Checking a box will display only the items associated with that checkbox.

Since the creation of the browser, moons have been split into Moons and Minor Moons.  The "Planets and moons" checkbox does not display minor moons.  Using this checkbox, Mars has no moons and Jupiter has only four.  This code revision adds the minor moons to the displayed output.

Since the creation of the browser, the classification of Components has been added to the Spacecraft object.  These components will not display when the "Spacecraft" checkbox is checked.  Components are important for spacecraft built of multiple parts such as the International Space Station, the new Chinese space station, Cassini/Huygens, etc.  This code revision adds components to the displayed output.

375gnu (approved):

PR #1416, opened 2022-09-08
[qt] update time toolbar pause button to pause/resume

dave-kaye:


dave-kaye:
Within the QT interface is a toolbar for the control of the timing of the simulation.  It contains a pause button that when pressed pauses the simulation and when pressed again restarts the simulation.  The icon for the button is "||".  In many programs, especially the multimedia players, the pause button is "||" and the play button is ">".  The Celestia pause button keeps the same icon and tool tip "pause time" even if it is functioning as a restart button.  The user does not have any visual cues to indicate the purpose of the button has changed.  Moreover, alongside the pause button is a button with the icon ">" and tool tip "real time" which sets the simulation speed to normal time.  This can be confusing to users, who anticipate thia button will function as a play button.

This code submission converts the pause button to a 2-way button with icon "||" and tooltip "pause time" for pausing the simulation; and icon ">" and tooltip "resume time" for restarting the simulation.  The single button cycles through the two functions, similar to a play/pause button in multimedia.  The submission also creates a new icon for the button that sets simulation speed to real time.

The original pause button is created by a Qt action.  Attempts to directly add the pause/resume function to the slot code caused invalid scope compile errors.  Attempts to pass the button identification through the slot parameters caused the button to fail to work or crash the program.  Attempts to use sender or widget calling functions to identify the button in the slot code also caused the button to fail to work or crash the program.  It appears the issue is in the use of the action method of creating the button.  The action method for button creation was replaced with creating the button, labeling it, adding a click event to signal/slot and adding it to the toolbar.  The name of the button was added to the class information which overcame variable scope issues in the slot code.  While not elegant, it is functional.  However, not using the action method to create the button means the button does not conform to the toolbuttonstyle setting.  If an icon is set the button style is icon, if only a text is set the button style is text.  A boolean variable was added to the class information to indicate if the button style was text.

If this code submission is accepted, the addition of "resume time" to the user interface must be integrated with language translations.  Also, there are 2 issues that still can be addressed in this submission.

The first issue is a compiler directive #if/#else which is permanently set to not compile the text versions of the toolbar buttons.  The text version was not maintained, and was out-of-sync with the icon button version.  This code submission updates the text buttons so both styles are the same.  But if the text buttons are never going to be used, they could be removed from the code and the added coding to track the button style type could also be removed.  This would simplify the code.  Let me know via comments to this submission if you want this to be done.

The second issue is in the fast/slow time button icons.  For slowing time they are "<<|" and "<|", a comparison symbol followed by a vertical line.  For faster time they are ">>>" and ">>", comparison symbols and no vertical line.  It would be simple to create uniform icons either with or without the vertical line.  If you wish this to be done, indicate the desired style via comments to this submission.

src/celestia/qt/qttimetoolbar.cpp
```diff
@@ -68,26 +68,49 @@ TimeToolBar::TimeToolBar(CelestiaCore* _appCore,
...
     addAction(halfTimeAction);
 
-    connect(pauseAction, SIGNAL(triggered()), this, SLOT(slotPauseTime()));
-    addAction(pauseAction);
-
-    connect(realTimeAction, SIGNAL(triggered()), this, SLOT(slotRealTime()));
-    addAction(realTimeAction);
+    pauseResumeBtn = new QToolButton(this);
+    connect(pauseResumeBtn, SIGNAL(clicked()), this, SLOT(slotPauseResumeTime()));
+    if (textButton) {
```
375gnu, line 73: Please use codestyle adopted in Celestia:
```
if ()
{
}
else
{
}
```

375gnu (changes requested):

375gnu:
Regarding the last question, i suppose it's better to have these buttons without bars, i.e. << and <<<.

src/celestia/qt/qttimetoolbar.cpp
dave-kaye: Will modify.  I did look at codestyle and matched it.  I must have gotten some older code by accident.  Sorry.

dave-kaye:
Will make the new buttons and modify code.  Will have to delete and re-issue this submission unless I can find a way to merge additions on the Github web interface.

375gnu:
@dave-kaye don't use github interface, it's usable only for trivial changes. For for serious work you need a local editor and a git client.

dave-kaye:
> @dave-kaye don't use github interface, it's usable only for trivial changes. For for serious work you need a local editor and a git client.

Unable to comply.  Must travel 2 miles to public internet site, have no control over computer software.  Deleting this request and reissuing as new single commit request.  Looks like I could squash merge commits, but not after making pull request.

Issue #1414, opened 2022-09-03
Orbits have gaps when lines are triangulated


375gnu:
**Describe the bug**
First I found this bug with #1402 and I realized that I forgot to add NextPos for the last 2 vertices. But it persisted then I checked `master` and found it there.

**To Reproduce**
1. Update `Renderer::shouldDrawLineAsTriangles()` to always return `true` or use a device with max line thickness equal to `1px`.
2. Enable planet orbits
3. Disable fading orbits
4. Scroll to see all planets of the Solar system.
5. See error

**Expected behavior**
No gaps.

**Screenshots**
![image](https://user-images.githubusercontent.com/1612688/188286134-f861acb2-4acf-4374-a1f4-2186a4cf7a1b.png)

**Desktop (please complete the following information):**
 - OS: any
 - Frontend: any
 - Version: 1.7.0

**Additional context**
The same orbits with lines drawn as lines doesn't have gaps.

levinli303:
it seems `vec2 transform = normalize(nextPos.xy - thisPos.xy);` resulted in nan, as thisPos and nextPos are the same.

levinli303:
one way to mitigate this issue might be passing in previous point as well, if we see nextPos == thisPos, calculate with prevPos == thisPos.

PR #1417, opened 2022-09-09
Inverse last point's scale value

levinli303:


375gnu (approved):

src/celengine/curveplot.cpp
```diff
@@ -299,8 +299,14 @@ class HighPrec_VertexBuffer
...
-            data[index] = data[index - 4];
-            data[index + 1] = data[index - 3];
+            // append the second to last point again to calculate the last line
+            // segment direction, only position is used
+            data[index].position = data[index - 4].position;
+            data[index + 1].position = data[index - 3].position;
+            // since the last line direction is calculated from last point to
+            // second to last point, set the scales of last point to their inverse
+            data[index - 2].scale = -data[index - 2].scale;
+            data[index - 1].scale = -data[index - 1].scale;
```
375gnu, line 309: it seams `data[index]` and `data[index + 1]` should be used in the left part

375gnu (changes requested):

375gnu (dismissed):

375gnu:
hmm, i don't see differences. orbits still have gaps.

levinli303:
the position for points of triangulated lines is line point position offset by the line width in a certain direction. this direction is calculated by the line segment direction rotated 90 degree. all the points before the last points have the line direction of prev->next. however to compensate the missing next point for last point, we used the point second to last as last point's next point, which causes the direction to be the opposite, this PR fixes this.

I recorded a video before the fix in gpu debugger. the blue point in the geometry window is the selected point in the table below. you can see that before the last two points, for every pair of points, the point above the line center comes first. (on the right side you can see the #1414 issue, but it is not relevant to this PR)

https://user-images.githubusercontent.com/17924287/190142789-191f915f-7d11-41ad-b2cf-10cb388086cd.mov

375gnu (approved):

PR #1418, opened 2022-09-09
Bug 1389 Missing license file

dave-kaye:
add COPYING file to base data sources directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -414,6 +414,7 @@ set(BASE_DATA_SOURCES
   start.cel
   celestia.cfg
   controls.txt
+  COPYING
 )
 install(FILES ${BASE_DATA_SOURCES} DESTINATION ${DATADIR})
 install(FILES "splash.png" DESTINATION "${DATADIR}/splash")


dave-kaye:
The bug exists in the initial CMakeLists.txt which is part of Celestia and under the control of the Celestia development team.  The file COPYING was never transferred to the common data sources area.  This submission modifies the CMakeLists.txt to transfer the file.  Utilizing this change, the GTK interface will successfully display the license file.

NOTE:  This patch is for version 1.7.0 as found in the GitHub source.  The bug is listed as being in version 1.6.1, where Makefile.am is used.  While the bug exists in version 1.7.0 it does not appear to exist in versions 1.6.1 or 1.6.2. (Makefile.am near line 20 appears to be correct.)  I do not run these older versions to verify this.

NOTE:  If you are a user, simply find the location of the file "controls.txt" then copy the file "COPYING" to this location.  You can see if the data directory is properly setup by the menu option Help -> Controls which should display a list of mouse and keyboard controls if the directory is setup properly.


While this submission will fix the problem from the Celestia viewpoint, it may not fix it from the user's viewpoint.  Many Linux distributions have specific locations in the file system where licenses and documentation are located.  The software packagers may install the COPYING file at the distribution's preferred license location, not the default Celestia location.  Moving the file during installation will cause the bug, but keeping the file at the default Celestia location and making copies of the file in the distribution's preferred locations will not cause the bug.  Distribution packagers such as @munix9  need to be aware of this issue.  Also, this fix will add the COPYING file to a different location and automatic scripts utilized by distribution packagers may need to be updated.

I do believe there is value in accepting this submission - thus making the Celestia development team distribution correct; and letting the distribution package managers handle any packaging problem.

A better, long-term solution would be to adopt the method used in the Qt interface.  There the license is discussed in a text box, and the user is referred to the licensing organization for more details.  Therefore there is no link to the COPYING file.  The file is still distributed with the Celestia code as required, but where it resides in the operating system's file tree is no longer important.  As an alternative, the entire license file could be incorporated into the code as a pop-up window and not as a file link.

This discussion is valid for the Qt and GTK interfaces.  As multiple additional interfaces exist, they must be evaluated individually.  It does appear that the Win32 code calls the COPYING file in the same way as the GTK code, but it is unknown if the file is located properly during install.  I do not run the Windows version to confirm; perhaps a Windows user can check for the bug and the validity of the fix.

375gnu (changes requested):
please rename the commit into
```
[cmake] install COPYING file
```

dave-kaye:
Checked version 1.6.2 rpms as distributed by @munix9 on openSUSE Leap 15.3 using Ark which will open the distribution rpms and show the file paths.  The COPYING file is in both locations, and GTK should work properly.

The version 1.7.0 is missing the COPYING file due to the bug in the cmake file on GitHub.  GTK does fail on this version.

The addition of the CMakeLists.txt file to the Celestia data directory may require the simple addition of a line to the spec file during openSUSE packaging in the FILES section to prevent an unpackaged file error.

375gnu:
@dave-kaye i'm talking about the commit message, not the pull request's title:
image

dave-kaye:
> @dave-kaye i'm talking about the commit message, not the pull request's title: 
Sorry, web interface does not allow this action.  Would have to delete and start over.  You have access to the code as I put reviewers can modify in the pull.

dave-kaye:
closing and updating file due to changes in files on the main github branch
reissuing shortly

PR #1419, opened 2022-09-10
Feature/linerender/v2

375gnu:


375gnu:
@levinli303 please build a beta version for mobile devices. I forced line triangulation and it looks good, but I would like to have response from users with real h/w.

375gnu:
@levinli303 regarding your changes made to `celmodel/*`. I don't see any support for lines wider than 1px in `cmod` files, so your changes currently are dead code. Should we revert them?

375gnu:
On the other hand we may want wider lines on high dpi screens.

levinli303:
yeah it looks like indeed there is no call to glLineWidth in model, we can revert them.

levinli303:
@375gnu  there is conflict with master branch, after fixing it, I'll build wip builds for iOS/Android

375gnu:
closing in favor of #1425. @levinli303 please build from it instead

Issue #1389, opened 2022-07-31
Missing licence file


ghost:
**Describe the bug**
The license button is unable to open file copying due to improper installation

![image](https://user-images.githubusercontent.com/94876428/182044356-e05d67b2-1e2f-45a4-a740-f97fce9ece99.png)

**To Reproduce**
1. Go to 'Help'
2. Click on 'About'
3. Scroll down to 'License'

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. GNU/Linux, Windows, macOS] Linux Mint 20.3
 - Frontend: [e.g. Qt, native Win, native Mac]  GTK+ Front End
 - Version: [e.g. 1.7.0] 1.6.1

**Additional context**

375gnu:
@natewrench where you installed Celestia from?

natewrench:
> @natewrench where you installed Celestia from?

celestia.space

375gnu:
@natewrench don't close, as the packages came from celestia.space then we should fix them.

dave-kaye:
Placed a proposed fix in the pull requests.

dave-kaye:
Commit 6863d44 fixed this issue.  @375gnu please close this issue.

dave-kaye:
Commit 6873d44 fixed this issue on 1.7.0, - recent compile works.

Downloaded 1.6.2.2 source rpm from munix9 and compiled it.  It also works.   Version 1.6.1 is very old but seems that it should also work.  Appears that problem is probably in Mint Linux, and the version quoted is also old.

I think the issue is resolved and can be closed.

dave-kaye:
@375gnu or @levinli303 please mark this issue #1389 as corrected and remove it from the list.  I have downloaded and installed the patched 1.7.0 development GIT 12/24/22 code and it works for the GTK About -> license, using my own Linux installation which uses the Cmake file supplied with Celestia.  I have also made a new openSUSE Tumbleweed installation and used munix9 Tumbleweed rpms from the 1.7.0 Celestia 12/4/22 GIT version.  They move the license file to a special directory as practiced by openSUSE.  But the corrected code made a copy of the license file in the location needed by GTK and munix9's version worked as well.  Therefore the issue appears to be resolved in the presently being developed code. 

[...]

PR #1428, opened 2022-09-22
[cmake] install COPYING file

dave-kaye:
CelestiaProject issue #1389 - missing license file 
add COPYING file to base data sources directory via cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -412,6 +412,7 @@ set(BASE_DATA_SOURCES
   start.cel
   celestia.cfg
   controls.txt
+  COPYING
 )
 install(FILES ${BASE_DATA_SOURCES} DESTINATION ${DATADIR})
 install(FILES "splash.png" DESTINATION "${DATADIR}/splash")


dave-kaye:
The bug exists in the initial CMakeLists.txt which is part of Celestia and under the control of the Celestia development team. The file COPYING was never transferred to the common data sources area. This submission modifies the CMakeLists.txt to transfer the file. Utilizing this change, the GTK interface will successfully display the license file.

NOTE: This patch is for version 1.7.0 as found in the GitHub source. The bug is listed as being in version 1.6.1, where Makefile.am is used. While the bug exists in version 1.7.0 it does not appear to exist in versions 1.6.1 or 1.6.2. (Makefile.am near line 20 appears to be correct.) I do not run these older versions to verify this.

NOTE: If you are a user, simply find the location of the file "controls.txt" then copy the file "COPYING" to this location. You can see if the data directory is properly setup by the menu option Help -> Controls which should display a list of mouse and keyboard controls if the directory is setup properly.

While this submission will fix the problem from the Celestia viewpoint, it may not fix it from the user's viewpoint. Many Linux distributions have specific locations in the file system where licenses and documentation are located. The software packagers may install the COPYING file at the distribution's preferred license location, not the default Celestia location. Moving the file during installation will cause the bug, but keeping the file at the default Celestia location and making copies of the file in the distribution's preferred locations will not cause the bug. Distribution packagers need to be aware of this issue. Also, this fix will add the COPYING file to a different location and automatic scripts utilized by distribution packagers may need to be updated.

I do believe there is value in accepting this submission - thus making the Celestia development team distribution correct; and letting the distribution package managers handle any packaging problem.

This discussion is valid for the Qt and GTK interfaces. It does appear that the Win32 code calls the COPYING file in the same way as the GTK code, but it is unknown if the file is located properly during install. I do not run the Windows version to confirm; perhaps a Windows user can check for the bug and the validity of the fix.

375gnu (approved):

PR #1432, opened 2022-09-23
[cmake] add Celestia Guide to installed files

dave-kaye:


dave-kaye:
A Celestia Guide is distributed with the install files.  It is never installed nor transferred to a user accessible location upon installation of Celestia.  The guide was recently updated and includes helpful information for the user during operation of Celestia.  However, it is lost when the installation media is removed from the system after Celestia is installed.

This modification of the CMakeLists.txt file will install the Celestia Guide to the main Celestia data directory where it is available to the user.

375gnu (approved):

PR #1440, opened 2022-10-14
[qt] Update time toolbar

dave-kaye:
Modify icons, update text based toolbar


dave-kaye:
Within the QT interface is a toolbar for the control of the timing of the simulation.  This submission modifies the graphic button icons and updates text based buttons.

In many programs, especially the multimedia players, the pause button is "||" and the play button is ">".  The Celestia pause button has the "||" icon and tool tip "pause time".  Alongside the pause button is a button with the icon ">" and tool tip "real time" which sets the simulation speed to normal time.  This can be confusing to users, who anticipate this button will function as a play button. This submission creates a new icon for the button that sets simulation speed to real time and moves its location on the toolbar.  The buttons for speed and pause are now grouped together similar to multimedia players.

A compiler directive #if/#else is set to not compile the text versions of the toolbar buttons.  The text version was not maintained, and was out-of-sync with the icon button version.  This submission updates the text button code to match the graphic code.

A third issue is in the fast/slow time button icons.  For slowing time they are "<<|" and "<|", a comparison symbol followed by a vertical line.  For faster time they are ">>>" and ">>", comparison symbols and no vertical line.  Per previous discussions, the icons are being made uniform using only comparison symbols.

1 original toolbar
Original toolbar
2 new toolbar
New toolbar

3 text toolbar
Text based toolbar

This submission does not modify the action of the pause button at this time.  It is updating the interface.

375gnu (approved):

levinli303 (approved):

PR #1443, opened 2022-10-15
[ci] enable Qt6 builds and remove azure pipelines

375gnu:
i don't think it makes sense to use two different CI systems


.github/workflows/ci.yml
```diff
@@ -85,14 +90,40 @@ jobs:
...
+        - os: ubuntu-22.04
+          avif: ON
+          gles: ON
+          glut: ON
+          gtk3: ON
+          minaudio: ON
+        - os: macos-latest
+          avif: ON
+          gles: OFF     # TODO: Figure out how to enable
+          glut: OFF     # TODO: Figure out how to enable
```
levinli303, line 114: glut can be turned on I think? I remember I could build it some time ago.

.github/workflows/ci.yml
```diff
@@ -85,14 +90,40 @@ jobs:
...
+          minaudio: OFF # Requires newer g++
+        - os: ubuntu-22.04
+          avif: ON
+          gles: ON
+          glut: ON
+          gtk3: ON
+          minaudio: ON
+        - os: macos-latest
+          avif: ON
+          gles: OFF     # TODO: Figure out how to enable
```
levinli303, line 113: only the iOS-on-Mac apps can use gles. Mac apps cannot access GLES APIs, so no need for this TODO

levinli303 (changes requested):

.github/workflows/ci.yml
375gnu: then i need your help, i tried to enable it first but the build finished with errors (perhaps glut library was not available to the linker)

375gnu: fixed this

375gnu:
@levinli303 any other changes to make?

levinli303 (approved):

Issue #1409, opened 2022-09-01
Black square around the Sun


375gnu:
![image](https://user-images.githubusercontent.com/1612688/187938312-b3bea32f-2fd3-4cc5-84e1-e1e50dea4cf7.png)

I was able to reproduce it with GTK frontend only.

Steps to reproduce: enable planet orbits, go to cel://Follow/Sol:Earth/2022-09-01T14:05:06.49829Z?x=MHL9zpmPEv///////////w&y=NP4T6lFEn////////////w&z=bE++EAGevP///////////w&ow=0.2696041&ox=0.61314183&oy=0.020633211&oz=0.7422567&fov=46.849846&ts=1.0&ltd=0&p=0&rf=136181183&nrf=255&lm=0&tsrc=0&ver=3

Maybe moving up & high will be required. Also depends on distance to the Sun. Much easier to reproduce with a small window:
![image](https://user-images.githubusercontent.com/1612688/187939905-42944c8a-0654-4845-b20b-3c4d47a33b39.png)
cel://Follow/Sol/2022-09-01T14:08:01.61868Z?x=AAB9QrTiPKgR&y=LCAzq32hzP///////////w&z=AIBlhuWIg9cG&ow=0.20013171&ox=0.54564375&oy=-0.145735&oz=0.8006132&select=Sol&fov=23.575752&ts=1.0&ltd=0&p=0&rf=136181183&nrf=191&lm=0&tsrc=0&ver=3

TheLostProbe:
i was able to fully confirm that this does indeed happen with planets. it also occurs around other stars. the bug is visible on planets with both the Fuzzy Points and Scaled Discs rendering options

cel://Follow/TOI-3207:TOI-3207.01/2022-09-02T00:09:44.82531Z?x=AOCydhdqsjP//////////w&y=AFCKtkxEtlb+/////////w&z=AABD6dmrQAoB&ow=-0.4358858&ox=0.41791612&oy=0.28216168&oz=-0.745476&select=TOI-3207:TOI-3207.01&fov=125.02859&ts=1&ltd=0&p=0&rf=138253747&nrf=255&lm=2048&tsrc=0&ver=3
![image](https://user-images.githubusercontent.com/67564416/188032635-f1d2d0a8-e511-4f2c-914a-a8f151bede05.png)
[...]

PR #1446, opened 2022-10-23
Fix blending state when drawing stars and objects as points

375gnu:
Closes: #1409

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -1651,6 +1651,12 @@ void renderLargePoint(Renderer &renderer,
                       float size,
                       const Matrices &mvp)
 {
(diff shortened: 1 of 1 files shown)


375gnu:
@TheLostProbe please confirm that this patch fixes the issue.

TheLostProbe:
its gone. good riddance black square bug!
image
image
image

levinli303 (approved):

levinli303:
@375gnu it seems we should still keep the one in `renderObjectAsPoint` to fix https://github.com/levinli303/AndroidCelestia/issues/47

375gnu:
@levinli303 please see my comment in #1454

PR #1447, opened 2022-10-23
Feature/linerender/v4

375gnu:


375gnu:
@levinli303 could you check here if you observe that issue with darkened orbits with this branch?

TheLostProbe:
i tested on desktop and the bug appears to be present, but it sort of flickers in a really weird way, and it only occurs at certain distances. also some depth sorting bugs with the orbits are present. i guess we will wait and see if this is apparent on mobile as well
https://user-images.githubusercontent.com/67564416/197439341-59aee95d-fa23-48d7-b62a-9cdf80135ffa.mp4

375gnu:
@TheLostProbe @levinli303 could you also check the same with `master` branch?

TheLostProbe:
I'm in bed on my phone so Lev will have to check

levinli303:
master branch does not have the issue.

levinli303:
Seems to have been fixed with the latest commit to this branch.

levinli303:
assertion has been hit starting up with ANGLE (drawing line as triangles)

Screen Shot 2022-10-26 at 18 59 01

375gnu:
@levinli303 i've removed assertions as they always fail because the vectors are freed.

TheLostProbe:
a thought came to my mind regarding the optimization of this line renderer, since the current one lags when there are too many orbits/labels on the screen. of course, this is to be expected, but it should at least not lag that badly when trying to view for example jupiter's minor moon system, which it does when using the current line renderer

375gnu:
@TheLostProbe for better precision we should use 64bit floats (doubles) instead of 32bit floats supported by GPUs. So all computation are done on CPU. Taking into account that most of the current CPUs are multicore we can try to parallelize computations and maybe even offload some computations to GPU, but this is not related to this change.

375gnu:
@levinli303 anything else to fix?

levinli303:
I'll verify on ANGLE later.

levinli303:
symbols are missing for reference vectors, XYZ, etc

Screenshot 2022-11-08 at 20 26 50

375gnu:
@levinli303 letters should be back now

PR #1454, opened 2022-11-16
Add back setting blend state in renderObjectAsPoint

levinli303:


levinli303:
fix https://github.com/levinli303/AndroidCelestia/issues/47

levinli303:
@375gnu it seems `renderBackgroundAnnotations` can be between `renderPointStars` and `renderObjectAsPoint` which caused the pipeline state to have changed.

375gnu:
let me check

levinli303:
@375gnu what do you think? Between `renderPointStars` and `renderSolarSystemObjects`, there were a few `renderBackgroundAnnotations` which has changed the pipeline state. but in the `renderObjectAsPoint`, the state was not changed back.

375gnu:
i haven't time to check yet and i'm afraid till next week i won't have it

levinli303:
@375gnu  take a look now?

375gnu:
I'm sorry. Not yet.

levinli303:
@375gnu  https://github.com/CelestiaProject/Celestia/blob/97d9df3582762f4d587a2b4d09c101d37e884f1e/src/celengine/pointstarvertexbuffer.h#L83 `render` might be called in`addStar` immediately.

375gnu:
You are correct. But the buffer capacity is 2048 stars. Can we have so many visible solar system objects?

levinli303:
> You are correct. But the buffer capacity is 2048 stars. Can we have so many visible solar system objects?

@TheLostProbe I guess the add-on you use hit that limit.

TheLostProbe:
> @TheLostProbe I guess the add-on you use hit that limit.

the addon adds 99000 stars in a tight globular cluster, so i would say that's definitely the case

PR #1470, opened 2022-12-28
Use iosfwd in header files, do some refactoring

ajtribick:
Various refactorings:

* Replace uses of `<iostream>` with `<iosfwd>` in header files
* Remove some usages of `using namespace std` and `using namespace Eigen`
* Some other refactoring in the cpp files associated with header files that previously contained `<iostream>`


375gnu (approved):

375gnu:
@ajtribick should be me merge?

ajtribick:
Yep, or feel free to send me an invite as I don't have membership anymore.

375gnu:
welcome back )

PR #1471, opened 2022-12-28
Change spelling from 'Synch Orbit' to 'Sync Orbit'

### LINKED ISSUE #1464: Spelling problem in Qt
when right-clicking on an object in the celestial browser, the "sync orbit" option in the right-click submenu is spelled as "synch orbit". both spellings are correct, but it should be standardized. ideally use the "sync" spelling instead of "synch", because the "synch" spelling is _very_ rarely used

**To Reproduce**
1. open the celestial browser
2. right click on an object in the list, doesnt have to be any one in particular
3. see error

**Expected behavior**
"synch orbit" should be "sync orbit" as it is more commonly used

**Screenshots**
image

**Desktop**
 - OS: Windows 10 Home 21H2 build 19044.2251
 - Frontend: Qt
 - Version: 1.7.0, latest `master` build

UPDATE: this also shows up in the right-click menu that pops up when right-clicking on a planet

ajtribick:
Resolves #1464

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -2615,7 +2615,7 @@ msgid "&Follow"
 msgstr "&التابع"
 
 #: ../src/celestia/qt/qtselectionpopup.cpp:153
-msgid "S&ynch Orbit"
+msgid "S&ync Orbit"
 msgstr "ت&زامن المدار"
 
 #: ../src/celestia/qt/qtselectionpopup.cpp:164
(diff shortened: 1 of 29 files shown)


ajtribick:
Unfortunately some languages have different translations for the equivalent menu items in the Windows and Qt interfaces, and changing the spelling from "Synch" to "Sync" requires these to be merged. Since I am unfamiliar with the languages in question, I'm going to leave this.

PR #1475, opened 2022-12-29
Refactor string parser in stardb to remove C string functions

ajtribick:
The static analysis tools don't like the C string functions very much, better to use `std::string_view` for a lot of this.

### DIFF
--- a/src/celengine/stardb.cpp
+++ b/src/celengine/stardb.cpp
@@ -212,7 +212,7 @@ std::string catalogNumberToString(AstroCatalog::IndexNumber catalogNumber)
 
 
 void stcError(const Tokenizer& tok,
-              const std::string& msg)
+              std::string_view msg)
 {
     GetLogger()->error(_("Error in .stc file (line {}): {}\n"), tok.getLineNumber(), msg);
 }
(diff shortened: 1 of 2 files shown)


ajtribick:
Yet another FlawFinder false positive. The code isn't in a loop and it's explicitly providing the buffer size, so the warning isn't relevant. My suspicion is that FlawFinder thinks this is the C standard library `read` call, not the `read` method on a `std::istream`.

src/celengine/stardb.cpp
```diff
@@ -47,107 +54,171 @@ constexpr const char LacailleCatalogPrefix[]  = "Lacaille ";
...
+        AstroCatalog::IndexNumber tyc2 = catalogNumber / TYC2_MULTIPLIER;
+        catalogNumber -= tyc2 * TYC2_MULTIPLIER;
+        AstroCatalog::IndexNumber tyc1 = catalogNumber;
+        return fmt::format("TYC {}-{}-{}", tyc1, tyc2, tyc3);
+    }
+}
+
+
+void stcError(const Tokenizer& tok,
+              const std::string& msg)
```
375gnu, line 215: could you check if it makes sense to use std::string_view here as well?

375gnu (changes requested):

src/celengine/stardb.cpp
ajtribick: done

375gnu (approved):

PR #1476, opened 2022-12-29
[qt] Fix nodiscard warning in Qt6

ajtribick:


ajtribick:
One static analysis tool wants me to put the declaration inside the `if` statement, another static analysis tool can't handle the syntax. Can't win.

375gnu (approved):

PR #1477, opened 2022-12-29
Refactor compareIgnoringCase

ajtribick:
Avoid undefined behavior when `char` is signed. Also removes the mixture of signed and unsigned types in the size calculation.


ajtribick:
Ignoring the reported code smells as the suggested fix would make the code less readable IMHO.

375gnu (approved):

PR #1480, opened 2023-01-01
Buffer loading of stars.dat and cross index files

ajtribick:
In my testing this roughly halved the time taken to load the stars.dat file.

### DIFF
--- a/src/celengine/stardb.cpp
+++ b/src/celengine/stardb.cpp
@@ -21,9 +21,10 @@
 #include <fmt/format.h>
 
 #include <celcompat/charconv.h>
-#include <celutil/binaryread.h>
+#include <celutil/bytes.h>
 #include <celutil/gettext.h>
 #include <celutil/logger.h>
+#include <celutil/timer.h>
 #include <celutil/tokenizer.h>
 #include <celutil/stringutils.h>
 #include "meshmanager.h"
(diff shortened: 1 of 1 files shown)


src/celengine/stardb.cpp
```diff
@@ -60,8 +61,8 @@ constexpr inline float STAR_OCTREE_ROOT_SIZE   = 1000000000.0f;
 constexpr inline float STAR_OCTREE_MAGNITUDE   = 6.0f;
 //constexpr const float STAR_EXTRA_ROOM        = 0.01f; // Reserve 1% capacity for extra stars
 
-constexpr inline std::string_view FILE_HEADER            = "CELSTARS"sv;
-constexpr inline std::string_view CROSSINDEX_FILE_HEADER = "CELINDEX"sv;
+constexpr inline std::string_view FILE_IDENTIFIER            = "CELSTARS"sv;
```
375gnu, line 64: it seems that using `inline` makes sense for header files only and it's noop for cpp files.

src/celengine/stardb.cpp
```diff
@@ -77,6 +78,15 @@ constexpr inline AstroCatalog::IndexNumber TYC3_MAX   = 3u;     // from TYC2
 constexpr inline AstroCatalog::IndexNumber TDSC_TYC3_MAX            = 4u;
 constexpr inline AstroCatalog::IndexNumber TDSC_TYC3_MAX_RANGE_TYC1 = 2907u;
 
+// stars.dat file header and record sizes
+constexpr inline std::size_t FILE_HEADER_SIZE = FILE_IDENTIFIER.size() + sizeof(std::uint16_t) + sizeof(std::uint32_t);
+constexpr inline std::size_t FILE_RECORD_SIZE = sizeof(AstroCatalog::IndexNumber)
+                                              + 3 * sizeof(float)
+                                              + sizeof(std::int16_t) + sizeof(std::uint16_t);
```
375gnu, line 85: This looks too fragile. What if we introduce structures for this purpose?
```
struct __attribute__((__packed__)) StarDBHeader
{
    unsigned char id[8];
    std::uint16_t version;
    std::uint32_t counter;
};

// and

struct __attribute__((__packed__)) StarDBRecord
{
    AstroCatalog::IndexNumber indexNumber;
    float x, y, z;
    std::int16_t absMag;
    std::uint16_t spectralType;
};
```

375gnu (changes requested):

src/celengine/stardb.cpp
ajtribick: Yep, I tend to prefer to keep it though in case of needing to move this to a header file later.

375gnu:
And also another question: doesn't make sense to use a bigger buffer (e.g. `lcd(4096, sizeof(StarDBRecord))`) instead of reading records one by one. I think this can boost loading further.

src/celengine/stardb.cpp
ajtribick: Done

375gnu (approved):

PR #1482, opened 2023-01-01
Use GPU memory to render fonts

375gnu:


src/celttf/truetypefont.cpp
```diff
@@ -483,23 +497,34 @@ TextureFontPrivate::flush()
...
+
     glEnableVertexAttribArray(CelestiaGLProgram::VertexCoordAttributeIndex);
     glEnableVertexAttribArray(CelestiaGLProgram::TextureCoord0AttributeIndex);
     glVertexAttribPointer(CelestiaGLProgram::VertexCoordAttributeIndex,
                           2,
                           GL_FLOAT,
                           GL_FALSE,
                           sizeof(FontVertex),
-                          &m_fontVertices[0].x);
+                          reinterpret_cast<void*>(offsetof(FontVertex, x)));
```
ajtribick, line 516: I think it is worth adding `static_assert(std::is_standard_layout_v<FontVertex>)` to the file to ensure that any future changes remain compatible with the use of `offsetof`.

ajtribick (approved):

levinli303 (approved):

src/celttf/truetypefont.cpp
375gnu: done

PR #1483, opened 2023-01-02
Rewrite tokenizer to buffer input

ajtribick:
Results in galaxies.dsc parsing time dropping from 0.34 to 0.24 seconds on my machine

### DIFF
--- a/src/celmodel/modelfile.cpp
+++ b/src/celmodel/modelfile.cpp
@@ -12,6 +12,7 @@
 #include <cassert>
 #include <cstring>
 #include <iostream>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <utility>
@@ -374,7 +375,7 @@ AsciiModelLoader::loadMaterial(Material& material)
             {
                 if (tok.nextToken() != Tokenizer::TokenNumber)
                 {
-                    reportError("Bad property value in material");
+                    reportError(fmt::format("Bad {} value in material", property));
                     return false;
                 }
                 data[i] = tok.getNumberValue();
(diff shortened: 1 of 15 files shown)


src/celutil/tokenizer.cpp
```diff
@@ -8,622 +8,800 @@
...
+    double operator()(std::int32_t i) const { return static_cast<double>(i); }
+    double operator()(double d) const { return d; }
+    double operator()(std::string_view) const { return std::nan(""); }
+    double operator()(const std::string&) const { return std::nan(""); }
+};
+
+
+struct VisitorInteger
+{
+    std::optional<std::int32_t> operator()(std::monostate) const { return std::nullopt; }
```
375gnu, line 144: is it a good idea to provide inconsistent api for the similar purpose? option for integers and actual types for double/string_view?

src/celmodel/modelfile.cpp
```diff
@@ -613,43 +613,61 @@ AsciiModelLoader::loadMesh(Mesh& mesh)
...
-        if (tok.nextToken() != Tokenizer::TokenNumber || !tok.isInteger())
+        if (tok.nextToken() != Tokenizer::TokenNumber)
         {
             reportError("Material index expected in primitive group");
             return false;
         }
 
         unsigned int materialIndex;
-        if (tok.getIntegerValue() == -1)
+        if (std::int32_t tokenValue = tok.getIntegerValue().value_or(-2); tokenValue == -1)
```
375gnu, line 623: how tokenValue can be -1 if you require to return -2?

src/celutil/tokenizer.cpp
ajtribick: Probably not, I've updated this, plus separated out `getNameValue()` and `getStringValue()` to combine the token type tests and data retrieval

src/celmodel/modelfile.cpp
ajtribick: Rewritten the condition, hopefully clearer now.

375gnu (approved):

PR #1492, opened 2023-01-06
Refactor Value and Hash classes

ajtribick:
Various refactoring of the value and parser system. The main points as follows:

* A less hacky way of handling units: store them in the `Value` itself instead of creating additional keys in the containing hash. This does not increase the size of the `Value` class due to the alignment constraints of the `Value::Data` union.
* Fix an issue where moving a `Value` could lead to double deletion.
* Enforce that `Value` creation from a `Hash` or `ValueArray` is done from a `unique_ptr` to indicate ownership.
* C++17 allows storing incomplete types in `std::vector`, so alter the definition of `ValueArray` accordingly, this should ensure that contained values get properly destructed.
* Do type tests and use `std::optional` in the `Value` getters. Some parts of the codebase were already assuming this was the case, which would only be enforced by debug asserts.
* Use comparators with deduced return type (e.g. `std::less<>`) to allow looking up `std::string`-keyed `std::map`s with `std::string_view`.
* Rewrite the `AssociativeArray` (=`Hash`) to return `std::optional` types, or in the case of `getString` a nullable pointer.

My intention is to squash this before merging, keeping it separated for the moment in case any bugs show up.

### DIFF
--- a/src/celengine/asterism.cpp
+++ b/src/celengine/asterism.cpp
@@ -136,7 +136,8 @@ AsterismList* ReadAsterismList(std::istream& in, const StarDatabase& stardb)
         for (const auto& chain : *chains)
         {
             const ValueArray* a = chain.getArray();
-            if (a == nullptr) { continue; }
+            if (a == nullptr)
+                continue;
 
             // skip empty (without or only with a single star) chains
             if (a->size() <= 1)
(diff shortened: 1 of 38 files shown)


375gnu (changes requested):

375gnu:
Does it makes sense to make `Value` a thin wrapper around `std::variant`?

ajtribick:
I initially considered using `variant` but there's no way to do that and store units in the `Value` without increasing the size of the `Value` class. Effectively what this is doing is storing the units in the padding bytes between the discriminant of the variant and the variant data, which can't be done with `variant`.

375gnu (approved):

PR #1499, opened 2023-01-09
test installer generation

levinli303:


.github/workflows/ci.yml
```diff
@@ -58,10 +58,34 @@ jobs:
...
         if ${{matrix.platform}}==x86 set GeneratorPlatform=Win32
-        cmake -B ${{github.workspace}}/build -DCMAKE_TOOLCHAIN_FILE=c:/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=${{matrix.platform}}-windows -DCMAKE_GENERATOR_PLATFORM=%GeneratorPlatform% -DENABLE_SPICE=ON -DENABLE_TOOLS=ON -DENABLE_TESTS=ON -DENABLE_SDL=OFF -DENABLE_QT=ON -DUSE_QT6=%UseQt6% -DENABLE_FFMPEG=ON -DENABLE_MINIAUDIO=ON
+        cmake -B ${{github.workspace}}/build -DCMAKE_TOOLCHAIN_FILE=c:/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=${{matrix.platform}}-windows -DCMAKE_GENERATOR_PLATFORM=%GeneratorPlatform% -DCMAKE_INSTALL_PREFIX=${{github.workspace}}/output -DENABLE_SPICE=ON -DENABLE_TOOLS=ON -DENABLE_TESTS=ON -DENABLE_SDL=OFF -DENABLE_QT=ON -DUSE_QT6=%UseQt6% -DENABLE_FFMPEG=ON -DENABLE_MINIAUDIO=ON -DX_VCPKG_APPLOCAL_DEPS_INSTALL=ON
 
     - name: Build
-      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} -- /maxcpucount:2 /nologo
+      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --target install -- /maxcpucount:2 /nologo
+
+    - name: 'Checkout data'
```
375gnu, line 66: i don't think it's a good idea to add data into the package, because celestia.iss will depend on data files and with breaking changes you will need to update  celestia.iss

celestia.iss
```diff
@@ -69,21 +64,33 @@ Source: "start.cel";    DestDir: "{app}"; Flags: ignoreversion
 Source: "celestia.cfg"; DestDir: "{app}"; Flags: ignoreversion
 Source: "demo.cel";     DestDir: "{app}"; Flags: ignoreversion
 Source: "guide.cel";    DestDir: "{app}"; Flags: ignoreversion
-Source: "intl.dll";     DestDir: "{app}"; Flags: ignoreversion
-Source: "iconv.dll";    DestDir: "{app}"; Flags: ignoreversion
-Source: "lua5.1.dll";   DestDir: "{app}"; Flags: ignoreversion
+Source: "avcodec-58.dll";     DestDir: "{app}"; Flags: ignoreversion
```
375gnu, line 67: is it possible to use just `*.dll`? filenames may change breaking the installer

levinli303: yeah let me try wildcard for all files, they should have been organized already by cmake.

.github/workflows/ci.yml
levinli303: do you think we should have a separate installer for data?

Lepestronik:
Just duplicating my old message from the forum.

Installation and removal generally work.
Only "native" versions are launched.
Qt versions give errors, see screenshots.

image

Text: "Procedure entry point ... not found in the DLL library ..."
image

Note that when I build the program from artifacts (_master_ branch), Qt5 versions work and Qt6 versions did not work at all, therefore, for your automatic installer at least for Qt5 you are able to correct the situation with the libraries.

Remarks:
1) Both on the desktop and in the main menu shortcuts are created only for the "native" version of the application. I think it's right to have a second shortcut for Qt version at the same time, because it's being installed anyway.
2) It is advisable to add a shortcut for the uninstaller to the main menu.

ajtribick:
> It is advisable to add a shortcut for the uninstaller to the main menu.

I disagree with this. It clutters the menu with "Uninstall" items that get in the way when using text search. From what I can see, the general recommendation these days is to not create uninstall links, since you can either right-click the program's icon to uninstall, or use Add/Remove Programs. Doing a quick check of what I have on my machine, the majority of applications do not provide uninstaller links on the Start Menu.

Vaguely relevant Microsoft blog: https://devblogs.microsoft.com/oldnewthing/20190319-00/?p=102342

levinli303:
@Lepestronik  I've updated this PR to

1. Add shortcuts for QT version
2. Add dependencies for QT so the QT version should be working now

https://github.com/CelestiaProject/Celestia/actions/runs/4231973141

Lepestronik:
**Results for Windows 8:**
1) x86 native, Qt5 - work.
2) x86 and x64 Qt6 - do not work (problem with libraries):
x86-x64-Qt6
3) x64 native freezes at startup.
4) x64 Qt5, when installed, replaces version 1.6.2.2 in the program registry and in the menu folder, it is installed where 1.6.2.2 was installed (although I specified different folders "Celestia 1.6.2.2" and "Celestia 1.7.0"), overwriting the data, keeping the folder name! When deleting 1.7.0, this folder is deleted!
5) x64 Qt5, Qt6 -- when installed, the status of the checkboxes for creating shortcuts and associations is identical to the one selected earlier when installing 1.6.2.2.
6) Not corrected dates ("Copyright (c) 2001-2021") in README.txt and in the "About the program" window.
7) The names of the Mir station and the ISS have not been translated.
8) Native -- encoding failure in the menu item "Small Moons" (For example, "Neptune - Satellites - Small Moons"), the names of the satellites have not been translated.

**Suggestions**
1) Specify the version in the shortcut names, installation paths and folders in the menu; while I do it manually;
2) It may be worth making two checkboxes (or one switch) for "Launch Celestia" with the choice of frontend; otherwise it is not clear which frontend will be launched;

levinli303:
1. ok
2. Qt6 does not support win8
3. guess something is wrong here but I have no idea
4. why should they coexist? it is the same app
5. not sure what you mean
6. #1502 
7. not installer issue
8. like #1586, strings need to be fixed in many places for the win32 frontend

1. it's not meant to co-exists with other version, so there is no need to.
2. probably valid, but there are issues shipping two frontends in general, like which frontend should be associated with .cel/.celx file and cel urls

Lepestronik:
> 5\. not sure what you mean

This means that if I previously installed 1.6.2.2 and checked or unchecked there, the same marks will be here. As if **the same instance of the installer**.

> 4\. why should they coexist? it is the same app

Maybe in this case it is necessary to warn the user that the application is already installed with the installation address?

> 2\. which frontend should be associated with .cel/.celx file and cel urls

I do not know the mechanism, but it seems that it is possible to assign several applications to open the same type of files -- one is launched by default (apparently, in the case of Celestia, you should specify _Native_), the rest through the drop-down list in the "Open with" menu.

levinli303:
5. still not sure what you mean. Do you mean that you uncheck something (like file association) when installing 1.6.2, and then run 1.7.0 installer, it is also unchecked?

4. I still do not see a strong need for an update. if anyone is willing to implement this, they can do it

2. I guess it is in those registry entries in the .iss file.

Lepestronik:
> 5\. it is also unchecked

Yes

Issue #1463, opened 2022-12-05
Orbits sometimes cannot be shown


SevenSpheres:
**Describe the bug**
On the latest build, enabling orbits sometimes has no effect, and this behavior persists until the program is restarted. This happens fairly often, but not consistently.

**To Reproduce**
Apparently random, cannot provide steps to reproduce

**Expected behavior**
Orbits should always be visible when enabled.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Frontend: Win/Qt
 - Version: 1.7.0, latest build as of now

TheLostProbe:
I've experienced this as well, but only once

levinli303:
I'm also seeing similar issue. My reproduction method only works on GLES though.

Basically in the demo script when it is zooming out from the Sun, the orbits will not be shown on GLES frontends. I tried to copy the URL, but with the copied URL directly, orbits will be shown.

From the GPU debugger I can see that in_Position for the line drawing commands are all 0,0,0,1
<img width="496" alt="Screenshot 2023-01-12 at 22 36 31" src="https://user-images.githubusercontent.com/17924287/212095126-c1888cb8-67af-481d-8c6e-b8c00cb35df9.png">

@onetwothree any idea?

levinli303:
Closing, reopen if there is a new repro.

PR #1504, opened 2023-01-12
Call enableAttribArrays in unbind if VertexObject is in Initialize state

levinli303:
If no draw call is made between the first bind and unbind, then enableAttribArrays will not be called at all. potential fix for #1463


375gnu (approved):
technically correct, but we need to go deeper and find why we don't call `draw`

levinli303:
This occurs when no vertex is added to CurvePlot

Issue #1489, opened 2023-01-04
Weird mouse drag behaviour with Qt on Wayland


Askaniy:
**Describe the bug**
The actions of camera rotation and target rotation are nearly impossible due to the unexpected reaction. Apparently, the direction of rotation is calculated by averaging the direction over the entire time the mouse button is held down.

This is most likely caused by my too modern configuration. Just installed the program from the [repository for openSUSE](https://download.opensuse.org/repositories/home:/munix9:/unstable/openSUSE_Tumbleweed/).
Note: There is another cursor actions bug that is seen on all devices: #1344 

**Configuration**
 - OS: GNU/Linux, openSUSE Tumbleweed 20221229
 - Desktop environment: KDE Plasma 5.26.4 + _Wayland_
 - Frontend: Qt
 - Version: 1.7.0 c722b7d
 - CPU: AMD Ryzen 5 5600G (x64), Radeon Graphics

ajtribick:
I can reproduce something which sounds similar: occurs when I launch from a Wayland session but works fine under X11.

ajtribick:
The immediate cause of this is fairly straightforward: Wayland does not support warping the cursor, so the call to `QCursor::setPos` in the mouse event handlers fails. This then makes the delta calculations wrong and hence the strange cursor behaviour.

A simple partial fix would be to update the saved local position with the current local position on platforms, except on platforms that support pointer warping (which I assume is at least `cocoa`, `windows`, and `xcb` - see the [documentation of `QGuiApplication::platformName()`](https://doc.qt.io/qt-5/qguiapplication.html#platformName-prop)). This is a partial fix because it would limit the drag distance to the desktop boundaries.

Wayland does support locking the pointer and getting relative mouse motion events via the [pointer constraints protocol](https://wayland.app/protocols/pointer-constraints-unstable-v1) but Qt does not provide an equivalent to this API so would require writing Wayland code directly.

Askaniy:
Just tested, #1507 fixes it 👍

PR #1507, opened 2023-01-14
[qt] Wayland drag handling

ajtribick:
An initial attempt at fixing #1489

* Rewrites the drag handling in Qt so that it only does pointer warping on platforms that support it (Cocoa, Windows, X11)
* On Wayland, use the pointer constraints and relative pointer manager protocols to implement infinite dragging

There is a new CMake option `USE_WAYLAND` to enable this feature (at the moment this defaults to OFF, but it may be worth defaulting this to ON on Linux?) - this requires access to the private Qt headers, plus the Wayland library, protocols and scanner. On Ubuntu Qt5 these are the packages qtbase5-private-dev, libwayland-dev, wayland-protocols and libwayland-bin packages.

So far I have not tested this with Qt6, for which the Ubuntu package would be qt6-base-private-dev.

I have done some brief testing on Wayland, X11 and Windows. It would be useful to have tests with high DPI screens, and on Apple (due to the rewrites all platforms are affected, so I cannot verify that the ALT+dragging behaves as expected).

### DIFF
--- a/src/celestia/qt/qtdraghandler.cpp
+++ b/src/celestia/qt/qtdraghandler.cpp
@@ -1,0 +1,90 @@
+#include "qtdraghandler.h"
+
+#include <QCursor>
+#include <QGuiApplication>
+#ifdef USE_WAYLAND
+#include "qtwaylanddraghandler.h"
+#endif
+
+void
+DragHandler::begin(const QMouseEvent &m, qreal s, int b)
+{
+    saveCursorPos = m.globalPos();
+    scale         = s;
+    buttons       = b;
+}
+
+void
+DragHandler::move(const QMouseEvent &m, qreal s)
+{
+    if (scale != s)
+        begin(m, s, buttons);
+    auto relativeMovement = m.globalPos() - saveCursorPos;
+    appCore->mouseMove(
+        static_cast<float>(relativeMovement.x() * scale),
+        static_cast<float>(relativeMovement.y() * scale),
+        effectiveButtons());
+    saveCursorPos = m.globalPos();
+}
+
+void
+DragHandler::setButton(int button)
+{
+    buttons |= button;
+}
+
+void
+DragHandler::clearButton(int button)
+{
+    buttons &= ~button;
+}
+
+int
+DragHandler::effectiveButtons() const
+{
+#ifdef __APPLE__
+    // On the Mac, right dragging is be simulated with Option+left drag.
+    // We may want to enable this on other platforms, though it's mostly only helpful
+    // for users with single button mice.
+    if (buttons & CelestiaCore::AltKey)
+        return (buttons | CelestiaCore::RightButton) & (~CelestiaCore::LeftButton);
+#endif
+    return buttons;
+}
+
+// Warping drag handler
+
+void
+WarpingDragHandler::move(const QMouseEvent &m, qreal s)
+{
+    if (scale != s)
+        begin(m, s, buttons);
+    auto relativeMovement = m.globalPos() - saveCursorPos;
+    appCore->mouseMove(
+        static_cast<float>(relativeMovement.x() * scale),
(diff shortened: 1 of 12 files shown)


src/celestia/qt/qtdraghandler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    // On the Mac, right dragging is be simulated with Option+left drag.
+    // We may want to enable this on other platforms, though it's mostly only helpful
+    // for users with single button mice.
+    if (buttons & CelestiaCore::AltKey)
+        return (buttons | CelestiaCore::RightButton) & (~CelestiaCore::LeftButton);
+    else
+        return buttons;
+#else
+    return buttons;
+#endif
```
375gnu, line 55: we can simplify this code:
```
#ifdef __APPLE__
    // On the Mac, right dragging is be simulated with Option+left drag.
    // We may want to enable this on other platforms, though it's mostly only helpful
    // for users with single button mice.
    if (buttons & CelestiaCore::AltKey)
        return (buttons | CelestiaCore::RightButton) & (~CelestiaCore::LeftButton);
#endif
    return buttons;
```

src/celestia/qt/qtwaylanddraghandler.h
```diff
@@ -0,0 +1,52 @@
...
+    int                                buttons{ 0 };
+    wl_surface                        *surface{ nullptr };
+    wl_pointer                        *pointer{ nullptr };
+    zwp_relative_pointer_v1           *relativePointer{ nullptr };
+    zwp_locked_pointer_v1             *lockedPointer{ nullptr };
+    bool                               fallback{ false };
+
+    static const zwp_relative_pointer_v1_listener listener;
+
+    static void                                   processRelativePointer(
```
375gnu, line 43: misindent

src/celestia/qt/qtdraghandler.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+void
+WarpingDragHandler::finish()
+{
+    QCursor::setPos(saveCursorPos);
+}
+
+#ifdef USE_WAYLAND
+std::unique_ptr<DragHandler>
+createDragHandler(QWidget *widget, CelestiaCore *appCore)
```
375gnu, line 81: let's use only this signature and mark `widget` as `[[maybe_unused]]`

375gnu (changes requested):

src/celestia/qt/qtdraghandler.cpp
ajtribick: done

src/celestia/qt/qtwaylanddraghandler.h
ajtribick: Fixed... turns out I forgot to commit the file after re-running clang-format to fix the odd indentation it created the first time.

src/celestia/qt/qtdraghandler.cpp
ajtribick: Done.

375gnu (approved):

ajtribick:
@Askaniy could you confirm whether this fixes the issue for you?

Askaniy:
We concluded in the PM that is easiest to test after merging than compiling the branch

ajtribick:
Probably @munix9 will need to incorporate the USE_WAYLAND option in their packaging process.

munix9:
> Probably @munix9 will need to incorporate the USE_WAYLAND option in their packaging process.

Done.  
qt5 + wayland: openSUSE Leap 15.3 + 15.4 + 15.5 + Tumbleweed, Debian 10 + 11, Ubuntu 20.04 + 22.04, AppImage  
qt6 + wayland: openSUSE Leap 15.4 + 15.5 + Tumbleweed

PR #1508, opened 2023-01-14
Migrate more code to use VBO + cleanups

375gnu:
Comet tail and atmosphere rendering were moved to their own files.

Only rectangle rendering is using client-side memory now. I'll take care of it and remove Renderer::getVertexObject in the next pr.

### DIFF
--- a/src/celrender/atmosphererenderer.cpp
+++ b/src/celrender/atmosphererenderer.cpp
@@ -1,0 +1,411 @@
+// atmosphererenderer.cpp
+//
+// Copyright (C) 2001-present, Celestia Development Team
+// Original version Chris Laurel <claurel@gmail.com>
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include <iostream>
+#include <celengine/atmosphere.h>
+#include <celengine/lightenv.h>
+#include <celengine/lodspheremesh.h>
+#include <celengine/render.h>
+#include <celengine/renderinfo.h>
+#include <celengine/shadermanager.h>
+#include <celengine/vecgl.h>
+#include <celutil/color.h>
+#include <celutil/indexlist.h>
+#include "atmosphererenderer.h"
+
+using celestia::util::BuildIndexList;
+using celestia::util::IndexListCapacity;
+using ushort = unsigned short;
+
+namespace
+{
+constexpr int MaxSkyRings = 32;
+constexpr int MaxSkySlices = 180;
+constexpr int MinSkySlices = 30;
+
+constexpr int MaxVertices = MaxSkySlices * (MaxSkyRings + 1);
+constexpr int MaxIndices = IndexListCapacity(MaxSkySlices,  MaxSkyRings + 1);
+}
+
+namespace celestia::render
+{
+AtmosphereRenderer::AtmosphereRenderer(Renderer &renderer) :
+    m_renderer(renderer)
+{
+}
+
+void AtmosphereRenderer::initGL()
+{
+    if (m_initialized)
+        return;
+
+    m_initialized = true;
+
+    m_skyVertices.reserve(MaxVertices);
+    m_skyIndices.reserve(MaxIndices);
+    m_skyContour.reserve(MaxSkySlices + 1);
+
+    m_vo = std::make_unique<IndexedVertexObject>(
+        MaxVertices * sizeof(SkyVertex),
+        GL_STREAM_DRAW,
+        GL_UNSIGNED_SHORT,
+        MaxIndices * sizeof(unsigned short));
+}
+
+void AtmosphereRenderer::deinitGL()
+{
+    m_initialized = false;
+    m_vo = nullptr;
+}
+
+void
+AtmosphereRenderer::computeLegacy(
+    const Atmosphere         &atmosphere,
+    const LightingState      &ls,
+    const Eigen::Vector3f    &center,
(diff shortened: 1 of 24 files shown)


src/celutil/indexlist.h
```diff
@@ -0,0 +1,81 @@
...
+}
+
+template <
+    typename T,
+    std::enable_if_t<std::is_integral_v<T>, bool> = true
+>
+void
+BuildIndexList(T I, T J, std::vector<T> &out)
+{
+    assert(I * J < std::numeric_limits<T>::max);
```
ajtribick, line 32: I think this needs to be done as a division instead, and `max()` needs to be done as a method call:
```cpp
assert(std::numeric_limits<T>::max() / I >= J);
```

ajtribick: Might be worth calling `reserve` on the vector before pushing the nodes?

ajtribick (changes requested):

src/celrender/atmosphererenderer.cpp
```diff
@@ -0,0 +1,399 @@
...
+    m_renderer.setPipelineState(ps);
+
+    if (m_vo->initialized())
+    {
+        m_vo->bindWritable();
+        m_vo->allocate(nullptr, nullptr); // orphan GPU memory buffers
+    }
+    else
+    {
+        initVertexObject();
```
levinli303, line 322: glBufferData is not called in this path

375gnu: it's called. see the next lines

src/celutil/util.h
```diff
@@ -0,0 +1,63 @@
+#include <type_traits>
```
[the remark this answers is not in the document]
375gnu: yes, it was in my previous versions and i forgot to remove it

src/celutil/indexlist.h
375gnu: yes, it does nothing if a new capacity is less then the current, so it's safe to add it

src/celrender/atmosphererenderer.cpp
levinli303: hmm I only see glBufferSubdata getting called in `setBufferData`

375gnu: you are right. it seems mesa supports such cases but they don't look as a proper gl code

src/celrender/atmosphererenderer.cpp
```diff
@@ -0,0 +1,396 @@
...
+
+    m_vo->allocate(nullptr, nullptr); // allocate or orphan GPU memory buffers
+
+    m_vo->setBufferData(m_skyVertices.data(), 0, m_skyVertices.size() * sizeof(SkyVertex));
+    m_vo->setIndexBufferData(m_skyIndices.data(), 0, m_skyIndices.size() * sizeof(unsigned short));
+
+    prog->use();
+    prog->setMVPMatrices(*m.projection, *m.modelview);
+    m_vo->draw(GL_TRIANGLE_STRIP, static_cast<int>(m_skyIndices.size()), 0);
```
ajtribick, line 327: I am getting a segmentation fault here.

375gnu: what object?

ajtribick: Crashes as soon as the Sun comes into view during the startup script.

ajtribick: If I replace this method with just `return` (disabling legacy atmospheres), I can also test the comet tails. They also segfault.

src/celrender/vertexobject.cpp
```diff
@@ -160,9 +154,110 @@ void VertexObject::disableAttribArrays() const noexcept
...
+void
+IndexedVertexObject::bind() noexcept
+{
+    if (getStreamType() != GL_STATIC_DRAW)
+        VertexObject::bind();
+
+    if ((m_state & State::Initialize) != 0)
+        glGenBuffers(1, &m_vioId);
+
+    if (!isVAOSupported() ||
```
ajtribick, line 226: I think the condition here is wrong - as far as I can tell the buffer never gets bound which is why I'm seeing the segfaults. Verified by changing from `!isVAOSupported()` to `isVAOSupported()` which stops the crashes.

375gnu: or index buffer is not bound to vao actually

375gnu: Please check now

375gnu: No, GL_ELEMENT_ARRAY_BUFFER should be managed by Vertex Array Objects

375gnu: Please check now

ajtribick: Segfault no longer occurs, thanks!

src/celrender/cometrenderer.h
```diff
@@ -0,0 +1,64 @@
...
+{
+class IndexedVertexObject;
+
+class CometRenderer
+{
+public:
+    explicit CometRenderer(Renderer &renderer);
+    ~CometRenderer() = default;
+    CometRenderer(const CometRenderer&) = delete;
+    CometRenderer(CometRenderer&&) = default;
```
ajtribick, line 30: Clang 15.0.2 gives a warning because the reference type member implicitly deletes the move constructor and move assignment operators (similarly for `AtmosphereRenderer`)

375gnu: let's remove them explicitly as we don't use them now and if we need them sometime then implement properly.

Issue #1513, opened 2023-01-15
Qt: Save/Open file dialogs are not rendered


375gnu:
**Describe the bug**
Since commit 7cf1aac save and open file dialogs don't work anymore: they don't render anything inside them.

**To Reproduce**
1. Press F10

**Expected behavior**
Dialog should render and allow file selection.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: GNU/Linux
 - Frontend: Qt
 - Version: 1.7.0

**Additional context**
I use XFCE, so Qt uses dialogs provided by Gtk+. With Qt-own dialogs (`QFileDialog::DontUseNativeDialog`) this issue doesn't emerge.

Askaniy:
No problems for me on KDE Plasma

ajtribick:
Works for me on i3, does not work on XFCE

PR #1514, opened 2023-01-15
[qt] Don't use native dialogs

375gnu:
This is a workaround for #1513, so let's not merge it unless we have a better solution.


ajtribick (approved):

375gnu:
This is a really annoying bug so I merge this pr but let's look for fix for DE integration.

PR #1515, opened 2023-01-18
Refactor align.h

ajtribick:
- Avoid creating unaligned pointers before aligning void*
- Use celestia::util namespace
- Replace std::alignment_of with alignof, use constexpr

### DIFF
--- a/src/celutil/align.h
+++ b/src/celutil/align.h
@@ -1,26 +1,28 @@
 #pragma once
 
-#include <type_traits>
+#include <cstddef>
+#include <cstdint>
+
+namespace celestia::util
+{
 
 /*! Returns aligned position for type T inside memory region.
- *  Designed for usage with functions which allocate unalligned
+ *  Designed for usage with functions which allocate unaligned
  *  memory regions.
  */
-template<typename T> T* aligned_addr(T* addr)
+template<typename T> T* aligned_addr(void* addr)
 {
-    size_t align = std::alignment_of<T>::value;
-    return (T*) (((size_t(addr) - 1) / align + 1) * align);
-}
-
-template<typename T> T* aligned_addr(T* addr, size_t align)
-{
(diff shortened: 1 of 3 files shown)


src/celutil/align.h
```diff
@@ -1,26 +1,28 @@
...
 
-#include <type_traits>
+#include <cstddef>
+#include <cstdint>
+
+namespace celestia::util
+{
 
 /*! Returns aligned position for type T inside memory region.
  *  Designed for usage with functions which allocate unalligned
```
375gnu, line 10: please fix `unalligned` as you updated it in other location

375gnu (changes requested):

src/celutil/align.h
```diff
@@ -1,26 +1,28 @@
...
 {
-    size_t align = std::alignment_of<T>::value;
-    return (T*) (((size_t(addr) - 1) / align + 1) * align);
-}
-
-template<typename T> T* aligned_addr(T* addr, size_t align)
-{
-    return (T*) (((size_t(addr) - 1) / align + 1) * align);
+    // alignof(T) is guaranteed to be a power of two
+    constexpr auto align_one = static_cast<std::uintptr_t>(alignof(T) - 1);
```
375gnu, line 16: sonar says this cast is not required. according to https://en.cppreference.com/w/cpp/language/alignof `alignof` returns `size_t`

ajtribick: I disagree with Sonar here: the type of `align_one` needs to be `uintptr_t` rather than `size_t`, as it is being used to manipulate the numerical value of a pointer. Admittedly `uintptr_t` and `size_t` are usually typedefs of the same thing these days.

src/celutil/align.h
ajtribick: done

375gnu (approved):

PR #1516, opened 2023-01-20
celephem refactor

ajtribick:
Refactoring of the celephem code. The relevant types are moved into the `celestia::ephem` namespace, and have done various upgrades to use more C++17 features. There's also a bugfix for the velocity vector direction for linearly-interpolated xyz files.

Check of the output of the custom orbits matches pre- and post-change.

Will squash this before merging, leaving expanded for now in case testing reveals any bugs.


src/celephem/samporbit.cpp
```diff
@@ -297,135 +296,123 @@ template <typename T> Vector3d SampledOrbit<T>::computeVelocity(double jd) const
...
-
-                vel = cubicInterpolateVelocity(p0, v0, p1, v1, t);
-                vel *= 1.0 / h;
-            }
-            else
-            {
-                // Unknown interpolation type
-                vel = Vector3d::Zero();
+            case TrajectoryInterpolation::Linear:
+                vel = computeVelocityLinear(jd, n);
```
ajtribick, line 315: In line 322 of the original, this value was immediately returned, skipping the z/-y swap. The old way looks incorrect to me, but need to check to make sure.

ajtribick: Ok, the original code was incorrect (velocity vector reference mark displayed at 90° to the actual velocity for linear interpolated xyz files)

375gnu (approved):

ajtribick:
Squashed commits

PR #1522, opened 2023-01-26
Fix compiling

levinli303:


375gnu (approved):

ajtribick:
oops... surprised that got past my compiler

PR #1523, opened 2023-01-26
Add Position to the stc file

ajtribick:
This property allows specifying a star's position in rectangular coordinates in the ecliptic reference frame, matching the coordinates used in stars.dat. This enables a 1:1 conversion of stars.dat entries into the .stc file format.

Open to suggestions for a different name for this property.

### DIFF
--- a/src/celengine/stardb.cpp
+++ b/src/celengine/stardb.cpp
@@ -1075,6 +1075,13 @@ bool StarDatabase::createStar(Star* star,
     {
         star->setPosition(barycenterPosition);
     }
+    else if (auto rectangularPos = starData->getLengthVector<float>("Position", KM_PER_LY<double>); rectangularPos.has_value())
+    {
+        // "Position" allows the position of the star to be specified in
+        // coordinates matching those used in stars.dat, allowing an exact
+        // translation of stars.dat entries to .stc.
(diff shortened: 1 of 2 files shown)


375gnu (approved):

375gnu:
Regarding the property name maybe the forum is better place to discuss.

ajtribick:
Was looking at adding this to .dsc files when I noticed that it's already there. In the .dsc file it is called "Position", so might as well use that.

src/celengine/stardb.cpp
```diff
@@ -1075,6 +1075,13 @@ bool StarDatabase::createStar(Star* star,
     {
         star->setPosition(barycenterPosition);
     }
+    else if (auto rectangularPos = starData->getLengthVector<float>("Position", KM_PER_LY<double>); rectangularPos.has_value())
+    {
+        // EclipticXYZ allows the position of the star to be specified in
```
375gnu, line 1080: EclipticXYZ left in comments

375gnu (changes requested):

src/celengine/stardb.cpp
ajtribick: Fixed

375gnu (approved):

PR #1526, opened 2023-01-29
Use JPEG::JPEG target for linking

ajtribick:
Create it if necessary on older CMake versions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,9 +239,10 @@ find_package(PNG REQUIRED)
 add_definitions(${PNG_DEFINITIONS})
 link_libraries(PNG::PNG)
 
+include(FixJPEG)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -239,9 +239,9 @@ find_package(PNG REQUIRED)
 add_definitions(${PNG_DEFINITIONS})
 link_libraries(PNG::PNG)
 
-find_package(JPEG REQUIRED) # -DJPEG_LIBRARY=...
```
375gnu, line 242: please leave this line and remove it it from `cmake/FixJPEG.cmake` so we can easier migrate to newer cmake  when needed

375gnu (changes requested):

CMakeLists.txt
ajtribick: done.

375gnu (approved):

PR #1537, opened 2023-02-03
Use geomerty shaders when available to render galaxies

375gnu:
This requires cleanups but we can test it on different hardware.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,7 @@ jobs:
           ${{github.workspace}}/build/src/**/*.dll
           ${{github.workspace}}/build/po/**/*.gmo
           ${{github.workspace}}/build/tools/**/*.exe
+          ${{github.workspace}}/shaders/*.glsl
 
   build-posix:
     name: "${{matrix.config.os}} with ${{matrix.qtversion}}"
--- a/shaders/galaxy150_frag.glsl
+++ b/shaders/galaxy150_frag.glsl
@@ -1,0 +1,11 @@
+in vec4 v_Color;
+in vec2 v_TexCoord;
+
+uniform sampler2D galaxyTex;
+
+out vec4 v_FragColor;
+
+void main()
+{
+    v_FragColor = vec4(v_Color.rgb, v_Color.a * texture(galaxyTex, v_TexCoord).r);
+}
--- a/shaders/galaxy150_geom.glsl
+++ b/shaders/galaxy150_geom.glsl
@@ -1,5 +1,3 @@
-layout(points) in;
-layout(triangle_strip, max_vertices = 4) out;
 
 uniform mat4 m;
 uniform mat3 viewMat;
--- a/shaders/galaxy150_geom.glsl
+++ b/shaders/galaxy150_geom.glsl
@@ -1,0 +1,62 @@
+layout(points) in;
+layout(triangle_strip, max_vertices = 4) out;
+
+uniform mat4 m;
+uniform mat3 viewMat;
+
+uniform float size;
+uniform float brightness;
+uniform float minimumFeatureSize;
+
+in Vertex
+{
+    vec3  color;
+    float size;
+    float brightness;
+} vertex[];
+
+out vec4 v_Color;
+out vec2 v_TexCoord;
+
+void main()
+{
+    float s = size * vertex[0].size;
+    if (s >= minimumFeatureSize)
+    {
+        vec4 p = m * gl_in[0].gl_Position;
+        float screenFrac = s / length(p);
+        if (screenFrac < 0.1)
+        {
+            /*
+             * This shader assumes that vertices are rendered in CCW order.
+             */
+            vec4 v0 = vec4(viewMat * vec3(-1.0,  1.0, 0.0) * s, 0.0);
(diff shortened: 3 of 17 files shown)


levinli303:
unable to test and verify on Mac because I would need to use 3.3 core profile, which throws
`"ERROR: 0:1: '' :  version '120' is not supported\nERROR: 0:2: '' :  #version required and missing.\nERROR: 0:33: 'attribute' : syntax error: syntax error\n"` for version 120 shaders.

on Android, however, I'm getting `"Internal compiler error: Error: array indexing out of boundary\nAssertion failed: me->getSize() == 1"`

375gnu:
@levinli303 could you test now?

levinli303:
@375gnu  I got it working after moving

```
layout(points) in;
layout(triangle_strip, max_vertices = 4) out;
```

to right after `precision highp float;` not sure if OpenGL spec has requirement for it.

375gnu:
@levinli303 please check now

levinli303:
Some weird behaviors: galaxy disappeared when it is zoomed out, when going back to sun, there is also a period time where the rendering is wrong.
```
https://user-images.githubusercontent.com/17924287/218458571-9e0e094e-22bf-4c83-9e38-0e7da6810e57.mp4

**Comment by 375gnu** [comment]
could you capture gl calls?

**Comment by 375gnu** [comment]
I'm able to reproduce the first issue, so only for the 2nd one capture is required.

**Comment by levinli303** [comment]
I'll see if I could capture trace with render doc later. the Android GPU inspector requires ANGLE which only goes up to ES 3.1

**Comment by levinli303** [comment]
captured on renderdoc

[issue.zip](https://github.com/CelestiaProject/Celestia/files/10742838/issue.zip)

**Comment by 375gnu** [comment]
@levinli303 thanks but i don't see Sun rendering in the capture. I see Milky way (2), globulars (3-73), stars (74-77), solar system bodies (78-80) and finally HUD.

**Comment by levinli303** [comment]
[time2.zip](https://github.com/CelestiaProject/Celestia/files/10743956/time2.zip)

a smaller frame capture where i disabled other thiings.

**Comment by 375gnu** [comment]
still no Sun rendering. Interesting.

**Comment by 375gnu** [comment]
`EGL_DEPTH_SIZE, 16,` - this should be 24 at least

**Comment by 375gnu** [comment]
@levinli303 could provide an apk? i'll try to test on my phone.

**Comment by levinli303** [comment]
> @levinli303 could provide an apk? i'll try to test on my phone.

I created a debug build here, install with `adb install -t xx.apk`
https://1drv.ms/u/s!AgSl7533DrJfimWyQQiUC5l-ELKj?e=4Qvzwa

**Comment by 375gnu** [comment]
on my phone (mali-g52 mc2) it's not reproducible

**Comment by 375gnu** [comment]
But your second video doesn't have the Galaxies behind the Sun. Are you sure that the issue is only with geometric shaders?

**Comment by levinli303** [comment]
if I return false in hasGeomShader, it renders without issue. I suspect it is an issue in the driver. mine is adreno 620.

**Comment by 375gnu** [comment]
@levinli303 what about such workaround?

**Comment by levinli303** [comment]
I think we should have define and preprocessor in code to control whether the geometry path can be used at all. I can't find a way to force GL ES 2.0 on EGL and it's likely some other GPU might have this issue as well, so I'd prefer to disable it on Android until we find the very root cause.

I'll try to see if I can debug this issue on my end when I have time. When I replayed the frame using renderdoc, the artifacts are no where to be seen but if I use the screenshot function in Celestia, the output image has the artifact, given it reads from the frame buffer so I suppose the artifacts are indeed rendered, it could still be somewhere in the code.

**Review by levinli303 — APPROVED** [review]

**Comment by levinli303** [comment]
attaching another video that you can see the shapes of the artifact, does it come from the galaxy templates?

https://user-images.githubusercontent.com/17924287/219869485-eaa8e75f-375f-4bf5-9773-2fc2583fc421.mp4

**Comment by 375gnu** [comment]
they could be. as in the old code we skip blobs with size > 0.1 screen. perhaps your gpu emits something with random parameters when no actual output is done. could you check the updated shader?

**Comment by levinli303** [comment]
it's not related to the recent push, but now it just straight up hangs the GPU

```
2023-02-19 11:20:08.406 22281-22370 Adreno-GSL              space.celestia.mobilecelestia        W  <gsl_ldd_control:553>: ioctl fd 91 code 0xc040094a 
(IOCTL_KGSL_GPU_COMMAND) failed: errno 35 Resource deadlock would occur
2023-02-19 11:20:08.406 22281-22370 Adreno-GSL              space.celestia.mobilecelestia        W  <log_gpu_snapshot:462>: panel.gpuSnapshotPath is not set.not 
generating user snapshot
2023-02-19 11:20:08.406 22281-22370 Adreno-GSL              space.celestia.mobilecelestia        W  <gsl_ldd_control:553>: ioctl fd 91 code 0xc040094a 
(IOCTL_KGSL_GPU_COMMAND) failed: errno 35 Resource deadlock would occur
2023-02-19 11:20:08.406 22281-22370 Adreno-GSL              space.celestia.mobilecelestia        W  <log_gpu_snapshot:462>: panel.gpuSnapshotPath is not set.not 
generating user snapshot
```

375gnu:
Let's disable them by default on phones.

PR #1538, opened 2023-02-04
Fix bug with sampled orbits introduced by 90bd582

375gnu:
Reported in https://celestia.space/forum/viewtopic.php?p=160875#p160875


ajtribick (approved):

ajtribick:
Yep, missed the y/z swap there.

PR #1543, opened 2023-02-05
Preparation for geometry shaders

375gnu:
Currently only core flavor so require GL3.2 (There is an extension exists for older GL versions but it's incompatible with the core counterpart). We could potentially lower the required version later but afaik all desktop videocards capable to run geometry shaders support GL3.2. I'm not aware about mobile chips so set version to minimally required, anyway decision whether to enable geometry shaders will be made depending on extension presence.


levinli303:
iOS is stuck on OpenGLES 3.0 forever. For Android, https://developer.android.com/about/dashboards#OpenGL 80% of the active devices support OpenGLES 3.2. I will keep requesting for OpenGLES 2.0 in the Android frontend, as EGL can take care of upgrading GLES version.

375gnu:
This is not a problem. We'll keep software path available.

levinli303 (approved):

PR #1550, opened 2023-02-07
[qt] Pass the --dir command line option to the init method

### LINKED ISSUE #1547: Data loading error when using --dir and --conf commands
**Describe the bug**
1. Data loading is not working correctly when using `--dir` and `--conf` commands (used simultaneously) on Linux Mint 20.3 Cinnamon. Apparently, _.cfg_ file is connected (since there is a reaction to the _LuaHook_ line, see below), but the data is not.

2. _LuaHook_ does not work when loading using the mentioned commands. If you put the data and _.cfg_ file with _LuaHook_ line to /usr/share/celestia (or make symlink there), the data is loading and Lua works.

**To Reproduce**
1. Copy /usr/share/celestia to another place (don't forget to replace symlink to _celestia.cfg_ with the file itself).
2. Install [LUT](https://celestia.space/lua-universal-tools.html) to this copy following the instructions.
3. Run Celestia with this external data: `celestia --dir "/path/to/your/folder" --conf "/path/to/your/folder/celestia.cfg"`
4. See error.

**Expected behavior**
Loading without errors like it takes place with mentioned commands on Windows.

**Screenshots**
Text: "Error opening LuaHook luahookinit.lua".
luahook

**Desktop:**
 - OS: Linux Mint 20.3 Cinnamon.
 - Frontend: Qt.
 - Version: 1.7.0.

ajtribick:
Resolves #1547

### DIFF
--- a/src/celestia/qt/qtmain.cpp
+++ b/src/celestia/qt/qtmain.cpp
@@ -16,23 +16,24 @@
  ***************************************************************************/
 
 
-#include <QTime>
-#include <QApplication>
-#include <QSplashScreen>
-#include <QDesktopServices>
-#include <QDir>
-#include <QPixmap>
-#include <QBitmap>
-#include "qtgettext.h"
-#include <QLocale>
-#include <QLibraryInfo>
-#include <vector>
-#include "qtappwin.h"
+#include <cstdio>
+#include <iostream>
+#include <string>
+
 #include <fmt/printf.h>
 
-using namespace std;
+#include <QApplication>
+#include <QBitmap>
+#include <QDesktopServices>
+#include <QDir>
+#include <QLibraryInfo>
+#include <QLocale>
+#include <QPixmap>
+#include <QSplashScreen>
+#include <QTime>
 
-//static const char *description = "Celestia";
+#include "qtgettext.h"
+#include "qtappwin.h"
 
 // Command line options
 static bool startFullscreen = false;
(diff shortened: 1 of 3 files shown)


src/celestia/qt/qtmain.cpp
```diff
@@ -16,33 +16,34 @@
...
 // Command line options
 static bool startFullscreen = false;
 static bool runOnce = false;
-static QString startURL;
-static QString logFilename;
-static QString startDirectory;
-static QString startScript;
-static QStringList extrasDirectories;
-static QString configFileName;
+static QString startURL{ };
```
375gnu, line 41: why to add these {}? static variables are initialized by default.

ajtribick: because honestly I can't remember the rules, so I assume anything not initialized is probably undefined behavior.

ajtribick: And furthermore, what we had before is technically zero-initializing the QString, which depending on the internals of the Qt framework may or may not be the same as default-initializing a QString. I prefer to err on the side of caution.

ajtribick: I guess we might also ask why bother initializing the static booleans to false when they will be zero-initialized?

ajtribick:
Actually on second thoughts, let's do this properly with QCommandLineParser

PR #1552, opened 2023-02-08
[qt] Rewrite command line parser

### LINKED ISSUE #1547: Data loading error when using --dir and --conf commands

ajtribick:
Rewrite the command line parser to use `QCommandLineParser`

Wire up some properties that were parsed but not used:

* `--dir` option: fixes #1547
* `-u`/`--url`
* `-s`/`--nosplash`
* `--fullscreen`

Add `--help` and `--version` options.

The `--once` option is no longer parsed: implementing this cross-platform is somewhat non-trivial, if we need it we can add it later. This doesn't actually remove any functionality from the Qt front-end as the value wasn't used.

### DIFF
--- a/src/celestia/qt/CMakeLists.txt
+++ b/src/celestia/qt/CMakeLists.txt
@@ -22,6 +22,7 @@ set(QT_SOURCES
   qtcelestialbrowser.cpp
   qtcelestiaactions.cpp
   qtcolorswatchwidget.cpp
+  qtcommandline.cpp
   qtdeepskybrowser.cpp
   qtdraghandler.cpp
   qteventfinder.cpp
@@ -43,6 +44,7 @@ set(QT_HEADERS
   qtcelestialbrowser.h
   qtcelestiaactions.h
   qtcolorswatchwidget.h
+  qtcommandline.h
   qtdeepskybrowser.h
   qtdraghandler.h
   qteventfinder.h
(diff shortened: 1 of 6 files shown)


375gnu:
some extra files have been committed

ajtribick:
fixed

375gnu (approved):

PR #1553, opened 2023-02-09
render.cpp Tinted Illumination Update

JiliTheSpaceboy:
2022-02-08 - The hard coded tinted illumination now uses the D58 white point blackbody color provided in [this issue](https://github.com/CelestiaProject/Celestia/issues/892). Draper point is also implemented - illumination for temperatures below 800 K is set to (0f, 0f, 0f).


SevenSpheres:
Unfortunately this isn't that simple, since the classic star colors still exist, so the classic tint colors should still be used when classic star colors are enabled.

JiliTheSpaceboy:
> Unfortunately this isn't that simple, since the classic star colors still exist, so the classic tint colors should still be used when classic star colors are enabled. (I wouldn't be against _removing_ the classic star colors, but that would be a different discussion.)

Yes, but this isn't really worse than the current colors. I hope we can make it tinted illumination takes colors from the same place as star colors, but for now this will do.

JiliTheSpaceboy:
A proper tinted illumination color lookup has been implemented by pelagornithid, so once that is incorporated into the master branch, this one can be discarded.

375gnu:
Who is pelagornithid and where is their implementation?

SevenSpheres:
pelagornithid = @ajtribick

ajtribick:
We did a live session playing around with the tinted illumination implementation. Using the blackbody colors directly produces effects that are pretty extreme for red stars, and using such a long series of `else if` blocks actually breaks MSVC.

I have some code to do a less saturated version but still need to hook it into the preferences system.

375gnu:
The obvious way is to pregenerate a texture with required colors and read it from shaders.

SevenSpheres:
superseded by #1585

PR #1554, opened 2023-02-09
Disable use of structs in shaders by default

levinli303:
I think it is better to disable it by default for compatibility

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,12 +64,18 @@ option(ENABLE_GLES          "Build for OpenGL ES 2.0 instead of OpenGL 2.1 (Defa
 option(USE_GTKGLEXT         "Use libgtkglext1 for GTK2 frontend (Default: on)" ON)
 option(USE_QT6              "Use Qt6 in Qt frontend (Default: off)" OFF)
 option(USE_GTK3             "Use Gtk3 in GTK2 frontend (Default: off)" OFF)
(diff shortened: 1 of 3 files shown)


375gnu:
@levinli303 Could you add another commit which adds this definition via cmake?

375gnu (approved):

CMakeLists.txt
```diff
@@ -64,14 +64,20 @@ option(ENABLE_GLES          "Build for OpenGL ES 2.0 instead of OpenGL 2.1 (Defa
 option(USE_GTKGLEXT         "Use libgtkglext1 for GTK2 frontend (Default: on)" ON)
 option(USE_QT6              "Use Qt6 in Qt frontend (Default: off)" OFF)
 option(USE_GTK3             "Use Gtk3 in GTK2 frontend (Default: off)" OFF)
-option(USE_WAYLAND          "Use Wayland in Qt frontend (Default: off)" OFF)
```
munix9, line 67: USE_WAYLAND option is gone again?

levinli303: oops, I think I’ve deleted this line by accident.

levinli303: I don’t have access to a PC now, so if anyone can help add it back, feel free to do so.

375gnu: oops! :) i returned it back

PR #1559, opened 2023-02-10
Use AppStream-conformant application ID

yselkowitz:
The [AppStream spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic) requires, and Flatpak enforces, a minimum three part application ID, and discourages capital letters and hyphens.

/cc @mattiaverga


375gnu (approved):

375gnu:
thanks for contribution!

PR #1564, opened 2023-02-12
Preserve ShowHorizonGrid and ShowEcliptic in cel URLs

ajtribick:
Using 4₁₆ (0100₂) as the leading digit leads to masking out of the ShowHorizonGrid and ShowEcliptic flags. The leading digit should be 7₁₆ (0111₂) instead.


375gnu:
Are they preserved in 1.6?

ajtribick:
Yes, 1.6.x just stores the render flags directly (split between `rf` and `nrf` is only done in 1.7): https://github.com/CelestiaProject/Celestia/blob/b85ed7aa94145ca9098047f8823769850a729e85/src/celestia/url.cpp#L386-L388

375gnu (approved):

PR #1570, opened 2023-02-13
[qt] Use spinboxes for feature size and ambient light level numeric input

ajtribick:
The feature size numeric input is extremely awkward to use. This replaces it with a spinbox, which does numeric input a lot more nicely than a masked line edit.

Also added a spinbox for ambient light level

### DIFF
--- a/src/celestia/qt/preferences.ui
+++ b/src/celestia/qt/preferences.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>456</width>
-    <height>409</height>
+    <width>479</width>
+    <height>505</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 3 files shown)


375gnu (approved):

375gnu (approved):
could you address sonar warnings? they look reasonable.

375gnu (changes requested):
.

ajtribick:
Using `const` feels somewhat wrong given that these methods are doing updates, although strictly speaking the output of the UI compiler does just expose raw mutable pointers so it would be possible.

As for the "identical implementations", this is something of a necessity given the convention-based hookup of the event handlers, that both need to update the same property on the app core.

ajtribick:
Also I don't want to refactor the entire preferences dialog right now

375gnu (approved):

PR #1574, opened 2023-02-16
Use constexpr, enum class, array_view in starcolors

ajtribick:


src/celestia/qt/qtglwidget.cpp
```diff
@@ -45,19 +45,24 @@
 using namespace Qt;
 
 
-const int DEFAULT_ORBIT_MASK = Body::Planet | Body::Moon | Body::Stellar;
+namespace
+{
+
+constexpr int DEFAULT_ORBIT_MASK = Body::Planet | Body::Moon | Body::Stellar;
+
+constexpr int DEFAULT_LABEL_MODE = 2176;
```
375gnu, line 53: please take care of this magic constant

375gnu (changes requested):

src/celestia/qt/qtglwidget.cpp
ajtribick: done

375gnu (approved):

PR #1575, opened 2023-02-17
only perform isdigit check on non-negative values

levinli303:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/isdigit-iswdigit-isdigit-l-iswdigit-l?view=msvc-170#return-value

required for MSVC in debug

### DIFF
--- a/src/celutil/greek.cpp
+++ b/src/celutil/greek.cpp
@@ -127,7 +127,7 @@ getFirstWordLength(std::string_view str)
         sp = str.length();
 
     // skip digits
-    while (sp > 0 && std::isdigit(str[sp - 1]) != 0)
+    while (sp > 0 && std::isdigit(static_cast<unsigned char>(str[sp - 1])) != 0)
         sp--;
 
     return sp;


src/celutil/greek.cpp
```diff
@@ -127,7 +127,7 @@ getFirstWordLength(std::string_view str)
         sp = str.length();
 
     // skip digits
-    while (sp > 0 && std::isdigit(str[sp - 1]) != 0)
+    while (sp > 0 && str[sp - 1] >= 0 && std::isdigit(str[sp - 1]) != 0)
```
375gnu, line 130: `while (sp > 0 && std::isdigit(static_cast<unsigned char>(str[sp - 1])) != 0)`

375gnu (changes requested):

src/celutil/greek.cpp
levinli303: fixed

375gnu (approved):

ajtribick:
I believe the recommended fix here is to cast the argument to `unsigned char`, e.g. https://en.cppreference.com/w/cpp/string/byte/isdigit

> Like all other functions from [`<cctype`>](https://en.cppreference.com/w/cpp/header/cctype), the behavior of `std::isdigit` is undefined if the argument's value is neither representable as `unsigned char` nor equal to EOF. To use these functions safely with plain `char`s (or `signed char`s), the argument should first be converted to `unsigned char`:

>
> ```cpp
> bool my_isdigit(char ch)
> {
>     return std::isdigit(static_cast<unsigned char>(ch));
> }
> ```

levinli303:
> I believe the recommended fix here is to cast the argument to `unsigned char`, e.g. https://en.cppreference.com/w/cpp/string/byte/isdigit

yeah I applied the fix, per 375gnu in the comment but did not update the PR title.

PR #1578, opened 2023-02-18
Use R128 instead of BigFix

375gnu:
r128.h is taken from https://github.com/fahickman/r128

With -O2 and GCC both show the similar performance (R128 is 1-4% faster). But with -O3 and GCC in multiplication testing R128 is ~40% (amd64) or even ~70% (arm64) faster.


ajtribick:
Nice. I have a slight preference towards putting r128.h into the thirdparty folder rather than celutil.

375gnu:
> Nice. I have a slight preference towards putting r128.h into the thirdparty folder rather than celutil.

If we add it there as a submodule then we make burden for packagers because there is no r128 packages nor in any unix distribution neither in vcpkg (e.g. for debian we will have to add another r128.h copy as a patch). If we add it as plain file then we we'll have two different ways to manage 3rd party libraries.

ajtribick:
Adding it into the src tree is itself a second way of handling third party libraries, so I'm not particularly sure that's necessarily a major strike against including actual subfolders of the thirdparty directory. I still think putting it in a subfolder of thirdparty (not necessarily a submodule, since I guess that would complicate automated checkouts) and defining it as an INTERFACE target is a cleaner approach.

375gnu:
When added directly to src/ we can treat it as a part of Celestia not a 3rd party dependency. But let's wait for @levinli303 opinion as well.

levinli303:
I'm ok with either. but I think I'll need to check compiling against Android/iOS first.

ajtribick:
Probably worth checking with @munix9 regarding the impact of putting stuff in the thirdparty directory on the packaging process.

ajtribick (approved):

375gnu:
@levinli303 have you had possibility to test this against android & ios?

levinli303 (approved):

levinli303:
Appears to be working fine.

PR #1581, opened 2023-02-18
Update Debian installation instruction

jkirk:
Debian 11 (bullseye) is supported a while now, but the installation instruction did not mention it.

While at it, also tell the user what and why she actually does it and as the given public key expired recently, also tell the user what to do, if that happens.

(Celebrating 2 years Mars Perseverance Rover landing: https://mars.nasa.gov/news/8865/touchdown-nasas-mars-perseverance-rover-safely-lands-on-red-planet/)


375gnu (approved):

375gnu:
thanks!

PR #1582, opened 2023-02-18
Always use gpu memory to render models and rectangles

375gnu:


src/celengine/rendcontext.cpp
```diff
@@ -282,10 +282,10 @@ RenderContext::drawGroup(const cmod::PrimitiveGroup& group)
         glActiveTexture(GL_TEXTURE0);
     }
 
-    glDrawElements(GLPrimitiveModes[(int)group.prim],
-                   group.indices.size(),
+    glDrawElements(GLPrimitiveModes[static_cast<int>(group.prim)],
+                   group.indicesCount,
                    GL_UNSIGNED_INT,
-                   group.indices.data());
+                   reinterpret_cast<void*>(group.indicesOffset*sizeof(GLuint)));
```
ajtribick, line 288: I think we should add `//NOSONAR` to these `reinterpret_cast`s to reduce the number of false positives.

levinli303 (approved):

ajtribick (approved):

PR #1585, opened 2023-02-19
Tinted illumination

### LINKED ISSUE #1579: URL parser should treat TintedIllumination as a new render flag
The URL parser code treats `TintedIllumination` as the last render flag supported by 1.6.x, however this setting is not in the 1.6 codebase: the last 1.6.x setting is actually `ShowEcliptic`. The consequence is that any cel URL generated in 1.6.x will switch the `TintedIllumination` flag off in 1.7.x. Probably `TintedIllumination` should be handled in the `nrf` query parameter, and presumably we need to avoid breaking any URLs that have already been generated by people using the 1.7.x prereleases.

ajtribick:
Updated tinted illumination functionality, ideas hashed out in live development session.

* Removes the `TintedIllumination` render flag, updates URLs (currently shifts all the other ones, but could also leave `nrf` unchanged and a hole in the enum). Fixes #1579
* Generation of blackbody tables is moved into the code, using a method consistent with the existing tables.
* If "Enhanced" star colors are selected, the legacy tinted illumination is used, with no change of saturation. This keeps things consistent for users of non-Qt front-ends, where the tinted illumination saturation slider is not implemented.
* If "Blackbody D65" star colors are selected, use a variant set of blackbody colors generated with the white point set to a 5700 K blackbody. This ensures that the Solar System planets do not get affected by tinted illumination. Light sources with temperatures <1000 K get faded to black linearly with temperature.
* Add a "tinted illumination saturation" slider to the Qt preferences dialog box and store in the settings.
* Add `celestia:gettintsaturation` and `celestia:settintsaturation` commands for Lua scripts


src/celengine/starcolors.h
```diff
@@ -32,15 +40,26 @@ class ColorTemperatureTable
             return colors[colorTableIndex];
     }
 
+    Color lookupTintColor(float temp, float saturation) const
+    {
+        Eigen::Vector3f color = lookupColor(temp).toVector3();
+        color = (1.0f - saturation) * Eigen::Vector3f::Ones() + saturation * color;
```
375gnu, line 46:
```
color = vecgl::mix(Eigen::Vector3f::Ones(), color, saturation);
```

ajtribick: annoyingly that doesn't work at the moment because `Eigen::Vector3f::Ones()` actually returns `Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<float>, Eigen::Vector3f>` rather than `Eigen::Vector3f`, so the template matching fails

ajtribick: fixed... just needed a bit more template magic

ajtribick:
The Cppcheck error [appears to have been fixed as a false-positive](https://github.com/danmar/cppcheck/pull/3950), but Codacy doesn't seem to have updated their Cppcheck version since 2020.

src/celengine/render.h
```diff
@@ -208,16 +208,15 @@ class Renderer
         ShowEclipticGrid        = 0x0000000000800000,
         ShowHorizonGrid         = 0x0000000001000000,
         ShowEcliptic            = 0x0000000002000000,
-        ShowTintedIllumination  = 0x0000000004000000,
```
375gnu, line 211: maybe better to leave a hole and use it later for any new flag?

src/celengine/vecgl.h
```diff
@@ -131,11 +131,14 @@ translate(T x, T y, T z)
...
 
-template <typename T, typename F, std::enable_if_t<std::is_floating_point_v<F>, bool> = true>
-inline T
-mix(const T &x, const T &y, F a)
-{
-    return x * (static_cast<F>(1.0) - a) + y * a;
+template <typename T,
+          typename U,
+          std::enable_if_t<std::is_same_v<typename T::Scalar, typename U::Scalar> &&
+                           std::is_floating_point_v<typename T::Scalar>, bool> = true>
```
375gnu, line 137: i'm afraid your change allows to pass Vector3f and Vector4f or Matrix*f and Vector*f together

ajtribick: True but you can't add those to each other, so the compilation would fail anyway.

src/celengine/render.h
ajtribick: Sounds reasonable, will update this.

375gnu:
What object can we test this?

ajtribick:
On thinking about this some more, I suspect this approach to fading out light sources will not play nicely with multiple stars, and want to try some Draper-point related stuff. Setting to draft for now but should have some updates in the near future.

ajtribick:
Report that `nBodies` isn't used in url.cpp is a false positive

ajtribick:
Updated the PR to suppress illumination below 780 K, to allow for slight lighting from objects at the Draper point.

gliese229_test

Test case for a satellite at 2 Jupiter radii from Gliese 229 B: most of the illumination comes from Gliese 229 A, but there's still a bit of a reddish glow from Gliese 229 B.

375gnu (approved):

ajtribick:
Squashed commits

PR #1589, opened 2023-02-19
Add a bound state to VertexObject

### LINKED ISSUE #1588: Starting up with Star rendering disabled crashes Celestia
Screenshot 2023-02-19 at 22 44 58

It crashed because VAO is not initialized (the first time entry `initialized` should false, but it is true here):
Screenshot 2023-02-19 at 22 47 02

levinli303:
And only unbinds if it is bound...

Fixes #1588

### DIFF
--- a/src/celrender/vertexobject.cpp
+++ b/src/celrender/vertexobject.cpp
@@ -75,6 +75,7 @@ VertexObject::~VertexObject()
 
 void VertexObject::bind() noexcept
 {
+    m_state |= State::Bound;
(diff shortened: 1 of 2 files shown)


ajtribick (approved):

375gnu (approved):
this is not an ideal fix, but i'm in mood of removing this crappy class and replacing it with VAO and GPU buffer wrappers like https://magnum.graphics/ does, so let's go with this chnage.

PR #1592, opened 2023-02-20
Fixes and refactorings in globulars rendering

375gnu:
Globulars were broken since I migrated them to more modern rendering code. Now they produce the same image as 1.6 does.

### DIFF
--- a/shaders/globular_vert.glsl
+++ b/shaders/globular_vert.glsl
@@ -1,56 +1,35 @@
 attribute vec3 in_Position;
-attribute vec4 in_Color;
-attribute float starSize;
-attribute float eta;
+attribute vec3 in_TexCoord0; // reuse it for starSize, relStarDensity and colorIndex
 
+uniform sampler2D colorTex;
 uniform mat3 m;
 uniform vec3 offset;
 uniform float brightness;
 uniform float pixelWeight;
-uniform float RRatio;
 uniform float scale;
 
-const float RRatio_min   = pow(10.0, 1.7);
 const float clipDistance = 100.0; // observer distance [ly] from globular, where we
                                   // start "morphing" the star-sprite sizes towards
                                   // their physical values
 
 varying vec4 color;
 
-float relStarDensity(void)
-{
-    /*! As alpha blending weight (relStarDensity) I take the theoretical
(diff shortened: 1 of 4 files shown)


ajtribick:
The appearance close-up is much improved. However I think the view from far away is too bright.

cel://Freeflight/2023-02-21T18:45:58.19478Z?x=wCVrj691pgoC&y=7kdWVw4mrPP//////////w&z=TtjieySSper//////////w&ow=0.044617686&ox=0.20519118&oy=-0.97737825&oz=-0.025250802&select=TYC%206271-1396-1&fov=23.50311&ts=1&ltd=0&p=0&rf=136107799&nrf=255&lm=128&tsrc=0&ver=3

v1.7 with this patch (note for some reason the grab image function results in an image with transparency, the "bullethole" effect isn't present in the application itself)
globular_brightness_1_7

v1.6
globular_brightness_1_6

375gnu:
should be fixed now

ajtribick:
Still seeing rendering with very bright globular clusters.

375gnu:
@same url? could you provide screenshots?

ajtribick:
Result of the same cel URL with commit 4949b64:

globulars_4949b64

375gnu:
@ajtribick do you see the proper image now? I have the same image as 1.6 now, only some globulars are twinkling maybe due to z-fighting, i'll try to fix this.

ajtribick:
Seems better though still a lot brighter than the 1.6 version. The behaviour as you move towards/away from the globular clusters also seems a bit odd. E.g. for Terzan 10, the stellar halo suddenly pops out of existence at ~1560 light years. It then gets dimmer as you move to ~2.9 kpc, then brightens again until it suddenly disappears at ~11.3 kpc.

375gnu:
I don't see any brightness difference (actually it's about 0.1, i haven't figured out why the same code give different result in 1.6 and 1.7):
image

ajtribick:
View from the URL in commit 31e5d39:
globulars_31e5d39

Will try doing some more testing later

375gnu:
@ajtribick should be fixed now

ajtribick (approved):

PR #1594, opened 2023-02-21
[qt] Fix transparent image capture

ajtribick:
Fixes a bug where image captures on Windows Qt were being generated with transparency, by turning off the alpha channel in the QOpenGLWidget's framebuffer.


375gnu:
could you confirm that gl es works fine with such default fb?

ajtribick:
From what I can tell, GLES seems to work just as badly with this patch (dull greyish stars, fails to initialize on Windows unless running in Qt Creator) as without it.

375gnu (approved):
it seems that gles supports default fb without alpha

PR #1598, opened 2023-02-23
Add a flipped user interface for overlay and label annotations

levinli303:
Screenshot 2023-02-23 at 12 55 19

### DIFF
--- a/src/celengine/overlay.h
+++ b/src/celengine/overlay.h
@@ -14,20 +14,15 @@
 #include <vector>
 #include <fmt/printf.h>
 #include <Eigen/Core>
+#include <celengine/textlayout.h>
 
 class Color;
 class Overlay;
 class Renderer;
-class TextureFont;
 
 namespace celestia
 {
 class Rect;
-
-namespace engine
-{
-class TextLayout;
-}
 }
 
 class Overlay
@@ -42,11 +37,13 @@ class Overlay
 
     void setWindowSize(int, int);
     void setFont(const std::shared_ptr<TextureFont>&);
+    void setTextAlignment(celestia::engine::TextLayout::HorizontalAlignment halign);
 
     void setColor(float r, float g, float b, float a);
     void setColor(const Color& c);
 
     void moveBy(float dx, float dy);
+    void moveBy(int dx, int dy);
     void savePos();
     void restorePos();
 
--- a/celestia.cfg
+++ b/celestia.cfg
@@ -428,4 +428,10 @@ StarTextures
 #------------------------------------------------------------------------
 # LeapSecondsFile "/usr/share/zoneinfo/leap-seconds.list"
 
+#------------------------------------------------------------------------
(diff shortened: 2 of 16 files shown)


375gnu:
What do you think about https://github.com/HOST-Oman/libraqm?

levinli303:
@375gnu  if we decide to use ICU, maybe just use the functions in ICU for bidi and shaping?

this is what I got with ICU's u_shapeArabic and ubidi

Screenshot 2023-02-23 at 17 32 32

375gnu:
If we can icu only to achieve the same, then i'm ok. actually icu is indirect dependency for qt frontend.

src/celengine/overlay.h
```diff
@@ -77,14 +79,15 @@ class Overlay
...
     float yoffset{ 0.0f };
+    TextureFont::Alignment textAlignment{ TextureFont::Alignment::Left };
 
     Renderer& renderer;
 
     struct CursorPosition
     {
         void reset()
         {
-            x = y = 0.125f;
```
375gnu, line 87: this was set intentionally to prevent jittering. i don't understand completely how it works but it was in 1.6 and with positions set to 0 we had labels jittering.

levinli303: I only found a note here https://github.com/CelestiaProject/Celestia/blob/b0f40cec084d85ce767e35d9f2d344fe12b72be1/src/celengine/render.cpp#L108 (which I also deleted). I did not see anything abnormal removing this.

375gnu:
```
.../Celestia/src/celengine/render.h:398:43: warning: ‘Renderer::Annotation::halign’ is too small to hold all values of ‘enum class Renderer::LabelHorizontalAlignment’
         LabelHorizontalAlignment halign : 3;
                                           ^
.../Celestia/src/celengine/render.h:399:41: warning: ‘Renderer::Annotation::valign’ is too small to hold all values of ‘enum class Renderer::LabelVerticalAlignment’
         LabelVerticalAlignment valign : 3;
```

375gnu:
image
Still extra symbols are printed.

375gnu (changes requested):
see comments

levinli303:
> image Still extra symbols are printed.

should have been fixed after rebasing.

levinli303:
> ```
> .../Celestia/src/celengine/render.h:398:43: warning: ‘Renderer::Annotation::halign’ is too small to hold all values of ‘enum class Renderer::LabelHorizontalAlignment’
>          LabelHorizontalAlignment halign : 3;
>                                            ^
> .../Celestia/src/celengine/render.h:399:41: warning: ‘Renderer::Annotation::valign’ is too small to hold all values of ‘enum class Renderer::LabelVerticalAlignment’
>          LabelVerticalAlignment valign : 3;
> ```

fixed.

375gnu (approved):

PR #1600, opened 2023-02-24
Keep xoffset and yoffset in overlay

### LINKED ISSUE #1599: overlapping time information
**Describe the bug**
Time information fields overlap each other (red "paused" is written above time speed (realtim, 2x slower, etc) 

**To Reproduce**
Pause time.

**Expected behavior**
"paused" should written after the speed.

**Screenshots**
image

This is caused by https://github.com/CelestiaProject/Celestia/pull/1595

levinli303:
Fix #1599 

Add a lineStart property so we can support

```
....... text text
continued text
```

### DIFF
--- a/src/celengine/overlay.cpp
+++ b/src/celengine/overlay.cpp
@@ -102,10 +102,12 @@ void Overlay::print_impl(const std::string& s)
         fontChanged = false;
     }
 
-    restorePos();
-    auto [_, y] = font->render(s, global.x + xoffset, global.y + yoffset);
-    global.y = y;
-    savePos();
+    auto [x, y] = font->render(s, global.x + xoffset, global.y + yoffset);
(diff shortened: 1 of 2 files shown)


375gnu:
@levinli303 I think we should discuss first what API level TextureFont should provide:
 * if it's low level then TextureFont::print should simply output a provided string at provided coordinates x and y
 * if it's high level then it should care about EOL, start line position and so on.

My proposal is to keep it as low level as possible and to manage higher level things create an additional middleware class.

levinli303:
@375gnu  I agree that we should keep this simple.

As for this PR, I think lineStart is a parameter that we can remove.

levinli303:
@375gnu I plan to add text layout class in another PR to handle line breaks, line start, alignment etc. This PR should be good now

375gnu (approved):

Issue #1551, opened 2023-02-08
Alpha values in render result


levinli303:
To solve #675, I set background color of GL view to black, but it should still be fixed in Celestia. given that we have cleared color to black before rendering, something wrong should be in the code.

A test screenshot with white background
<img width="1450" alt="Screenshot 2023-02-08 at 23 35 20" src="https://user-images.githubusercontent.com/17924287/217578331-96e529ed-e59c-4d91-8704-a2296ffd530c.png">

image captured from appCore.saveScreenShot (I edited the code to keep the alpha bits)
![Celestia Screenshot](https://user-images.githubusercontent.com/17924287/217578039-a6d3f3ac-2cb2-4459-8058-d4b784beb90c.png)

ajtribick:
https://github.com/CelestiaProject/Celestia/blob/15df4ba74b3ae5364cc470c262da4cbd6c1ebc9c/src/celengine/shadermanager.cpp#L2648-L2653

It looks a bit odd that for the case where specular is in the diffuse alpha, the sampled alpha texture still gets used in the diffuse part. Does it help if you change line 2649 to

```cpp
source += "gl_FragColor = vec4(color.rgb, 1.0) * diff + float(color.a) * spec;\n";
```

ajtribick:
can we close this one in favour of #1601 given that most of the discussion has ended up there?

PR #1605, opened 2023-02-25
Don't use alpha as both alpha and specular

ajtribick:
Potential fix for some of the transparency issues, e.g. the Earth continents issue in the screenshots in #1551


375gnu (approved):
it looks reasonable, so let' merge and see

Issue #1590, opened 2023-02-19
[qt] [Linux] Custom splash doesn't load


Lepestronik:
**Describe the bug**
After https://github.com/CelestiaProject/Celestia/pull/1552 custom spalsh stopped loading via `--dir` option on Linux, the original one is used. At the same time, custom splash works on Windows.

**Desktop:**
 - OS: Linux Mint 20.3 Cinnamon.
 - Frontend: Qt.
 - Version: 1.7.0.

PR #1609, opened 2023-02-28
Read splash from a custom directory

375gnu:
... and fallback to the default one if the custom doesn't exist

Closes: #1590


375gnu:
only qt & gtk were tested

ajtribick (approved):

Issue #1614, opened 2023-03-02
Loading some 3DS models crashes Celestia


Lepestronik:
**Describe the bug**
Some 3DS models crashes Celestia when trying to load.

**To Reproduce**
1. Unpack [helios.zip](https://github.com/CelestiaProject/Celestia/files/10870227/helios.zip) to _extras_ directory.
2. Run Celestia.
3. Set time to 1979 and go to "Helios 1" or "Helios 2".
4. See crashing and dump writing to `/var/lib/systemd/coredump`.

**Desktop:**
 - OS: Linux.
 - Frontend: Qt, GTK.
 - Version: 1.7.0.

**Additional context**
3DS models from CelestiaContent loads correctly.

375gnu:
The model itself is bad.
```
gdb build/src/tools/cmod/cmodfix/cmodfix
...
(gdb) run extras/helios/models/helios.3ds
Starting program: /home/itrexgroup/my/Celestia/build/src/tools/cmod/cmodfix/cmodfix extras/helios/models/helios.3ds
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
cmod::(anonymous namespace)::openModel (getHandle=..., in=...) at /home/itrexgroup/my/Celestia/src/celmodel/modelfile.cpp:2032
2032	        celutil::GetLogger()->error("Model file has invalid header.\n");
(gdb) bt
#0  cmod::(anonymous namespace)::openModel (getHandle=..., in=...) at /home/itrexgroup/my/Celestia/src/celmodel/modelfile.cpp:2032
#1  cmod::LoadModel(std::istream&, std::function<int (std::experimental::filesystem::v1::__cxx11::path const&)>) (in=..., handleGetter=...) at /home/itrexgroup/my/Celestia/src/celmodel/modelfile.cpp:2044
#2  0x0000555555556807 in main (argc=<optimized out>, argv=<optimized out>) at /home/itrexgroup/my/Celestia/src/tools/cmod/cmodfix/cmodfix.cpp:166
(gdb) quit
```

Lepestronik:
"3DS to CMOD converter for Windows" creates a working CMOD model. Can this be considered to eliminate **_the root_** of the problem?

![image](https://user-images.githubusercontent.com/89343875/222502342-3a30d343-0b31-4375-9b42-512acdd53433.png)

PR #1615, opened 2023-03-02
Workaround for cases when mesh has 0 primitive groups

375gnu:
-This is not a proper fix but as Celestia loads bad models it should not crash.-

Up: now it should be treated as a fix. @ajtribick could you review again?

Closes: #1614

### DIFF
--- a/src/celengine/meshmanager.cpp
+++ b/src/celengine/meshmanager.cpp
@@ -100,6 +100,10 @@ ConvertTriangleMesh(const M3DTriangleMesh& mesh,
     int nVertices  = mesh.getVertexCount();
     int nTexCoords = mesh.getTexCoordCount();
 
+    // Some community addons ships buggy meshes.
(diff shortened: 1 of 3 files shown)


ajtribick (approved):

src/celengine/meshmanager.cpp
```diff
@@ -96,6 +96,10 @@ cmod::Mesh
 ConvertTriangleMesh(const M3DTriangleMesh& mesh,
                     const M3DScene& scene)
 {
+    // Some community addons ships buggy meshes.
+    if (mesh.getMeshMaterialGroupCount() == 0)
```
ajtribick, line 100: Would also make sense to abort processing in the case where `nFaces` or `nVertices` are zero

levinli303:
i'm seeing crash in this line

https://github.com/CelestiaProject/Celestia/blob/a46a12ca1f177e76950e091911e89ea2a701b03c/src/celmodel/model.cpp#L342

should we also add the check for this line.

Issue #1601, opened 2023-02-24
Transparency render bug


Askaniy:
**Screenshots**
![изображение](https://user-images.githubusercontent.com/44873271/220948414-7a662f89-cd48-4426-8405-7514106e1362.png)

**Desktop**
 - OS: Linux (openSUSE Tumbleweed + KDE 5.27 + Wayland)
 - Frontend: Qt 5
 - Version: 1.7.0 b0f40cec084d85ce767e35d9f2d344fe12b72be1

levinli303:
probably same issues as #1551

375gnu:
if you revert  #1594 does it help?
```
git revert 69be12027e30d60d7861f6c6954cb15e4d1c4a01
```

Askaniy:
Can't revert it easily, I use compiled versions provided by [this repository](https://download.opensuse.org/repositories/home:/munix9:/unstable/openSUSE_Tumbleweed/).

munix9:
I will try to test it this weekend, but my hardware is currently a bit unstable.

Have you tested the AppImage? It is basically the same code base, but compiled under openSUSE Leap 15.4 - maybe you get a different result.

Askaniy:
I hope I did what you meant. It seems to work there.
![screenshot](https://user-images.githubusercontent.com/44873271/221292576-8066a24f-b984-4886-a0c6-731cb01a0fb2.png)

munix9:
Yes, that's what I meant, thank you.
Now we know that in general it works on openSUSE, but on Tumbleweed apparently there is "currently" a problem.
openSUSE Tumbleweed is a rolling release, so it is quite possible that a new os component/library can cause problems.
Let's see if I get something out this weekend.

munix9:
Ok, I can reproduce it:

openSUSE Tumbleweed 20230223
Celestia 1.7.0~git20230224+15df4ba

In a Plasma-X11 session the display is correct.
In a Plasma-Wayland session the rendering of the clouds is wrong, i.e. without transparency.

This is both with celestia-qt and with celestia-qt6.
With celestia-sdl the display is correct again in Plasma-Wayland.

If I start celestia-qt or celestia-qt6 with
`$ QT_QPA_PLATFORM=xcb celestia-qt`
the display is correct again, because then X11 is used for Celestia in the Wayland session.

For further details/testing, I need more instructions from the pros here on how to proceed.

[...]

PR #1616, opened 2023-03-02
Don't write alpha to destination buffer when blending

ajtribick:
I think this should solve the transparent atmospheres in #1601


Askaniy:
I can do the tests after merging

375gnu:
No we need to test before merging cause this fix may break translucent object.

ajtribick:
Things I have checked: atmospheres (inside atmosphere and outside atmosphere), galaxy blobs, Saturn's rings (from both inside and outside Saturn's atmosphere). With #1594 reverted and this change not applied, the image capture gives transparency, once I apply this change (without re-applying #1594) the transparent areas of the capture disappear. I don't see any noticeable difference in the rendered output with/without this patch.

375gnu (approved):

PR #1621, opened 2023-03-04
Legacy scripts code cleanup

ajtribick:
Refactoring of the legacy (.cel) scripts code. More use of smart pointers, break up some of the larger methods, move everything into the `celestia::scripts` namespace

### DIFF
--- a/src/celscript/legacy/cmdparser.cpp
+++ b/src/celscript/legacy/cmdparser.cpp
@@ -9,930 +9,55 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-#include <algorithm>
-#include <iostream>
-#include <sstream>
-#include <string_view>
-#include <Eigen/Geometry>
-#include <celutil/logger.h>
-#include <celmath/mathlib.h>
-#include <celengine/astro.h>
-#include <celengine/parser.h>
-#include <celutil/stringutils.h>
-#include <celutil/tokenizer.h>
-#include <celestia/audiosession.h>
 #include "cmdparser.h"
 
-using namespace std;
-using namespace celmath;
-using namespace celestia;
-using namespace celestia::scripts;
+#ifdef USE_MINIAUDIO
+#include <algorithm>
+#endif
+#include <array>
+#include <cctype>
+#include <cstdint>
+#include <cstdio>
+#include <initializer_list>
+#include <istream>
+#include <map>
+#include <optional>
+#include <sstream>
+#include <variant>
+
+#include <Eigen/Core>
+#include <fmt/format.h>
+
+#include <celengine/hash.h>
+#include <celengine/parser.h>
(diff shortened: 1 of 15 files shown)


src/celscript/legacy/cmdparser.cpp
```diff
@@ -9,1075 +9,1128 @@
...
 
-    if (ttype != Tokenizer::TokenEndGroup)
+    return flags;
+}
+
+
+ObserverFrame::CoordinateSystem parseCoordinateSystem(const std::string& name)
+{
+    using CoordinateSystemMap = std::map<std::string_view, ObserverFrame::CoordinateSystem, CompareIgnoringCasePredicate>;
+    static const CoordinateSystemMap* nameMap =
```
375gnu, line 218: i don't think that static map is a good replacement for the old code especially taking into account the tiny number of parameters.

old code should be faster and it doesn't need additional heap memory, it doesn't raise an issue with cleanup.

ajtribick: What cleanup issue are you referring to here? Destructor doesn't get run since it is leaked (this is just a raw pointer, the `unique_ptr` gets immediately released), so shouldn't do anything to block process exit.

375gnu: i meant exactly that destructor is not running at all. just another issue reported by asan/valgrind. i think we should avoid such cases as much as we can or we'll be buried under asan reports.

ajtribick: Not running destructors on static data and relying on the OS to clean up at process exit is a perfectly valid pattern though. If the tools can't be made aware of that, then they're bad tools.

375gnu: ok, but i still think that std::map should be used here. original code is much easier to understand

ajtribick: Ok will remove it for some of these smaller ones, (the main if...else block that used to be in `parseCommand` I will leave as a map though, because that one was really unwieldy!).

Re: valgrind, maybe possible to set up a suppression file but not sure how portable that would be.

src/celscript/legacy/cmdparser.cpp
```diff
@@ -9,1076 +9,1152 @@
...
 
-    return seq;
+    return flags;
 }
 
 
-auto CommandParser::getErrors() const
-    -> celestia::util::array_view<std::string>
+int
+parseConstellationColor(CommandConstellationColor* cmd, const std::string &s, Eigen::Vector3d *col, int act)
```
375gnu, line 177: const Eigen::Vector3d *col, bool act

375gnu (changes requested):

375gnu (approved):

ajtribick:
squashed commits

PR #1631, opened 2023-03-10
Use glBlendFuncSeparate instead of glBlendFunc in celx scripts

ajtribick:
This would bring setting the blend functions from celx in line with the changes in #1616, but need to figure out a good way of testing it before we merge this.


ajtribick:
Verified this by reverting #1594 and using the Lua Universal Tools. Without this change, LUT causes the transparency bug, with this fix the transparency bug is solved.

375gnu (approved):

PR #1641, opened 2023-03-14
Update lodspheremesh trigonometry calculations

ajtribick:
- Add an ArrayVector class
- Ensure trigonometry arrays have exact values at multiples of 90°
- Use symmetry of sin/cos functions when populating trig arrays
- Use `glBufferSubData` to transfer index data
- Clean up buffers in LODSphereMesh destructor
- Reduce MaxVertexSize since normals just re-use the position data

### DIFF
--- a/src/celengine/lodspheremesh.h
+++ b/src/celengine/lodspheremesh.h
@@ -10,20 +10,34 @@
 
 #pragma once
 
-#include <Eigen/Geometry>
-#include <celengine/texture.h>
+#include <array>
+#include <cstddef>
+#include <vector>
+
+#include <Eigen/Core>
+
 #include <celengine/glsupport.h>
-#include <celmath/frustum.h>
 
+class Texture;
 
-#define MAX_SPHERE_MESH_TEXTURES 6
-#define NUM_SPHERE_VERTEX_BUFFERS 2
+namespace celmath
+{
+class Frustum;
+}
 
 class LODSphereMesh
 {
 public:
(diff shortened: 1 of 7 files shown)


src/celengine/lodspheremesh.h
```diff
@@ -64,20 +98,18 @@ class LODSphereMesh
 
     void renderSection(int phi0, int theta0, int extent, const RenderInfo&);
 
-    float* vertices{ nullptr };
+    std::unique_ptr<VertexArray> vertices{ nullptr };
```
375gnu, line 101: std::vector ?

src/celengine/lodspheremesh.h
```diff
@@ -64,20 +98,18 @@ class LODSphereMesh
...
 
-    int maxVertices{ 0 };
     int vertexSize{ 0 };
 
-    int nIndices{ 0 };
-    unsigned short* indices{ nullptr };
+    std::unique_ptr<IndexArray> indices{ nullptr };
```
375gnu, line 105: std::vector?

375gnu (changes requested):

375gnu (approved):

ajtribick:
To replicate behaviour of the 4 textures `render()` API, we need to remove the `nullptr` elements from the `Texture*` array in a couple of places, so added the `erase` APIs to `ArrayVector`.

375gnu (approved):

PR #1642, opened 2023-03-14
Update dds_decompress to C++17

ajtribick:
Replaces some probably invalid pointer casts with the equivalent `memcpy` calls, uses `constexpr` for the alpha=255 array.


ajtribick:
On second thoughts, I can't be bothered trying to appease SonarScanner on this.

PR #1643, opened 2023-03-15
Apply bidi reorder and arabic shaping

levinli303:


src/celutil/unicode.h
```diff
@@ -14,6 +14,13 @@
 namespace celestia::util
 {
 
-bool UnicodeStringToWString(const icu::UnicodeString &input, std::wstring &output);
+enum ConversionOption : unsigned int
```
levinli303, line 17: @375gnu @ajtribick  what do you suggest we use here. using `enum class` makes it difficult to use bitwise operations.

375gnu:
```
#include <iostream>

enum class flag {
        flag0 = 0,
        flag1 = 1,
        flag2 = 2,
        flag3 = 4
};

flag operator|(flag f1, flag f2) {
        using type = std::underlying_type_t<flag>;
        return flag(type(f1) | type(f2));
}

flag operator&(flag f1, flag f2) {
        using type = std::underlying_type_t<flag>;
        return flag(type(f1) & type(f2));
}

bool is_set(flag f, flag t) {
        using type = std::underlying_type_t<flag>;
        return ((type(f) & type(t)) != 0);
}

int main()
{
        std::cout << int(flag::flag1 | flag::flag2); // 3
        std::cout << int(flag::flag3|flag::flag2); // 6
        std::cout << int((flag::flag3|flag::flag2) & flag::flag1); // 0
        std::cout << int((flag::flag3|flag::flag2) & flag::flag2); // 2
        std::cout << is_set(flag::flag3|flag::flag2, flag::flag1); // 0
        std::cout << is_set(flag::flag3|flag::flag2, flag::flag2); // 1
}

```

375gnu (approved):

PR #1657, opened 2023-03-25
Add reference counting for StarDetails objects

ajtribick:
- Add a new `IntrusivePtr` class for shared ownership with reference count stored in the object. This results in the `IntrusivePtr` being the same size as a raw pointer but at the cost of not supporting weak pointers.
- Use `IntrusivePtr` to manage the ownership of custom `StarDetails` objects.


src/celengine/star.h
```diff
@@ -124,7 +141,7 @@ class StarDetails
 
     std::string infoURL;
 
-    std::unique_ptr<std::vector<Star*>> orbitingStars{ nullptr };
+    std::vector<Star*> orbitingStars{ };
```
375gnu, line 144: std::vector is 32 bytes while pointer is 8 bytes only (in 64 bit systems). i don't think it's an equivalent replacement

375gnu (approved):

PR #1660, opened 2023-03-28
Find first non empty mesh to start with

levinli303:
#1614

### DIFF
--- a/src/celmodel/model.cpp
+++ b/src/celmodel/model.cpp
@@ -339,19 +339,24 @@ Model::sortMeshes(const MeshComparator& comparator)
     std::sort(meshes.begin(), meshes.end(), std::ref(comparator));
 
     std::vector<Mesh> newMeshes;
-    newMeshes.push_back(meshes[0].clone());
-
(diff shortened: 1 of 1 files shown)


src/celmodel/model.cpp
```diff
@@ -339,19 +339,33 @@ Model::sortMeshes(const MeshComparator& comparator)
     std::sort(meshes.begin(), meshes.end(), std::ref(comparator));
 
     std::vector<Mesh> newMeshes;
-    newMeshes.push_back(meshes[0].clone());
+    size_t meshIndex = 0;
+    // Find the first mesh that is not empty to start with
+    while (meshIndex < meshes.size())
```
375gnu, line 344: could you use `for` loop as well?

375gnu (changes requested):

src/celmodel/model.cpp
levinli303: simplified a bit.

375gnu (approved):

PR #1671, opened 2023-04-11
[ci] enable freebsd builds

375gnu:


375gnu:
We need to figure out why ICU tests fail.

PR #1674, opened 2023-04-14
Remove the load fragment script commands

ajtribick:
This does not play nicely with areas of the codebase that assume object pointers remain stable after the load phase is complete, e.g. the category system.

For now I think it's best to remove this functionality (which does not exist yet in any final release). It might make sense to re-add this if we allow script hooks to run during loading, if we want to support updating the databases after loading this will likely need more re-engineering.


375gnu (approved):
it's a pity

PR #1675, opened 2023-04-15
Allow building both Qt5 and Qt6 in same configuration

ajtribick:
Enables both Qt5 and Qt6 front-ends in the same build, except on Mac where thanks to the way homebrew installs things, the Qt6 include paths end up taking priority over the Qt5 ones if both are installed. (Probably this is fixable by someone who actually has a Mac available).

Also updates vcpkg to only do release builds, which should save some time and space.

Differences:
* The `USE_QT6` option is removed, now there is a new `ENABLE_QT6` option which is independent of `ENABLE_QT` (which now only controls whether the Qt5 front-end is built)
* The Qt6 front-end is now called "celestia-qt6" (on Windows "celestia-qt6.exe") and is located in src/celestia/qt6 instead of src/celestia/qt
* The windows artifacts are now named celestia-windows-x86 and celestia-windows-x64


375gnu:
imho it's better to leave `qt` directory's name as is and add a new `qt5`. then once qt5 is dead we can simply remove it and once qt7 is out we will create `qt7` dir the same way. i don't see any problem renaming `celeatia-qt` to `celestia-qt5`.

ajtribick:
Do you think we should rename `ENABLE_QT` to `ENABLE_QT5` in the CMake options also?

375gnu:
yes

375gnu (approved):

ajtribick:
Also updated the debian package definitions and the various desktop files.

375gnu (approved):

PR #1679, opened 2023-04-17
New VBO and VAO wrappers

375gnu:
1. Old VAO+VBO combined wrapper is replaced with new separate ones. They are designed after https://github.com/mosra/magnum which seems to be the best C++ wrapper for OpenGL. If once we decide to create Celestia 2.0 working with higher GL versions Magnum seems to be the best choice to be used instead of home grown code, so let's make our codebase compatible in advance. And old wrappers are just ugly.
2. glmarkers.cpp is another exxample of ugly code so I replaced it with partially generated code.
3. array_view changes are required to work properly with different std::array and std::vector.
4. Old code used to use orphan VBO + glBufferSubData, new one uses orphan VBO + glBufferData because with the old one my GL driver was reporting warnings with `INTEL_DEBUG=perf`. Though I'm not sure that the new code is always faster.


src/celttf/truetypefont.cpp
```diff
@@ -495,8 +495,8 @@ TextureFontPrivate::flush()
     glBufferData(GL_ARRAY_BUFFER, sizeof(FontVertex) * MaxVertices, nullptr, GL_STREAM_DRAW);
     glBufferData(GL_ELEMENT_ARRAY_BUFFER, sizeof(GLushort) * MaxIndices, nullptr, GL_STREAM_DRAW);
 
-    glBufferSubData(GL_ARRAY_BUFFER, 0, sizeof(FontVertex) * m_fontVertices.size(), m_fontVertices.data());
-    glBufferSubData(GL_ELEMENT_ARRAY_BUFFER, 0, sizeof(GLushort) * indexes.size(), indexes.data());
+    glBufferData(GL_ARRAY_BUFFER, sizeof(FontVertex) * m_fontVertices.size(), m_fontVertices.data(), GL_STREAM_DRAW);
```
ajtribick, line 498: Do we still need the previous `glBufferData` calls on lines 495-496?

375gnu: yes, but with a proper size, lem me fix this. (passing data==nullptr and the same size and usage as earlier means "i don't need this data, if the buffer is in use, then allocate a new one").

src/celttf/truetypefont.cpp
```diff
@@ -113,15 +117,14 @@ struct TextureFontPrivate
     bool                    m_shaderInUse{ false };
     std::vector<FontVertex> m_fontVertices;
 
-    static GLuint m_vbo;
-    static GLuint m_vio;
+    gl::VertexObject m_vao;
+    gl::Buffer       m_vbo{ gl::Buffer::TargetHint::Array };
```
ajtribick, line 121: Since the `gl::Buffer` is non-copyable, the copy constructor and assignment operator on `TextureFontPrivate` should be switched to `delete` rather than `default` (causes a warning on clang)

ajtribick:
I'm getting a segfault when the star labels are being rendered, need to get some time to investigate it.

src/celrender/gl/vertexobject.cpp
```diff
@@ -0,0 +1,273 @@
...
+        {
+            glBindVertexArray(m_id);
+            enableAttribArrays();
+            m_bufferDesc.clear();
+        }
+    }
+    else
+    {
+        if (isVAOSupported())
+            glBindVertexArray(m_id);
```
ajtribick, line 199: I think we also need to bind the index buffer here in case of `isIndexed()` - at least this seems to resolve the segfault

375gnu: Do you have both in GNU/Linux and Windows?

375gnu: > I'm getting a segfault when the star labels are being rendered, need to get some time to investigate it.

Do you have it both in GNU/Linux and Windows? What HW do you use?

ajtribick: Occurs on both Linux (Kubuntu 23.04) and Windows, using Nvidia GTX 1070 card. The error occurs in the `glDrawElements` call on line 207. Previous calls to the code for non-indexed buffers succeed, so does the first call to this method for an indexed buffer (which hits the `!m_initialized` code path). The second call, where the buffer is already initialized triggers the segfault.

ajtribick: On both Windows and Linux, I can avoid the segfault by using:

```cpp
if (!m_initialized)
{
    // ... snip ...
}
else
{
    if (isVAOSupported())
    {
        glBindVertexArray(m_id);
        if (isIndexed())
            glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, m_idxBufferId);
    }
    else
        enableAttribArrays();
}
```

375gnu: Then this looks like a bug in Nvidia drivers. I'll add an workaround for them.

375gnu: please check now. (btw, there is still a bug in nvidia drivers because they should not segfault)

ajtribick (approved):

PR #1689, opened 2023-04-22
Category refactor

ajtribick:
Re-implement the category system to store information outside the objects.
By eliminating the vtable and the category set, this reduces the size of `Star` from 56 to 32 bytes (x64)


375gnu (approved):

375gnu:
Around 100Mb less in use now!

PR #1690, opened 2023-04-22
Remove old versions of stars.dat generation tools

ajtribick:
These probably aren't necessary any more.


375gnu (dismissed):

375gnu:
@ajtribick what about lightweight databases? I remember once you told about XHIP database.

ajtribick:
I think it makes sense to create those by pruning the full database to just the HIP entries rather than using an entirely separate generation process.

375gnu (approved):

PR #1691, opened 2023-04-22
Add tool to reduce star database files to HIP entries only

ajtribick:
As discussed in #1690

### DIFF
--- a/src/tools/stardb/prunestars.py
+++ b/src/tools/stardb/prunestars.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+
+# prunestars.py
+#
+# Original version by Andrew Tribick
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 2 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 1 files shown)


src/tools/stardb/prunestars.py
```diff
@@ -0,0 +1,133 @@
...
+
+def prune_stars_dat(stars_path: pathlib.Path, out_stars_path: pathlib.Path) -> None:
+    """Prunes the stars.dat file"""
+    with open(stars_path, 'rb') as file:
+        data: bytes = file.read()
+
+    header_format = struct.Struct('<8sHI')
+    star_format = struct.Struct('<I')
+    header, version, total_stars = header_format.unpack_from(data, 0)
+    if header != b'CELSTARS' or version != 0x0100:
```
375gnu, line 37: what about version 0x0200?

375gnu (changes requested):

src/tools/stardb/prunestars.py
ajtribick: So far, only version 0x0100 is permitted in stardb.cpp: https://github.com/CelestiaProject/Celestia/blob/15e15532e3acaf055525d74bdee73c07b79b4f10/src/celengine/stardb.cpp#L673

375gnu (approved):

Issue #1461, opened 2022-12-02
Screenshot saved as mirrored image


mattiaverga:
**Describe the bug**
When I save a screenshot with "File -> Grab image" or F10, the result is vertical mirrored screenshot.

**To Reproduce**
1. Open Celestia
2. Choose File -> grab image
3. Save a png or jpeg screenshot

**Expected behavior**
A properly saved screenshot

**Screenshots**
![immagine](https://user-images.githubusercontent.com/17218209/205358951-83c6d8a3-8da7-499d-8c48-20f992b9644d.png)

**Desktop (please complete the following information):**
 - OS: Linux Fedora (RPMs self compiled from [COPR repository](https://copr.fedorainfracloud.org/coprs/mattia/Astronomy/))
 - Frontend: QT6
 - Version: 1.7.0 (master branch checkout)

mattiaverga:
With `-DENABLE_GLES=ON` the screenshot is saved correctly, but stars are not rendered in fuzzy nor scaled mode:
![immagine](https://user-images.githubusercontent.com/17218209/205439696-382cd874-002f-48ce-888c-12771b7632ad.png)

375gnu:
> With -DENABLE_GLES=ON the screenshot is saved correctly, but stars are not rendered in fuzzy nor scaled mode:

That's because you need a GLES build of Qt as well.

mattiaverga:
After #1696 I now get a totally garbage image using qt6 frontend:
![test](https://github.com/CelestiaProject/Celestia/assets/17218209/396d7434-bbf3-4fe8-87b3-aea88e1598c8)

gtk frontend still works fine.
Running celestia-qt6 from CLI doesn't show any wrong message, so I'm not sure how to report more useful information.

PR #1696, opened 2023-04-29
[qt] mirror captured imaged

375gnu:
Closes: #1461


ajtribick:
This results in the capture being mirrored for me, i.e. applying this change results in the same orientation as for the bug report.

Kubuntu 23.04, Nvidia GTX 1070

375gnu:
So the problem should be fixed by setting some Qt options.

375gnu:
Now it should work properly. @ajtribick please check.

ajtribick (approved):
lgtm

PR #1700, opened 2023-05-04
Updates to testing

ajtribick:
- Update catch2 to 2.13.10
- Combine test executables, use catch2 cmake integration
- [win] Copy dependency dlls to output folder


375gnu (changes requested):
could you add authorship information for added cmake files? and also update `debian/copyright`?

ajtribick:
I think it's worth switching the tests to use doctest instead of catch2, since the current version of catch2 has moved away from header-only. Will update this PR once I have done that - for now will keep the tests in their own separate files rather than using in-source testing though.

375gnu (approved):

PR #1711, opened 2023-05-24
Fix feature check for numbers header

ajtribick:
Add a missing include to define the library test macro


375gnu (approved):

375gnu:
@ajtribick could you check your e-mail?

Issue #1587, opened 2023-02-19
Font file searching should also go through LocaleFilename


levinli303:
We will need to bundle CJK fonts in the future, and having font file searching go through `LocaleFilename` allows us to specify one single name in celestia.cfg. #761

PR #1713, opened 2023-05-28
Process font name with LocaleFilename

375gnu:
Closes: #1587


375gnu:
@levinli303 `Celestia::set*Font` API wasn't updated because it's not used inside Celestia, so I leave this up to you to decide if we need to update them as well.

levinli303 (approved):

PR #1716, opened 2023-06-02
Allow to use ICU from Windows SDK

levinli303:
Off in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       shell: pwsh
       run: |
         vcpkg --triplet=${{matrix.platform}}-windows install --recurse `
-          cspice eigen3 ffmpeg[x264] fmt freetype gettext icu libepoxy libjpeg-turbo libpng luajit `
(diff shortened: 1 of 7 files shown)


ajtribick:
vcpkg issue is fixed by #1717

ajtribick:
I guess the question here is whether or not we should use this as part of the default Windows build, if I understand it this would essentially result in the compiled application dropping support for Windows versions earlier than Windows 10 1903, though I think all affected versions are no longer supported by Microsoft.

levinli303:
> I guess the question here is whether or not we should use this as part of the default Windows build, if I understand it this would essentially result in the compiled application dropping support for Windows versions earlier than Windows 10 1903, though I think all affected versions are no longer supported by Microsoft.

i think we will be turning this option off and bundling external icu if we make a release. one of the reasons i made this PR is that i can get rid of the patch i had in my UWP build for the headers (vcpkg's ICU does not support UWP and i do not have want to dig into the reason why, also saves space if we do not need to bundle icu).

PR #1719, opened 2023-06-08
Remove the legacy GTK front-end

ajtribick:
Since we don't really have the resources to maintain multiple full-featured front-ends, removing the legacy GTK interface seems to me to be an option worth considering. This PR does that.


375gnu:
I need it because Qt frontend 1) doesn't work under vogl and renderdoc (gl debuggers) 2) doesn't support gl es as easy as gtk. sdl is not a choice for debuggers also because often i need to disable unneeded features and leave only thing i'm debugging.

375gnu:
We can simply leave it without updates.

ajtribick:
The problem is that it still costs to keep this around even if it is unmaintained - if a change to the core propagates sufficiently far upwards, we then have to figure out how to get it to build. This makes it harder to do changes in areas like celestiacore.

If it is really so valuable to have GTK around, an alternative would be to keep the files and the build option, but remove it from the CI and the packages list, so that breaking GTK doesn't cause a build failure.

ajtribick:
And at the very least we should drop the GTK2 support

375gnu:
Let's disable it in CI. GTK2 is still required as vogl doesn't support EGL, only GLX.

PR #1721, opened 2023-06-11
Track VBO/VAO bind calls and related changes

375gnu:
The idea is to reduce number of `bind` calls and do them only when needed.


ajtribick:
I get a segmentation fault on the second rendered frame (the `glDrawArrays` call on line 187 of vertexobject.cpp - during the galaxy rendering), need to do more investigation here to figure out what's going on.

Issue #1735, opened 2023-06-23
qt: Event finder broken in non english locales


375gnu:
**Describe the bug**
For any object it shows message box "$Object is not allowed object", where $Object is a localized planet name.

**Desktop (please complete the following information):**
 - OS: GNU/Linux, perhaps all
 - Frontend: Qt
 - Version: 1.7.0

Issue #1736, opened 2023-06-25
Universe::findPath doesn't work with localized names 


375gnu:
**Describe the bug**
`Universe::findPath` doesn't work with localized names even when `i18n` parameter is `true`.

**To Reproduce**
1. Open search console.
2. Type `sol/зямля` or any other full object name in your language.
3. See that nothing is selected.

**Expected behavior**
Object searched should be found, e.g. Earth in the example above.

**Desktop (please complete the following information):**
 - OS: all
 - Frontend: all
 - Version: 1.6, 1.7.0

ajtribick:
Search works fine for me when using German. I suspect the issue is that the case-insensitive searching is built on top of the standard library `tolower`/`towlower` functions, which may well be limited to the ASCII subset of characters. If so, the search would work if you capitalised the Cyrillic letters exactly the same as they are in the file.

Fix would then be to use either ICU casefolding or the string collation functions.

375gnu:
https://en.cppreference.com/w/cpp/string/byte/tolower says that locale is taken into account:

> Converts the given character to lowercase according to the character conversion rules defined by the currently installed C locale.

PR #1737, opened 2023-06-25
[qt] use english names to search for eclipses

375gnu:
Closes: #1735 

But the root cause is #1736


375gnu:
@ajtribick @levinli303 please check

PR #1741, opened 2023-06-27
Move zoom to observer class

levinli303:
moving zoom to Observer so that renderer class can access this information to support more flexible projection mode

### DIFF
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -585,14 +585,16 @@ void CelestiaCore::mouseButtonUp(float x, float y, int button)
         }
         else if (button == MiddleButton)
         {
-            if ((*activeView)->zoom != 1)
+            auto observer = (*activeView)->getObserver();
+            auto currentZoom = observer->getZoom();
+            if (currentZoom != 1.0f)
             {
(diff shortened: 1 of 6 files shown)


src/celestia/celestiacore.cpp
```diff
@@ -2416,8 +2418,8 @@ void CelestiaCore::setFOVFromZoom()
                 v->observer->setFOV(MaximumFOVFisheye);
             else
             {
-                double fov = 2 * atan(height * v->height / (screenDpi / 25.4) / 2. / distanceToScreen) / v->zoom;
-                v->observer->setFOV((float) fov);
+                float fov = 2.0f * std::atan(static_cast<float>(height) * v->height / (static_cast<float>(screenDpi) / 25.4f) / 2.0f / static_cast<float>(distanceToScreen)) / v->getObserver()->getZoom();
```
375gnu, line 2421: could you create a helper method to calculate `2.0f * std::atan(static_cast<float>(height) * v->height / (static_cast<float>(screenDpi) / 25.4f) / 2.0f / static_cast<float>(distanceToScreen))`?

375gnu (changes requested):

375gnu (approved):

PR #1748, opened 2023-06-29
Add projection mode class to replace perspective/fisheye handling

levinli303:


src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include "fisheyeprojectionmode.h"
+#include <celmath/geomutil.h>
+
+using namespace Eigen;
```
375gnu, line 13: remove this line

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+using namespace celmath;
```
375gnu, line 14: and this one

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+Matrix4f FisheyeProjectionMode::getProjectionMatrix(float nearZ, float farZ, float zoom) const
+{
+    float aspectRatio = width / height;
+    return Ortho(-aspectRatio, aspectRatio, -1.0f, 1.0f, nearZ, farZ);
+}
+
+float FisheyeProjectionMode::getMinimumFOV() const
+{
+    return degToRad(179.99f);
```
375gnu, line 31: provide a constant instead of hardcoding the same value here and there

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include "fisheyeprojectionmode.h"
+#include <celmath/geomutil.h>
+
+using namespace Eigen;
+using namespace celmath;
+
+namespace celengine
```
375gnu, line 16:
```suggestion
namespace celestia::engine
```

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+
+FisheyeProjectionMode::FisheyeProjectionMode(float width, float height) : ProjectionMode(), width(width), height(height)
```
375gnu, line 19:
```suggestion
FisheyeProjectionMode::FisheyeProjectionMode(float width, float height) :
    ProjectionMode(),
    width(width),
    height(height)
```

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+}
+
+Matrix4f FisheyeProjectionMode::getProjectionMatrix(float nearZ, float farZ, float zoom) const
```
375gnu, line 23:
```suggestion
Matrix4f FisheyeProjectionMode::getProjectionMatrix(float nearZ, float farZ, float /*zoom*/) const
```

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    return degToRad(179.99f);
+}
+
+float FisheyeProjectionMode::getFOV(float zoom) const
+{
+    return degToRad(179.99f);
+}
+
+float FisheyeProjectionMode::getZoom(float fov) const
```
375gnu, line 44:
```suggestion
float FisheyeProjectionMode::getZoom(float /*fov*/) const
```

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    return 1.0f;
+}
+
+float FisheyeProjectionMode::getPixelSize(float zoom) const
```
375gnu, line 49:
```suggestion
float FisheyeProjectionMode::getPixelSize(float /*zoom*/) const
```

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    return 2.0f / height;
+}
+
+double FisheyeProjectionMode::getViewConeAngleMax(float zoom) const
+{
+    return std::cos(static_cast<double>(getFOV(zoom)) / 2.0);
+}
+
+Vector3f FisheyeProjectionMode::getPickRay(float x, float y, float zoom) const
```
375gnu, line 59:
```suggestion
Vector3f FisheyeProjectionMode::getPickRay(float x, float y, float /*zoom*/) const
```

src/celengine/fisheyeprojectionmode.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void FisheyeProjectionMode::setSize(float w, float h)
+{
+    width = w;
+    height = h;
+}
+
+std::shared_ptr<ProjectionMode> FisheyeProjectionMode::clone()
+{
+    return std::make_shared<FisheyeProjectionMode>(*this);
```
375gnu, line 97: i'm unsure if this works as expected

src/celengine/fisheyeprojectionmode.h
```diff
@@ -0,0 +1,52 @@
...
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#pragma once
+
+#include <celengine/projectionmode.h>
+
+namespace celengine
```
375gnu, line 14:
```suggestion
namespace celestia::engine
```

src/celengine/perspectiveprojectionmode.cpp
```diff
@@ -0,0 +1,97 @@
...
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include "perspectiveprojectionmode.h"
+#include <celmath/geomutil.h>
+
+using namespace Eigen;
+using namespace celmath;
+
+namespace celengine
```
375gnu, line 16: see above

src/celengine/perspectiveprojectionmode.cpp
```diff
@@ -0,0 +1,97 @@
...
+{
+
+PerspectiveProjectionMode::PerspectiveProjectionMode(float width, float height, int distanceToScreen, int screenDpi) : ProjectionMode(), width(width), height(height), distanceToScreen(distanceToScreen), screenDpi(screenDpi)
```
375gnu, line 19: as above

src/celengine/render.cpp
```diff
@@ -247,11 +236,10 @@ Renderer::Renderer() :
     windowWidth(0),
     windowHeight(0),
     fov(FOV),
-    cosViewConeAngle(computeCosViewConeAngle(fov, 1, 1)),
+    cosViewConeAngle(std::sqrt(3.0)),
```
375gnu, line 239: why 3.0?

src/celengine/render.cpp
```diff
@@ -1336,7 +1316,7 @@ static Vector3d astrocentricPosition(const UniversalCoord& pos,
 void Renderer::autoMag(float& faintestMag)
 {
     float fieldCorr;
-    if (getProjectionMode() == ProjectionMode::FisheyeMode)
+    if (getProjectionMode()->isFisheye())
         fieldCorr = 2.0f - 2000.0f / (windowHeight / (screenDpi / 25.4f / 3.78f) + 1000.0f); // larger window height = more stars to display
```
375gnu, line 1320: instead of providing `isFisheye()` provide `getFieldCorrection()`

src/celengine/render.cpp
```diff
@@ -4295,7 +4277,7 @@ Renderer::renderAnnotations(vector<Annotation>::iterator startIter,
     float d0 = farDist - nearDist;
     float d1 = -(farDist + nearDist) / d0; // Used in perspective projection
     float d2 = -2.0f * nearDist * farDist / d0; // Used in perspective projection
-    bool fisheye = projectionMode == ProjectionMode::FisheyeMode;
+    bool fisheye = getProjectionMode()->isFisheye();
```
375gnu, line 4280: as you provide new classes to handle projection, try to avoid direct `fisheye` usage

src/celrender/openclusterrenderer.h
```diff
@@ -27,7 +27,7 @@ class OpenClusterRenderer
 
     void add(const OpenCluster *cluster, const Eigen::Vector3f &offset, float brightness, float nearZ, float farZ) const;
 
-    void update(const Eigen::Quaternionf &viewerOrientation, float pixelSize, float fov) const;
+    void update(const Eigen::Quaternionf &viewerOrientation, float pixelSize, float fov, float zoom) const;
```
375gnu, line 30: it seems it makes sense to pass Observer instead of viewerOrientation and zoom

src/celengine/fisheyeprojectionmode.cpp
375gnu: and i don't think that we actually need this.

you have an object that owns ProjectionMode, it should have std::unique_ptr to it. other classes that doesn't own it should use raw pointer.

375gnu (changes requested):

src/celengine/fisheyeprojectionmode.cpp
levinli303: yeah this should be unique_ptr, this is used in setFOVFromZoom/setZoomFromFOV where i do not wish to change renderer's projectionmode's width/height

src/celrender/openclusterrenderer.h
levinli303: i'm adding support for transform for camera to support VR, so viewerOrientation != observer.getOrientation(), the i intend to store camera transform in Render class.

src/celengine/render.cpp
levinli303: this should be 1 / sqrt(3), since the old computeCosViewConeAngle(fov, 1, 1) with fov = 45 deg would return this value.

src/celengine/projectionmode.h
```diff
@@ -0,0 +1,49 @@
...
+    virtual float getNormalizedDeviceZ(float nearZ, float farZ, float z) const = 0;
+
+    virtual Eigen::Vector3f getPickRay(float x, float y, float zoom) const = 0;
+    virtual bool isFisheye() const = 0;
+
+    virtual void setScreenDpi(int screenDpi) = 0;
+    virtual void setDistanceToScreen(int distanceToScreen) = 0;
+    virtual void setSize(float width, float height) = 0;
+
+    virtual std::unique_ptr<ProjectionMode> clone() = 0;
```
375gnu, line 46: you don't need this method. just use the raw pointer

src/celengine/perspectiveprojectionmode.h
```diff
@@ -0,0 +1,56 @@
...
+    float getNormalizedDeviceZ(float nearZ, float farZ, float z) const override;
+
+    Eigen::Vector3f getPickRay(float x, float y, float zoom) const override;
+
+    bool isFisheye() const override;
+
+    void setScreenDpi(int screenDpi) override;
+    void setDistanceToScreen(int distanceToScreen) override;
+    void setSize(float width, float height) override;
+
```
375gnu, line 46: you don't need so many virtual functions. just add width, height, minFox, maxFov etc to the base class. override only methods which need to compute something

src/celengine/projectionmode.h
```diff
@@ -0,0 +1,49 @@
...
+    virtual float getFieldCorrection(float zoom) const = 0;
+
+    // Calculate the cosine of half the maximum field of view. We'll use this for
+    // fast testing of object visibility.
+    virtual double getViewConeAngleMax(float zoom) const = 0;
+
+    virtual float getNormalizedDeviceZ(float nearZ, float farZ, float z) const = 0;
+
+    virtual Eigen::Vector3f getPickRay(float x, float y, float zoom) const = 0;
+    virtual bool isFisheye() const = 0;
```
375gnu, line 40: try to re-implement your code without this method, it breaks the idea of base and derived classes

375gnu (approved):

src/celengine/render.cpp
375gnu: I'm asking why do you use a magic constant? please add a `const` or `constexpr` variable with comment why its value is such.

levinli303: i'll set to 0 since the initial value does not realy matter, it gets calculated before usage.

src/celengine/perspectiveprojectionmode.h
levinli303: moved set methods to save values in protected.

src/celengine/projectionmode.h
levinli303: removed

PR #1755, opened 2023-07-03
Add support for observer transform

levinli303:
#1752 is used for per eye rendering and it does not change observer's orientation.

This PR changes the value retrieved from `observer.getOrientation()` because the orientation of the device might change and VR frameworks gives us a transform from original pose (thus we cannot update observer's orientation directly).

In Apple's ARKit/CompositorServices:
ar_pose_get_origin_from_device_transform gets the transform from device origin used in this PR
cp_view_get_transform gets the per eye transform used in #1752

### DIFF
--- a/src/celengine/observer.h
+++ b/src/celengine/observer.h
@@ -128,6 +128,9 @@ public:
     void          setOrientation(const Eigen::Quaternionf&);
     void          setOrientation(const Eigen::Quaterniond&);
 
+    Eigen::Matrix3d getOrientationTransform() const;
+    void          setOrientationTransform(const Eigen::Matrix3d&);
+
     Eigen::Vector3d getVelocity() const;
     void          setVelocity(const Eigen::Vector3d&);
     Eigen::Vector3d getAngularVelocity() const;
@@ -282,7 +285,11 @@ public:
                                    JourneyParams &jparams,
                                    double centerTime);
 
+    void setOriginalOrientation(const Eigen::Quaternionf&);
+    void setOriginalOrientation(const Eigen::Quaterniond&);
     void updateUniversal();
+    void updateOrientation();
+    Eigen::Quaterniond undoTransform(const Eigen::Quaterniond&);
     void convertFrameCoordinates(const ObserverFrame::SharedConstPtr &newFrame);
 
  private:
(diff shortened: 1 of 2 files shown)


src/celengine/observer.h
```diff
@@ -282,22 +285,29 @@ class Observer
                                    JourneyParams &jparams,
                                    double centerTime);
 
+    void          setOriginalOrientation(const Eigen::Quaternionf&);
```
375gnu, line 288:
```suggestion
    void setOriginalOrientation(const Eigen::Quaternionf&);
```

src/celengine/observer.h
```diff
@@ -282,22 +285,29 @@ class Observer
...
+    void          setOriginalOrientation(const Eigen::Quaterniond&);
```
375gnu, line 289:
```suggestion
    void setOriginalOrientation(const Eigen::Quaterniond&);
```

src/celengine/observer.h
```diff
@@ -282,22 +285,29 @@ class Observer
...
+    Eigen::Quaterniond undoTransform(const Eigen::Quaterniond&);
     void convertFrameCoordinates(const ObserverFrame::SharedConstPtr &newFrame);
 
  private:
     double              simTime{ 0.0 };
 
     // Position, orientation, and velocity in the observer's reference frame
     UniversalCoord      position{ 0.0, 0.0, 0.0 };
-    Eigen::Quaterniond  orientation{ Eigen::Quaternionf::Identity() };
+    Eigen::Quaterniond  originalOrientation{ Eigen::Quaternionf::Identity() };
```
375gnu, line 300:
```suggestion
    Eigen::Quaterniond  originalOrientation{ Eigen::Quaterniond::Identity() };
```

src/celengine/observer.h
```diff
@@ -282,22 +285,29 @@ class Observer
...
+    Eigen::Quaterniond  transformedOrientation{ Eigen::Quaternionf::Identity() };
```
375gnu, line 301:
```suggestion
    Eigen::Quaterniond  transformedOrientation{ Eigen::Quaterniond::Identity() };
```

375gnu (changes requested):

375gnu (approved):

PR #1757, opened 2023-07-04
Add toggles for message text and selection pointer

levinli303:


src/celestia/celestiacore.cpp
```diff
@@ -4555,8 +4552,7 @@ bool CelestiaCore::isRecording()
 
 void CelestiaCore::flash(const string& s, double duration)
 {
-    if (hudDetail > 0)
```
levinli303, line 4558: i changed this to ignore HudDetail value because i feel that it does not fit with other HUDs, this is more like a message shown to the user.

src/celengine/render.h
```diff
@@ -200,6 +200,7 @@ class Renderer
         ShowSpacecrafts         = 0x0000000100000000,
         ShowFadingOrbits        = 0x0000000200000000,
         ShowPlanetRings         = 0x0000000400000000,
+        ShowTextMessage         = 0x0000000800000000,
```
375gnu, line 203: this definitely shouldn't be a render flag as it's never used in renderer.

src/celengine/render.cpp
```diff
@@ -633,6 +633,16 @@ unsigned int Renderer::getResolution() const
     return textureResolution;
 }
 
+void Renderer::setBoundless(bool b)
```
375gnu, line 636: It's better to call this `disableSelectionPoiner()` as I suppose it's the only purpose of this method.

src/celestia/celestiacore.cpp
375gnu: `hudDetail == 0` means "don't show any information on the screen", so it's better to restore it

src/celestia/celestiacore.cpp
```diff
@@ -5001,6 +5003,13 @@ void CelestiaCore::setLayoutDirection(LayoutDirection value)
     renderer->setRTL(layoutDirection == LayoutDirection::RightToLeft);
 }
 
+std::string CelestiaCore::getCurrentText() const
```
375gnu, line 5006:
```suggestion
std::string_view CelestiaCore::getCurrentText() const
```

src/celestia/celestiacore.cpp
```diff
@@ -5001,6 +5003,13 @@ void CelestiaCore::setLayoutDirection(LayoutDirection value)
...
+{
+    if (currentTime < messageStart + messageDuration && messageTextPosition)
+        return messageText;
+    return "";
```
375gnu, line 5010:
```suggestion
    return {};
```

375gnu (changes requested):

src/celengine/render.cpp
```diff
@@ -633,6 +633,16 @@ unsigned int Renderer::getResolution() const
     return textureResolution;
 }
 
+void Renderer::enableSelectionPointer()
+{
+    isSelectionPointerEnabled = true;
```
375gnu, line 638: this is C++, not Java :) just `showSelPointer` or `showSelectionPointer`

src/celengine/render.cpp
```diff
@@ -1678,7 +1688,7 @@ void Renderer::draw(const Observer& observer,
 
     renderForegroundAnnotations(FontNormal);
 
-    if (!selectionVisible && (renderFlags & ShowMarkers))
+    if (isSelectionPointerEnabled && !selectionVisible && (renderFlags & ShowMarkers))
```
375gnu, line 1691: as you're changing this line, update `(renderFlags & ShowMarkers)` to `(renderFlags & ShowMarkers) != 0`

src/celestia/celestiacore.cpp
```diff
@@ -3577,26 +3577,29 @@ void CelestiaCore::renderOverlay()
     }
 
     // Text messages
-    if (messageText != "" && currentTime < messageStart + messageDuration && messageTextPosition)
+    if (showMessageText)
```
375gnu, line 3580: just `showMessage`

src/celestia/celestiacore.cpp
```diff
@@ -5001,6 +5004,23 @@ void CelestiaCore::setLayoutDirection(LayoutDirection value)
...
+{
+    return showMessageText;
+}
+
+void CelestiaCore::setShowMessageText(bool value)
+{
+    showMessageText = value;
+}
+
+std::string_view CelestiaCore::getCurrentMessageText() const
```
375gnu, line 5017:
```suggestion
std::string_view CelestiaCore::getCurrentMessage() const
```

src/celestia/celestiacore.cpp
```diff
@@ -5001,6 +5004,23 @@ void CelestiaCore::setLayoutDirection(LayoutDirection value)
     renderer->setRTL(layoutDirection == LayoutDirection::RightToLeft);
 }
 
+bool CelestiaCore::getShowMessageText() const
```
375gnu, line 5007: do we need this method?

src/celestia/celestiacore.cpp
```diff
@@ -5001,6 +5004,23 @@ void CelestiaCore::setLayoutDirection(LayoutDirection value)
...
+{
+    return showMessageText;
+}
+
+void CelestiaCore::setShowMessageText(bool value)
```
375gnu, line 5012:
```suggestion
void CelestiaCore::enableMessages(bool value)
```

375gnu (changes requested):

375gnu (approved):

PR #1762, opened 2023-07-08
Fix some warnings reported by gcc/Sonar

ajtribick:


ajtribick:
Will update this after #1763 is merged

ajtribick:
Not going to fix the duplication here as would require restructuring/eliminating the (const) void specializations of `array_view`

375gnu (approved):

PR #1763, opened 2023-07-08
Replace array_view<const void> specialization with array_view<std::byte>

ajtribick:


375gnu:
This is designed after https://github.com/mosra/magnum which uses `const void`. I would like to be close to it as much as possible.

ajtribick:
I prefer to be closer to `std::span` (and the utility functions `as_bytes`/`as_writable_bytes`) with the view to replacing this class with a backport.

375gnu:
As I wrote in PR which provided updated updated Buffer & VertexObject classes, I want to replace own wrappers with Magnun in Celestia 1.8 so now I want to have difference as small as possible between two implementations.

375gnu:
And span is not a view, it allows writing. It's like slices in some other languages.

ajtribick:
`std::span<const T>` is a view, `std::span<T>` allows writing.

PR #1764, opened 2023-07-08
Another attempt at fixing the Qt mirrored capture issue

### LINKED ISSUE #1759: After #1696 I now get a totally garbage image using qt6 frontend:
After #1696 I now get a totally garbage image using qt6 frontend:
test

gtk frontend still works fine.
Running celestia-qt6 from CLI doesn't show any wrong message, so I'm not sure how to report more useful information.

_Originally posted by @mattiaverga in https://github.com/CelestiaProject/Celestia/issues/1461#issuecomment-1621292138_

ajtribick:
Hypothesis: Qt doesn't like us inverting the pixel format using the MESA extension

Resolves #1759

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -553,8 +553,10 @@ bool Renderer::init(int winWidth, int winHeight, const DetailOptions& _detailOpt
     glCullFace(GL_BACK);
 
 #ifndef GL_ES
-    if (gl::MESA_pack_invert)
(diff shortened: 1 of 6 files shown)


ajtribick:
So far this appears to work with/without antialiasing on my machine (NVidia), would be good to have further tests.

375gnu (approved):
Tested with intel/mesa.

PR #1767, opened 2023-07-09
Add coefficient setting for acceleration and deceleration

levinli303:
... and set speed to zero when decelerating from (-0.1, 0.1) instead of going on (almost) forever to reach 0

The coefficient could be useful for gas/brake like controls which are more than just on/off.

### DIFF
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -2078,20 +2078,25 @@ void CelestiaCore::tick()
 
     sim->getObserver().setAngularVelocity(av);
 
-    if (keysPressed[(int)'A'] || joyButtonsPressed[JoyButton2])
+    if (keysPressed[static_cast<int>('A')] || joyButtonsPressed[JoyButton2])
     {
         bSetTargetSpeed = true;
 
-        if (sim->getTargetSpeed() == 0.0f)
-            sim->setTargetSpeed(0.1f);
+        if (float currentSpeed = sim->getTargetSpeed(); currentSpeed != 0.0f)
+            sim->setTargetSpeed(currentSpeed * std::exp(static_cast<float>(dt) * 3.0f * accelerationCoefficient));
         else
-            sim->setTargetSpeed(sim->getTargetSpeed() * (float) exp(dt * 3));
+            sim->setTargetSpeed(0.1f);
     }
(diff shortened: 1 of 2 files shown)


src/celestia/celestiacore.cpp
```diff
@@ -2078,20 +2078,31 @@ void CelestiaCore::tick()
...
     sim->getObserver().setAngularVelocity(av);
 
-    if (keysPressed[(int)'A'] || joyButtonsPressed[JoyButton2])
+    if (keysPressed[static_cast<int>('A')] || joyButtonsPressed[JoyButton2])
     {
         bSetTargetSpeed = true;
 
-        if (sim->getTargetSpeed() == 0.0f)
+        float currentSpeed = sim->getTargetSpeed();
+        if (currentSpeed == 0.0f)
```
375gnu, line 2086:
```suggestion
        if (float currentSpeed = sim->getTargetSpeed(); currentSpeed == 0.0f)
```

src/celestia/celestiacore.cpp
```diff
@@ -2078,20 +2078,31 @@ void CelestiaCore::tick()
...
+            sim->setTargetSpeed(currentSpeed * std::exp(static_cast<float>(dt) * 3.0f * accelerationCoefficient));
     }
-    if (keysPressed[(int)'Z'] || joyButtonsPressed[JoyButton1])
+    if (keysPressed[static_cast<int>('Z')] || joyButtonsPressed[JoyButton1])
     {
-        bSetTargetSpeed = true;
-
-        sim->setTargetSpeed(sim->getTargetSpeed() / (float) exp(dt * 3));
+        float currentSpeed = sim->getTargetSpeed();
+        if (currentSpeed != 0.0f)
```
375gnu, line 2094:
```suggestion
        if (float currentSpeed = sim->getTargetSpeed(); currentSpeed != 0.0f)
```

src/celestia/celestiacore.cpp
```diff
@@ -2078,20 +2078,31 @@ void CelestiaCore::tick()
...
+        {
+            bSetTargetSpeed = true;
+            if (std::abs(currentSpeed) < 0.1f)
+            {
```
375gnu, line 2098: remove brackets in if & else blocks

375gnu (approved):

PR #1770, opened 2023-07-12
Add bit.h implementing C++20 <bit> and use it in byte manipulations

375gnu:


src/celcompat/bit.h
```diff
@@ -0,0 +1,183 @@
...
+namespace celestia::compat
+{
+
+#if __cpp_lib_byteswap
+
+using std::byteswap;
+
+#else
+
+namespace
```
ajtribick, line 126: This should be named (e.g. "internal" or "impl"), otherwise it will create a new anonymous namespace in all translation units.

src/celcompat/bit.h
```diff
@@ -0,0 +1,183 @@
...
+    if constexpr (sizeof(T) == 1)
+        return n;
+    else if constexpr (sizeof(T) == 2)
+        return static_cast<T>(bswap_16(static_cast<std::uint16_t>(n)));
+    else if constexpr (sizeof(T) == 4)
+        return static_cast<T>(bswap_32(static_cast<std::uint32_t>(n)));
+    else if constexpr (sizeof(T) == 8)
+        return static_cast<T>(bswap_64(static_cast<std::uint64_t>(n)));
+    else
+        static_assert(true, "Unexpected integer size");
```
ajtribick, line 178: This assertion will never fire, probably better to do `static_assert(sizeof(T) == 1 || sizeof(T) == 2 || /* etc */)`

ajtribick (changes requested):

ajtribick (approved):

PR #1772, opened 2023-07-13
Updates to strnatcmp

ajtribick:
- Only skip leading space rather than all space within the string.
- Use a mapper class to avoid allocating wstrings.
- Skip over runs of digits once the comparison is complete.


375gnu (dismissed):

375gnu (approved):
actually that's ok

PR #1775, opened 2023-07-16
Duplicate vertex instead to avoid repeated glBufferSubData for drawing galaxies

levinli303:
Galaxy rendering is most lagging, this should fix it with a trade off for memory usage.


375gnu:
Don't mix GL2 and GL3 methods into one function. I have an idea to provide GL3.3/GLES3.2 builds and with them we can simply disable/remove GL2 code path. And the code is more readable when we have separate functions for different render methods.

levinli303:
> Don't mix GL2 and GL3 methods into one function. I have an idea to provide GL3.3/GLES3.2 builds and with them we can simply disable/remove GL2 code path. And the code is more readable when we have separate functions for different render methods.

but we would still need to keep the GL2 path indefinitely given that Apple will not bring ES3.2 to Apple platforms and ANGLE's ES3.2 support probably never will not be available on all platforms, so we need to improve the performance. do you suggest duplicate the code GL2 instead?

375gnu:
Yes, duplicate GL2. In GL3 build we'll use `#ifdef GL3` guards so old code won't be included.

levinli303:
> Yes, duplicate GL2. In GL3 build we'll use `#ifdef GL3` guards so old code won't be included.

created separated methods renderGL2/3 and initlaizeGL2/3

shaders/galaxy_vert.glsl
```diff
@@ -1,18 +1,29 @@
...
-    // we pass color index as a normalized byte reusing in_TexCoord0.z
-    float t = in_TexCoord0.z;
-    // we pass alpha values as a normalized byte reusing in_TexCoord0.w
-    float a = in_TexCoord0.w;
-    color = vec4(texture2D(colorTex, vec2(t, 0.0)).rgb, a);
-    texCoord = in_TexCoord0.st;
-    set_vp(in_Position);
+    float s = size * in_Size;
+    vec4 p = m * in_Position;
+    float screenFrac = s / length(p);
```
375gnu, line 22: Original code uses check
```
if (screenFrac < 0.1f)
```
to add a new billboard.

src/celrender/galaxyrenderer.cpp
```diff
@@ -245,144 +215,187 @@ GalaxyRenderer::render()
...
-                auto power = static_cast<unsigned>(std::log(minimumFeatureSize/size)/std::log(kSpriteScaleFactor));
-                if (power < std::numeric_limits<decltype(nPoints)>::digits)
-                    nPoints = std::min(nPoints, 1 << power);
+                if ((i & pow2) != 0)
+                {
+                    pow2 <<= 1;
+                    sizeFactor *= kSpriteScaleFactor;
+                }
+
+                for (int j = 0; j < texCoords.size(); ++j)
```
375gnu, line 293: you can speed up this block if you move all but `texCoord` out of the loop

src/celrender/galaxyrenderer.cpp
```diff
@@ -245,144 +215,187 @@ GalaxyRenderer::render()
...
+    // (indexed) on all the vertices and add the texture coordinate to each
+    const std::array<Eigen::Matrix<GLubyte, 2, 1>, 4> texCoords =
+    {
+        Eigen::Matrix<GLubyte, 2, 1>(  std::uint8_t(0),   std::uint8_t(0)),
+        Eigen::Matrix<GLubyte, 2, 1>(std::uint8_t(255),   std::uint8_t(0)),
+        Eigen::Matrix<GLubyte, 2, 1>(std::uint8_t(255), std::uint8_t(255)),
+        Eigen::Matrix<GLubyte, 2, 1>(  std::uint8_t(0), std::uint8_t(255)),
+    };
+
+    auto s = prog->attribIndex("in_Size");
```
375gnu, line 268: sizeLoc and so on

src/celrender/galaxyrenderer.cpp
```diff
@@ -453,12 +464,10 @@ GalaxyRenderer::initializeGL3(const CelestiaGLProgram *prog)
...
                 true, sizeof(GalaxyVtx), offsetof(GalaxyVtx, brightness));
 
-            m_renderData.emplace_back(std::move(bo), std::move(vo));
+            m_renderDataGL3.emplace_back(std::move(bo), std::move(vo));
         }
         else
-        {
-            m_renderData.emplace_back(gl::Buffer(util::NoCreateT{}), gl::VertexObject(util::NoCreateT{}));
-        }
+            m_renderDataGL3.emplace_back(gl::Buffer(util::NoCreateT{}), gl::VertexObject(util::NoCreateT{}));
```
375gnu, line 470: add `{}` to `else` block as `if` has them

375gnu (changes requested):

shaders/galaxy_vert.glsl
levinli303: should we pass a value to fragment shader to let it discard? 

```
varying float valid;

if (screenFrac < 0.1f)
  valid = 1.0
else
  valid = -1.0
```

```
varying float valid;

if (valid < 0.0)
    discard;
```

375gnu: `discard` is supported with desktop GL since version 3, if i remember correctly. what you can do is to leave position as is, in that case you'll have _degenerated_ triangle (triangle with zero area) and hardware won't draw it.

levinli303: set `gl_Position = vec4(0.0, 0.0, 0.0, 1.0);` when `screenFrac >= 0.1`

src/celrender/galaxyrenderer.cpp
```diff
@@ -207,35 +132,80 @@ GalaxyRenderer::render()
...
+        if (power < std::numeric_limits<decltype(pointCount)>::digits)
+            pointCount = std::min(pointCount, 1 << power);
     }
+    nPoints = pointCount;
+    return true;
+}
+
+struct GalaxyRenderer::RenderDataGL2
+{
+    RenderDataGL2(gl::Buffer &&bo, gl::Buffer &&io, gl::VertexObject &&vo) : bo(std::move(bo)), io(std::move(io)), vo(std::move(vo)) {}
```
375gnu, line 188:
```suggestion
    RenderDataGL2(gl::Buffer &&bo, gl::Buffer &&io, gl::VertexObject &&vo) :
        bo(std::move(bo)),
        io(std::move(io)),
        vo(std::move(vo))
    {
    }
```

src/celrender/galaxyrenderer.cpp
```diff
@@ -245,144 +215,189 @@ GalaxyRenderer::render()
...
+        indices.clear();
     }
-
-    glActiveTexture(GL_TEXTURE0);
-    m_objects.clear();
 }
 
+struct GalaxyRenderer::RenderDataGL3
+{
+    RenderDataGL3(gl::Buffer &&bo, gl::VertexObject &&vo) : bo(std::move(bo)), vo(std::move(vo)) {}
```
375gnu, line 350:
```suggestion
    RenderDataGL3(gl::Buffer &&bo, gl::VertexObject &&vo) :
        bo(std::move(bo)),
        vo(std::move(vo))
    {
    }
```

shaders/galaxy_vert.glsl
```diff
@@ -1,18 +1,36 @@
...
+    if (screenFrac >= 0.1)
+    {
+        // Output viewport center for discarded vertices
```
375gnu, line 25:
```suggestion
        // Output the same value for all discarded vertices so GPU will discard degenerate triangles
```

375gnu (changes requested):

375gnu (approved):

PR #1776, opened 2023-07-16
Add a tick method with elapsed time parameter

levinli303:


375gnu:
What's the purpose of this method?

levinli303:
platform API can give us a predicted timing for presentation, using this time instead of current time can probably reduce latency.

375gnu:
Could you leave an appropriate comment in celestiacore.h then?

levinli303:
> Could you leave an appropriate comment in celestiacore.h then?

added

375gnu (approved):

PR #1779, opened 2023-07-16
Switch most non-Windows usages of wchar_t to char16_t

ajtribick:
- Reduces the size of the console buffer on Linux
- char16_t better matches the ICU C APIs
- Simplify/remove some of the UTF-8 handling routines

Would like to go further to reduce dependency on the `<cwctype>` functions but will leave that for another pull request.

### DIFF
--- a/src/celengine/console.cpp
+++ b/src/celengine/console.cpp
@@ -37,7 +37,7 @@ Console::Console(Renderer& _renderer, int _nRows, int _nColumns) :
     nColumns(_nColumns)
 {
     sbuf.setConsole(this);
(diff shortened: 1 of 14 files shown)


375gnu:
> Reduces the size of the console buffer on Linux

Actually no as we use 16bit wchar_t

ajtribick:
`wchar_t` is 32 bits on Linux - just tested with `static_assert(sizeof(wchar_t) == 4)`

375gnu:
Maybe it was my idea only to set it to 16 bits? Or it was with qmake?

375gnu (approved):

PR #1794, opened 2023-07-23
Save some memory by moving texture coordinates transform to uniforms in lodsphere

levinli303:
Re-opening with the source branch in this repo.

### DIFF
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -870,10 +870,7 @@ ShadowDepth(unsigned int i)
 std::string
 TexCoord2D(unsigned int i, bool transform)
 {
-    auto str = fmt::format("in_TexCoord{}.st", i);
-    if (!transform)
-        return str;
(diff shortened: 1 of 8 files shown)


375gnu:
Let's test this in various scenarios.

src/celengine/shadermanager.cpp
```diff
@@ -851,9 +862,12 @@ ShadowDepth(unsigned int i)
 
 
 std::string
-TexCoord2D(unsigned int i)
+TexCoord2D(unsigned int i, bool transform)
 {
-    return fmt::format("in_TexCoord{}.st", i);
+    auto str = fmt::format("in_TexCoord{}.st", i);
```
ajtribick, line 867: Could avoid the intermediate string by doing

```c++
return transform
    ? fmt::format("(texCoordBase{0} + texCoordDelta{0} * in_TexCoord{0}.st", i)
    : fmt::format("in_TexCoord{}.st", i);
````

ajtribick:
So far I haven't run into any issues running this.

PR #1801, opened 2023-07-24
Remove unused default move constructors of Universe and DSODatabase

ajtribick:
Possible fix for the build issues on Debian 10?


ajtribick:
Apparently SonarCloud counts deleted members in the count of methods in a class. I have raised a false-positive report for this.

ajtribick:
Ah right, it's a gcc bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85552 - fixed in 9.2/10

PR #1805, opened 2023-07-25
Fix compilation with older compilers

375gnu:
E.g. g++ 7.x

### DIFF
--- a/src/celengine/dsodb.h
+++ b/src/celengine/dsodb.h
@@ -24,8 +24,7 @@
 
 #include <celcompat/filesystem.h>
 #include <celengine/dsooctree.h>
-
-class DSONameDatabase;
+#include <celengine/dsoname.h>
 
 constexpr inline unsigned int MAX_DSO_NAMES = 10;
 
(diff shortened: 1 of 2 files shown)


ajtribick:
How far back do we need to support here? E.g. Debian oldstable has 10.2 and oldoldstable is no longer getting standard releases.

375gnu:
I still use Ubuntu 18.04 with gcc 7. Also what about Android and iOS?

375gnu:
Oops! You just commented, not approved :)

levinli303:
On Android there is also a fix i added to make NDK happy because NDK never updated their libc++ header since 2020. But that should be fixed in the upcoming NDK release in the future

https://github.com/celestiamobile/Celestia/commit/4e446db036636b0a1254c86eb58bf8e99ee0563a

375gnu:
@levinli303 could it be added but with appropriate guards, i.e.
```
#if defined(__ANDROID__)
```

later when a fixed NDK is released then we update it to something like
```
#if defined(__ANDROID__) && NDKVERSION < 1234
```

levinli303:
yeah it's possible.  although to make the define __NDK_MAJOR__ work, we need to include another header since the define is not added by compiler.

Issue #1557, opened 2023-02-10
Landing locations on ellipsoidal bodies


Askaniy:
**Problem:**
For bodies using constructions `SemiAxes` and `Oblateness`, the positions of locations are approximated by a sphere, and they are floating in space (see the screenshots below).

**Way to solve:**
Let's notice, that
- Celestia uses planetocentric (projection from center) coordinate system for texturing models and for their locations
- Celestia uses planetographic (projection by normal) for texturing ellipsoids, and _spherical approximation_ for their locations

So, request is to implement recalculation of input planetographic longitude (λ), latitude (φ) and height (h) into x, y, z coordinates using ellipsoid proprieties (semiaxes a_x, a_y, b). I found the formulas in [DOI 10.1007/s00190-011-0514-7](https://link.springer.com/article/10.1007/s00190-011-0514-7):
![formulas1](https://user-images.githubusercontent.com/44873271/218064798-f7f913e9-5f17-4342-9b94-1c5e84d70b1d.png)
![formulas2](https://user-images.githubusercontent.com/44873271/218061889-2fa0a718-ec6b-4952-890f-6edb197d95f8.png)

![Screenshot_20230131_120154](https://user-images.githubusercontent.com/44873271/218057458-042f6ce3-5959-4ce6-b8aa-760c0b05d30b.png)
![Screenshot_20230130_210805](https://user-images.githubusercontent.com/44873271/218057632-e9f22bee-9870-4059-99a9-72f4a1957d4c.png)
(Duplicate here from the Discord server)

375gnu:
@Askaniy could you provide test addons?

Askaniy:
These locations of Ceres already in the content, I think they are good for testing

375gnu:
The formulae should be adjust a bit to match Celestia coordinate system but the result is good:
![image](https://github.com/CelestiaProject/Celestia/assets/1612688/d4fddfe2-9096-4b5f-9ff9-265a0fe6cd13)

375gnu:
But with an object placed on a planet surface image is broken due to #542:

![image](https://github.com/CelestiaProject/Celestia/assets/1612688/f22afc0d-1639-4fea-8ef6-c2e342cc9656)

So I gonna fix this for locations only for now.

SevenSpheres:
[...]

Issue #542, opened 2019-11-29
Frustum cuts part of planet if it's LODSphareMesh


375gnu:
cel://SyncOrbit/Sol:Mars:Curiosity/2019-11-29T08:58:58.22608?x=exzPv8///////////////w&y=HG7FcfX//////////////w&z=R9pp5xQ&ow=-0.823087&ox=0.0355385&oy=-0.560108&oz=-0.0868586&select=Sol:Mars:Curiosity&fov=34.8847&ts=0.5&ltd=0&p=0&rf=8455783187&lm=2&tsrc=0&ver=4

cel://SyncOrbit/Sol:Mars:Curiosity/2019-11-29T08:59:13.71741?x=fda/MxI&y=aVB4ZP7//////////////w&z=apWqbDI&ow=-0.975708&ox=-0.121786&oy=0.150493&oz=-0.10254&select=Sol:Mars:Curiosity&fov=34.8847&ts=0.5&ltd=0&p=0&rf=8455783187&lm=2&tsrc=0&ver=4

![image](https://user-images.githubusercontent.com/1612688/69856732-57c74a80-129f-11ea-9906-e1b782a790a9.png)

375gnu:
Doesn't present in 1.6, but present in all 1.7 versions.

SevenSpheres:
Those URLs don't work in 1.6.1, but it does have the same problem.

375gnu:
@SevenSpheres just change `ver=4` to `ver=3` and they will work in 1.6. Could you also provide URL and screenshots for 1.6.1?

SevenSpheres:
With the same URL changed to `ver=3` I see this:

![issue542](https://user-images.githubusercontent.com/56284207/69904246-9c1b1f00-1361-11ea-8538-d1982e47fd4f.png)

It doesn't look exactly the same, but the surface of Mars disappears at that angle.

375gnu:
hmm, i have correct image, but i don't have textures for Mars installed in 1.6.1, i'll checkwith textures installed

375gnu:
[...]

PR #1808, opened 2023-07-27
Use geodetic coordinates for locations

375gnu:
Closes: #1557

For bodies attached to the surface it's disabled because #542 cuts planet too aggressively, see the screenshot in #1557.

### DIFF
--- a/src/celengine/body.cpp
+++ b/src/celengine/body.cpp
@@ -720,16 +720,15 @@ Quaterniond Body::getEquatorialToBodyFixed(double tdb) const
  */
 Matrix4d Body::getBodyFixedToAstrocentric(double tdb) const
 {
-    //return getEquatorialToBodyFixed(tdb).toMatrix4() * getLocalToAstrocentric(tdb);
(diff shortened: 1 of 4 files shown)


src/celengine/body.cpp
```diff
@@ -746,6 +745,44 @@ Vector3d Body::planetocentricToCartesian(const Vector3d& lonLatAlt) const
...
+ *
+ * Formulae are taken from DOI 10.1007/s00190-011-0514-7.
+ *
+ * @param lon longitude
+ * @param lat latitude
+ * @param alt altitude (height above the surface)
+ *
+ * @return geodetic coordinates
+ */
+Vector3d Body::planetocentricToGeodetic(double lon, double lat, double alt) const
```
ajtribick, line 758: I think this name is confusing - the calculation being done here appears to be the geodetic-to-Cartesian transformation.

ajtribick (changes requested):

ajtribick (approved):

PR #1809, opened 2023-07-28
[qt] Add Celestia Guide to Help Menu

dave-kaye:
This takes a compile time either/or choice for Wiki or Guide in Help menu and instead displays both.

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -1089,12 +1089,14 @@ void CelestiaAppWindow::slotBookmarkTriggered(const QString& url)
 
 void CelestiaAppWindow::slotManual()
 {
-#if 0
     QString MANUAL_FILE = "CelestiaGuide.html";
-    QDesktopServices::openUrl(QUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(QApplication::applicationDirPath()) + QDir::toNativeSeparators(QDir::separator()) + "help" + QDir::toNativeSeparators(QDir::separator()) + MANUAL_FILE)));
-#else
+    QDesktopServices::openUrl(QUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(QDir::currentPath()) + QDir::toNativeSeparators(QDir::separator()) + "help" + QDir::toNativeSeparators(QDir::separator()) + MANUAL_FILE)));
+}
+
+
+void CelestiaAppWindow::slotWiki()
+{
     QDesktopServices::openUrl(QUrl("https://en.wikibooks.org/wiki/Celestia"));
-#endif
 }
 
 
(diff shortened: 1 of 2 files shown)


dave-kaye:
The present code has an either/or choice for the Help menu, displaying the Wiki link or Guide manual link and calling it Manual in the menu. The compile directive if/else is permanently set to select the Wiki link.  This code adds a Qt action and Help menu item so both links are available to the user.

The path for the guide manual is the current directory, which the present code sets to the Celestia install directory in earlier statements. Therefore the path will be correct to select the guide.  If the --dir command line option is set, then the current directory is the --dir directory option.  Therefore the guide file will need to be in the -dir directory.  This is as it should be, since all other items such as fonts, shaders, data, etc are also being called from the --dir directory. If a manual is not found, the code is automatically spawning an internet browser (user's default browser) and the browser will display a "file not found" message. There will be no system crash.

Note: In both cases the full path is "directory/help" as the help subdirectory is added in the code.

The Windows version of Celestia displays only the Guide in the Help menu, plus controls.txt. The GTK version displays neither but displays controls.txt.

src/celestia/qt/qtappwin.cpp
```diff
@@ -1089,12 +1089,14 @@ void CelestiaAppWindow::slotBookmarkTriggered(const QString& url)
 
 void CelestiaAppWindow::slotManual()
 {
-#if 0
     QString MANUAL_FILE = "CelestiaGuide.html";
-    QDesktopServices::openUrl(QUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(QApplication::applicationDirPath()) + QDir::toNativeSeparators(QDir::separator()) + "help" + QDir::toNativeSeparators(QDir::separator()) + MANUAL_FILE)));
-#else
+    QDesktopServices::openUrl(QUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(QDir::currentPath()) + QDir::toNativeSeparators(QDir::separator()) + "help" + QDir::toNativeSeparators(QDir::separator()) + MANUAL_FILE)));
```
375gnu, line 1093:
```suggestion
    QDesktopServices::openUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(QDir::currentPath() + QDir::separator() + "help" + QDir::separator() + "CelestiaGuide.html")));
```

src/celestia/qt/qtappwin.cpp
```diff
@@ -1089,12 +1089,14 @@ void CelestiaAppWindow::slotBookmarkTriggered(const QString& url)
 
 void CelestiaAppWindow::slotManual()
 {
-#if 0
     QString MANUAL_FILE = "CelestiaGuide.html";
```
375gnu, line 1092: remove this line

375gnu (changes requested):

375gnu:
Also it should be fixed to work in Unix-like OSes, e.g. GNU/Linux. I.e. instead of `QDir::currentPath()` a proper directory defined by CMake should be used.

PR #1813, opened 2023-07-31
Replace fmt, miniaudio submodules with CMake ExternalProject

ajtribick:
Detect installed miniaudio header, e.g. from vcpkg or Fedora miniaudio-devel

For now I am leaving CSPICE as a submodule, since the repository we link to does not provide an install target, and in any case I want to replace this with calceph in the near future.


munix9:
FYI: Access to external sources is not possible during the build process on OBS (this may also apply to other build systems).
So with these changes the build will fail.
You will either have to make appropriate packages for each distribution, or include the sources in the build project itself.

ajtribick:
How annoying.

PR #1815, opened 2023-08-01
Fix gaps in draw line as triangles

### LINKED ISSUE #1414: Orbits have gaps when lines are triangulated

levinli303:
Closes #1414

### DIFF
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -145,14 +145,11 @@ constexpr std::string_view NormalVertexPosition = R"glsl(
 constexpr std::string_view LineVertexPosition = R"glsl(
     vec4 thisPos = calc_vp(in_Position);
     vec4 nextPos = calc_vp(in_PositionNext);
-    float w = thisPos.w;
-    thisPos /= w;
-    nextPos /= nextPos.w;
+    thisPos.xy /= thisPos.w;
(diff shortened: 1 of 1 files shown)


levinli303:
i'm not sure why. the only change here is that i do not divide zw by w and then multiply by w again. might it be precision?

In the vertex shader, z and w value are identical so it seems that the z value is changed during the process of dividing and multiplying back, making z/w > 1 thus not rendered

Screenshot 2023-08-01 at 16 04 06

375gnu (approved):

375gnu:
i've completely missed that you multiply it back. in this case it's better to leave it intact.

PR #1824, opened 2023-08-02
[qt] Add Celestia Guide to Help Menu

dave-kaye:
This takes a compile time either/or choice for Wiki or Guide in Help menu and instead displays both.

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -219,8 +219,10 @@ void CelestiaAppWindow::init(const CelestiaCommandLineOptions& options)
         QString dataDir = CONFIG_DATA_DIR;
         QString celestia_data_dir = dataDir;
         QDir::setCurrent(celestia_data_dir);
+        m_celestia_data_dir = QDir::currentPath();
     } else if (QDir(celestia_data_dir).isReadable()) {
         QDir::setCurrent(celestia_data_dir);
+        m_celestia_data_dir = QDir::currentPath();
     } else {
         QMessageBox::critical(0, "Celestia",
             _("Celestia is unable to run because the data directory was not "
@@ -1089,12 +1091,13 @@ void CelestiaAppWindow::slotBookmarkTriggered(const QString& url)
 
 void CelestiaAppWindow::slotManual()
 {
-#if 0
-    QString MANUAL_FILE = "CelestiaGuide.html";
-    QDesktopServices::openUrl(QUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(QApplication::applicationDirPath()) + QDir::toNativeSeparators(QDir::separator()) + "help" + QDir::toNativeSeparators(QDir::separator()) + MANUAL_FILE)));
-#else
+    QDesktopServices::openUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(m_celestia_data_dir) + QDir::toNativeSeparators("/help/CelestiaGuide.html")));
+}
+
+
+void CelestiaAppWindow::slotWiki()
+{
     QDesktopServices::openUrl(QUrl("https://en.wikibooks.org/wiki/Celestia"));
-#endif
 }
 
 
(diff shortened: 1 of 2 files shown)


dave-kaye:
The present code has an either/or choice for the Help menu, displaying the Wiki link or Guide manual link and calling it Manual in the menu. The compile directive if/else is permanently set to select the Wiki link.  This code adds a Qt action and Help menu item so both links are available to the user.

The guide manual is installed in the help subdirectory within the main Celestia data directory. If the --dir command line option is set, then the main Celestia data directory is the user's --dir directory option. After the existing initialization code locates the data directory, this code sets a variable to store the absolute path of the data directory; which remains constant in case some future code or Celx file should change the current directory setting. The guide manual path is derived from this variable. After the data directory is made the current directory by existing code, the QDir function currentPath() is used to guarantee an absolute path, since the variables setting the path could potentially be relative paths.

A CMake file entry for the help directory location will not work due to the command line option to change the data directory location.

This code has been tested with both the original location and --dir data directory option (both absolute and relative paths). It was developed on a Linux system, but the use of Qt QDir structures will make it portable to Windows and macOS. The guide manual is expected to be within the data directory structure, just as fonts, shaders and other items are. If a manual is not found, the existing code is automatically spawning an internet browser (user's default browser) and the browser will display a "file not found" message along with the expected path. There will be no system crash.

note:  The native Windows version of Celestia displays only the Guide in the Help menu, plus controls.txt. The GTK version displays neither but displays controls.txt.

ajtribick:
ui_preferences.h is a generated file, so not surprising it can't find it. The way we've currently got SonarScanner set up will always result in failures on external pull requests, since these cannot access the relevant tokens for uploading the analysis. I think this can be fixed but this will require adjusting the workflows.

src/celestia/qt/qtappwin.h
```diff
@@ -157,6 +158,7 @@ class CelestiaAppWindow : public QMainWindow, public CelestiaCore::ContextMenuHa
     BookmarkToolBar* m_bookmarkToolBar{ nullptr };
 
     QString m_dataDirPath;
+    QString m_celestia_data_dir;
```
375gnu, line 161:
```suggestion
    QString m_dataHome;
```

src/celestia/qt/qtappwin.cpp
```diff
@@ -219,8 +219,10 @@ void CelestiaAppWindow::init(const CelestiaCommandLineOptions& options)
         QString dataDir = CONFIG_DATA_DIR;
         QString celestia_data_dir = dataDir;
         QDir::setCurrent(celestia_data_dir);
+        m_celestia_data_dir = QDir::currentPath();
```
375gnu, line 222:
```suggestion
        m_dataHome = CONFIG_DATA_DIR;
        QDir::setCurrent(m_dataHome);
```

src/celestia/qt/qtappwin.cpp
```diff
@@ -219,8 +219,10 @@ void CelestiaAppWindow::init(const CelestiaCommandLineOptions& options)
...
     } else if (QDir(celestia_data_dir).isReadable()) {
         QDir::setCurrent(celestia_data_dir);
+        m_celestia_data_dir = QDir::currentPath();
```
375gnu, line 225:
```suggestion
        m_dataHome = celestia_data_dir;
        QDir::setCurrent(m_dataHome);
```

src/celestia/qt/qtappwin.cpp
```diff
@@ -1089,12 +1091,13 @@ void CelestiaAppWindow::slotBookmarkTriggered(const QString& url)
 
 void CelestiaAppWindow::slotManual()
 {
-#if 0
-    QString MANUAL_FILE = "CelestiaGuide.html";
-    QDesktopServices::openUrl(QUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(QApplication::applicationDirPath()) + QDir::toNativeSeparators(QDir::separator()) + "help" + QDir::toNativeSeparators(QDir::separator()) + MANUAL_FILE)));
-#else
+    QDesktopServices::openUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(m_celestia_data_dir) + QDir::toNativeSeparators("/help/CelestiaGuide.html")));
```
375gnu, line 1094:
```suggestion
    QDesktopServices::openUrl(QUrl::fromLocalFile(QDir::toNativeSeparators(m_dataHome + "/help/CelestiaGuide.html")));
```

375gnu (changes requested):

PR #1826, opened 2023-08-02
Use unique_ptr for CelestiaCore::config, make getter return const pointer

ajtribick:


src/celestia/configfile.cpp
```diff
@@ -37,96 +37,93 @@ CelestiaConfig* ReadCelestiaConfig(const fs::path& filename, CelestiaConfig *con
...
-        return config;
+        return false;
     }
 
-    if (config == nullptr)
-        config = new CelestiaConfig();
-
-    config->faintestVisible = configParams->getNumber<float>("FaintestVisibleMagnitude").value_or(6.0f);
+    config.faintestVisible = configParams->getNumber<float>("FaintestVisibleMagnitude").value_or(6.0f);
     if (auto path = configParams->getPath("FavoritesFile"); path.has_value())
```
375gnu, line 52: i wonder if c++ ever get macro like lisp?

PR #1827, opened 2023-08-03
Remove the demo key

### LINKED ISSUE #1180: Disable Demo binding
**Is your feature request related to a problem? Please describe.**

Thank you for a demo mode around our Solar System. Let's say I travel through space and accidentally hit D... and you're suddenly near the Earth trying to guess a key how to stop it. This make me frustrated and could potentially turn away from using the application

**Describe the solution you'd like**

Demo mode is nice, and it should be a way in preferences at least to disable demo key. Placing it in Celectia menu make it unbindable even by macOS settings. I'd use this key for another more useful action. 

**Describe alternatives you've considered**
Alternatively, remove binding to demo at all. During normal operation you won't run demo for many times. If a developer actually does, macOS supplies a way to bind a menu in `System Settings.app`.

**Additional context**

ajtribick:
Add "Run Demo" item to File menu in Windows and Qt front ends to replace it.

Resolves #1180


375gnu (approved):
it makes sence

PR #1828, opened 2023-08-03
[ci] Run SonarScanner with Java 17

ajtribick:
Silences the warning in the SonarScanner output.


ajtribick:
Ok I give up for now.

PR #1838, opened 2023-08-06
Split Sonar scanning into separate workflow

### LINKED ISSUE #1761: SonarScanner failing?
Run for #1760 fails with the following error:

```
ERROR: Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator
```

ajtribick:
Testing splitting out into separate workflow so that it doesn't always fail for external PRs

Resolves #1761


ajtribick:
The sonar workflow will only trigger once it is merged to master, so may need to tweak this afterwards.

A question is whether to split out the Ubuntu builds into their own workflow and trigger the Sonar scanner from that, so it doesn't need to wait for the (usually slower) Windows/Mac builds.

375gnu:
> The sonar workflow will only trigger once it is merged to master, so may need to tweak this afterwards.

So, there is no difference with the current state, isn't it?

ajtribick:
> So, there is no difference with the current state, isn't it?

Sorry, miscommunication here. The difference is that the `workflow_run` trigger only runs from a .yml file on the default branch (master), but it should nevertheless trigger for a PR. The workflow therefore gets to access the tokens, since any changes made in a PR do not affect it until it is merged.

The downside is, we can't see it operating in *this* PR either until it is merged to master.

375gnu (approved):

ajtribick:
Made a few changes to ensure the autogenerated files go to the correct location when extracted in the Sonar scanner workflow.

375gnu (approved):

ajtribick:
Ok let's see what happens...

PR #1840, opened 2023-08-08
Use optional instead of infinity

levinli303:


src/celscript/lua/celx_phase.cpp
```diff
@@ -160,10 +157,7 @@ static int phase_getorientation(lua_State* l)
     auto phase = this_phase(l);
 
     double tdb = celx.safeGetNumber(2, WrongType, "Argument to phase:getorientation() must be number", 0.0);
-    if (tdb < phase->startTime())
-        tdb = phase->startTime();
-    else if (tdb > phase->endTime())
-        tdb = phase->endTime();
+    tdb = phase->clamp(tdb);
```
375gnu, line 160: clamp to [0, 1]?

levinli303: this is to clamp between start time and end time of a phase.

levinli303:
> what's wrong with infinity?

i was hoping to enable fastmath some time. frustum.cpp checks isinf which can easily be avoided. i have isinf checks for startTime/endTime of phases in other frontends id rather not replace with checking against 1e9

ajtribick:
I have mixed feelings about `-ffast-math` - some bits of it make a lot more sense than others and I would prefer to rely on the individual sub-flags. Of the ones that aren't switched on by default in C++ on x86_64 (`-fno-signaling-nans`, `-fno-rounding-math`, `-fexcess-precision=fast`), I think `-fno-math-errno` and `-fno-trapping-math` are ones we should probably be using: `errno` for math functions makes sense from the point of view of the C language, which targets machines where floating point representations do not have infinities/NaNs, but is fairly redundant in the case of IEEE floats. `-freciprocal-math` and `-fassociative-math` also seem pretty reasonable, in particular the latter seems to be useful for vectorization.

The `-ffinite-math-only` flag seems much less useful to the point of being potentially dangerous - it makes detecting these cases in user-provided binary-format input files harder (if you've told the system that only finite numbers are present, the compiler is allowed to optimize away checks for these values), and combined with disabling `errno` makes it really hard to detect the results of calculations doing overflows. `-fno-signed-zeros` is perhaps less clear-cut, although enabling it prevents roundtripping of certain CMOD files (including the iss.cmod used in the integration tests) from binary to text to binary again.

ajtribick:
Closing this due to conflicts, and would prefer to use specific fast math optimizations rather than blanket enabling them all.

PR #1845, opened 2023-08-11
Additional sonar.yml fixes

ajtribick:


ajtribick:
this shouldn't affect the build process on freebsd, not proposing to fix that build failure here

375gnu (approved):

PR #1854, opened 2023-08-15
Update unit conversions

### LINKED ISSUE #1825: Incorrect light years ↔ parsecs conversion
The conversion between light years and parsecs is defined here:

https://github.com/CelestiaProject/Celestia/blob/9b8dcfc696dcf8e046d8075c94dde635f14267f2/src/celengine/astro.h#L23-L24

This value should be 3.26156 (to 5 decimal places). This should be fixed, but not sure if that would have implications for existing add-ons.

ajtribick:
- Template-typed values are given to 128 bits (calculated with Sollya)
- Shortest representation of the LN_MAG constant has an additional decimal place

Resolves #1825

### DIFF
--- a/src/celastro/astro.h
+++ b/src/celastro/astro.h
@@ -13,6 +13,7 @@
 #include <cmath>
 #include <cstdint>
 #include <optional>
+#include <type_traits>
 
 #include <Eigen/Core>
 #include <Eigen/Geometry>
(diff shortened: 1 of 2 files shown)


375gnu (approved):

ajtribick:
fixed conflicts caused by refactoring of astro.h into celastro

PR #1855, opened 2023-08-17
[gtk] Fix demo after commit 21f526

dave-kaye:
Commit 21f526 caused demo script not to run in GTK interface. This code calls demo script via action addition duplicating Qt action addition.

The commit also modified user information files to indicate to run demo script from entry in File menu.  GTK runs script from entry in Help menu. This moves/updates submenu entry in GTK.  (Win and Qt use File menu.)

This code also removes the D key indicator on the demo menu item.

(note: During testing, new celestiaui.xml file must replace version in main data directory in addition to compiling GTK executable.)

### DIFF
--- a/src/celestia/gtk/actions.cpp
+++ b/src/celestia/gtk/actions.cpp
@@ -346,6 +346,18 @@ void actionCaptureMovie(GtkAction*, AppData* app)
 }
 
 
+/* File -> Run Demo... */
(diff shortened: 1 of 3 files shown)


dave-kaye:
Code failing unexpectedly after new commit 4c52c60.  Will look at issue later.

PR #1859, opened 2023-08-19
Use star color to tint atmospheres

ajtribick:
I merged the two sets of light parameters in the shaders (`FragLightParameter(i, "name")`/`LightParameter(i, "name")`) as they appeared to be doing the same thing (first commit).

The second commit wires up the light source color to the atmosphere scattering calculations.

### DIFF
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -820,6 +820,8 @@ DeclareLights(const ShaderProperties& props)
         }
         if (props.texUsage & ShaderProperties::NightTexture)
             stream << DeclareUniform(fmt::format("light{}_brightness", i), Shader_Float);
+        if ((props.texUsage & ShaderProperties::AtmosphereModel) || props.hasScattering())
+            stream << DeclareUniform(fmt::format("light{}_color", i), Shader_Vector3);
     }
 
 #else
@@ -1181,7 +1183,7 @@ AtmosphericEffects(const ShaderProperties& props)
     std::string scatter;
     if (hasAbsorption)
     {
-        source += "    vec3 sunColor = exp(-extinctionCoeff * density * distSun);\n";
+        source += "    vec3 sunColor = " + LightProperty(0, "color") + " * exp(-extinctionCoeff * density * distSun);\n";
         source += "    vec3 ex = exp(-extinctionCoeff * density * distAtm);\n";
 
         scatter = "(1.0 - exp(-scatterCoeffSum * density * distAtm))";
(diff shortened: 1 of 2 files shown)


src/celengine/shadermanager.cpp
```diff
@@ -826,6 +820,8 @@ DeclareLights(const ShaderProperties& props)
         }
         if (props.texUsage & ShaderProperties::NightTexture)
             stream << DeclareUniform(fmt::format("light{}_brightness", i), Shader_Float);
+        if ((props.texUsage & ShaderProperties::AtmosphereModel) || props.hasScattering())
```
375gnu, line 823: `#else` branch should also have this change

src/celengine/shadermanager.cpp
```diff
@@ -2431,19 +2415,19 @@ ShaderManager::buildFragmentShader(const ShaderProperties& props)
...
             if ((props.texUsage & ShaderProperties::NightTexture) != 0)
-                source += "totalLight += l * " + FragLightProperty(i, "brightness") + ";\n";
+                source += "totalLight += l * " + LightProperty(i, "brightness") + ";\n";
 
             if (props.hasShadowsForLight(i))
                 source += ShadowsForLightSource(props, i);
 
             std::string illum(props.hasShadowsForLight(i) ? "l * shadow" : "l");
-            source += "diff.rgb += " + illum + " * " + FragLightProperty(i, "color") + ";\n";
+            source += "diff.rgb += " + illum + " * " + LightProperty(i, "diffuse") + ";\n";
```
375gnu, line 2424: why do you replace color with diffuse?

375gnu (changes requested):

src/celengine/shadermanager.cpp
ajtribick: Because this replicates the same behavior as the original version. The previous code used both `fragLightColor[i]` and `lights[i].diffuse` to store the quantity `lightColor.cwiseProduct(diffuseColor)`

See L3672/3681 at https://github.com/CelestiaProject/Celestia/pull/1859/files#diff-69cd8d8749fa1d3212a10624731fe412e2afaaed33931085c3427a3009a981acL3667-L3682

There was not originally a `color` property on the lights, this is a new addition to provide the unchanged light color to the atmosphere shader.

src/celengine/shadermanager.cpp
ajtribick: Done. Also inverted `#ifndef...#else` to `#ifdef...#else` as it reads better

375gnu (approved):

PR #1869, opened 2023-08-24
Use ICU to format time when available

levinli303:


src/celengine/dateformatter.cpp
```diff
@@ -0,0 +1,132 @@
...
+            pattern = u"yyyy MMM dd HH:mm:ss ZZ";
+            dateStyle = UDAT_PATTERN;
+            timeStyle = UDAT_PATTERN;
+            break;
+        }
+        UErrorCode error = U_ZERO_ERROR;
+        formatter = udat_open(timeStyle, dateStyle, locale, local ? nullptr : u"UTC", -1, pattern, -1, &error);
+        if (U_FAILURE(error))
+            return nullptr;
+        formatters.insert(std::pair<astro::Date::Format, UDateFormat*>(format, formatter));
```
375gnu, line 126:
```suggestion
        formatters.insert(std::make_pair(format, formatter));
```

src/celengine/dateformatter.cpp
```diff
@@ -0,0 +1,132 @@
...
+    {
+        static const char *locale = nullptr;
+        if (locale == nullptr)
+        {
+            const char *orig = N_("LANGUAGE");
+            const char *lang = _(orig);
+            if (lang == orig)
+                locale = "en";
+            else
+                locale = lang;
```
375gnu, line 94:
```suggestion
            locale = lang == orig ? "en" : lang;
```

src/celengine/dateformatter.cpp
```diff
@@ -0,0 +1,132 @@
...
+    auto iter = formatters.find(format);
+    UDateFormat *formatter = nullptr;
+    if (iter != formatters.end())
+    {
+        formatter = iter->second;
+    }
+    else
+    {
+        static const char *locale = nullptr;
+        if (locale == nullptr)
```
375gnu, line 87:
```suggestion
        if (static const char *locale = nullptr; locale == nullptr)
```

src/celengine/dateformatter.cpp
```diff
@@ -0,0 +1,132 @@
...
+#endif
+}
+
+#ifdef USE_ICU
+UDateFormat *DateFormatter::getFormatter(bool local, astro::Date::Format format)
+{
+    auto& formatters = local ? localFormatters : utcFormatters;
+    auto iter = formatters.find(format);
+    UDateFormat *formatter = nullptr;
+    if (iter != formatters.end())
```
375gnu, line 80:
```suggestion
    if (static const char *locale = nullptr; iter != formatters.end())
```

src/celengine/dateformatter.cpp
```diff
@@ -0,0 +1,132 @@
...
+    astro::Date d = local ? astro::TDBtoLocal(tdb) : astro::TDBtoUTC(tdb);
+    return d.toCStr(format);
+#endif
+}
+
+#ifdef USE_ICU
+UDateFormat *DateFormatter::getFormatter(bool local, astro::Date::Format format)
+{
+    auto& formatters = local ? localFormatters : utcFormatters;
+    auto iter = formatters.find(format);
```
375gnu, line 78:
```suggestion
```

src/celengine/dateformatter.h
```diff
@@ -0,0 +1,53 @@
...
+    DateFormatter(const DateFormatter &) = delete;
+    DateFormatter(DateFormatter &&) = default;
+    DateFormatter &operator=(const DateFormatter &) = delete;
+    DateFormatter &operator=(DateFormatter &&) = delete;
+
+    std::string formatDate(double tdb, bool local, astro::Date::Format format);
+
+#ifdef USE_ICU
+private:
+    std::unordered_map<astro::Date::Format, UDateFormat*> localFormatters;
```
375gnu, line 46: Does it make sense to use maps if all we need is to store 4 pairs? Maybe std::array will be enough? std::unordered_map size is 56 (gcc, clang) or 64 (msvc) bytes (amd64). 4 element array with pair members aligned at 8 bytes boundaries requires the same 64 bytes.

375gnu (changes requested):

src/celengine/dateformatter.h
levinli303: i've changed to `std::array<UDateFormat*, static_cast<size_t>(astro::Date::ISO8601) + 1>` but i wonder if we should add a Count case instead.

375gnu: Yes, it's better to add `astro::Date::Format::FormatCount` rather then rely on ISO8601 as the last element

src/celengine/dateformatter.h
```diff
@@ -0,0 +1,53 @@
...
+    DateFormatter(const DateFormatter &) = delete;
+    DateFormatter(DateFormatter &&) = default;
+    DateFormatter &operator=(const DateFormatter &) = delete;
+    DateFormatter &operator=(DateFormatter &&) = delete;
+
+    std::string formatDate(double tdb, bool local, astro::Date::Format format);
+
+#ifdef USE_ICU
+private:
+    std::array<UDateFormat*, static_cast<size_t>(astro::Date::ISO8601) + 1> localFormatters;
```
ajtribick, line 46: I think it would be better to define a `COUNT` member of the enum and reference that - would be less brittle in case someone adds/reorders date formats.

375gnu (approved):

PR #1872, opened 2023-08-26
Don't include Windows.h from glsupport.h

### LINKED ISSUE #1871: Windows ICU compilation errors due to Windows.h/icu.h conflict
There appears to be a conflict between Windows.h and icu.h that occurs if you attempt to import icu.h *after* Windows.h - this causes compilation to fail since the date formatter changes when using `-DUSE_ICU:BOOL=TRUE -DUSE_WIN_ICU:BOOL=TRUE`

This occurs at several places in the codebase due to the inclusion of Windows.h from glsupport.h: https://github.com/CelestiaProject/Celestia/blob/d2c93d2abb5a5910498be8727d70f3716c26e5c0/src/celengine/glsupport.h#L7

Do we need the Windows.h inclusion there? Celestia *seems* to work without it.

ajtribick:
Fixes a conflict between Windows-supplied icu.h and Windows.h, caused if dateformatter.h is included after glsupport.h.

Also replace winglext.h/.cpp with epoxy

Resolves #1871


375gnu:
> Also replace winglext.h/.cpp with epoxy

This needs testing.

ajtribick:
SevenSpheres mentioned in Discord that it was working.

375gnu (approved):

Issue #1771, opened 2023-07-13
Qt - Full Screen toggle key not working


dave-kaye:
Help files and key documentation state that the alt+enter key combination will toggle full screen mode.  These keys work on 1.7.0 interfaces GTK and SDL, and probably on Windows.  However, they do not work on Qt interface.

Qt interface seems to have changed the keys to shift+F11.  These keys work.

Was there a reason for the key change in Qt?

It would be preferable to have the same keys in all interfaces, as they previously were.  The Qt interface then needs to be fixed. If there is a reason not to use the alt+enter sequence, then a new common key sequence should be utilized.  This will touch GTK, SDL, Windows, Help and control.txt areas.

I do not know if this is an issue that affects the mobile versions of Celestia.

PR #1874, opened 2023-08-29
[qt] Full screen toggle to common ALT + Enter  Issue #1771

dave-kaye:
Utilize the ALT + Enter key sequence to toggle full screen mode which is common among the Win32, SDL and Gtk interfaces, and specified in the Help manual and controls.txt file.

Closes: #1771

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -415,7 +415,12 @@ void CelestiaAppWindow::init(const CelestiaCommandLineOptions& options)
 
     QAction* fullScreenAction = new QAction(_("Full screen"), this);
     fullScreenAction->setCheckable(true);
-    fullScreenAction->setShortcut(QString(_("Shift+F11")));
+
+    // Qt defines Key_Return as enter key on keyboard and Key_Enter as enter key on numeric keypad
+    // Capture both keys
+    QList<QKeySequence> shortcuts;
+    shortcuts << QKeySequence(_("ALT+Enter")) << QKeySequence(_("ALT+Return"));
+    fullScreenAction->setShortcuts(shortcuts);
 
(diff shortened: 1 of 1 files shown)


dave-kaye:
KDE3 interface specifies full screen mode with two key combinations,` CTRL + Key_F` and `ALT + Key_Return`.  When it was converted to Qt it became `Shift + F11`, reason unknown.  All the other interfaces in 1.7.0 utilize the `ALT + Enter` key sequence, and it is specified in the help files. This submission returns the full screen toggle to the common key sequence. Qt considers the enter keys on the main keyboard and numeric keypad as separate keys, so both must be captured.

This was tested on Qt5. It should readily work on Qt6.  I do not know what the effect is on mobile versions (@levinli303).  Do not think there is an issue on macOS.  If there is an issue, then the other interfaces should move to the Shift + F11 key sequence to keep it common.

Note: Gtk captures the keyboard enter key, but not the numeric keypad enter key for full screen (bug). It does capture both for hand entering (typing in) objects.  SDL uses only the keyboard enter key for both functions, it misses the numeric keypad enter key.  Win32 uses both the keyboard and numeric keypad enter keys for both functions.

src/celestia/qt/qtappwin.cpp
```diff
@@ -415,7 +415,12 @@ void CelestiaAppWindow::init(const CelestiaCommandLineOptions& options)
 
     QAction* fullScreenAction = new QAction(_("Full screen"), this);
     fullScreenAction->setCheckable(true);
-    fullScreenAction->setShortcut(QString(_("Shift+F11")));
+
```
375gnu, line 418:
```suggestion
```

375gnu (approved):

375gnu:
Please use `squash and merge`

PR #1876, opened 2023-09-02
Remove unused m_name field from Url class

ajtribick:
This field was only written, never read.


375gnu (approved):
perhaps remnant of the old url versions

PR #1881, opened 2023-09-05
Refactor the Qt interface

ajtribick:
- Clean up includes
- Remove using namespace std
- Put local classes into unnamed namespace
- Nest abstract item classes in parents

Also fix includes for the eclipsefinder base class in src/celestia


ajtribick:
Not going to fix the Sonar whinging in this PR... enough changes here already

375gnu (approved):

PR #1883, opened 2023-09-07
Use smart pointers in virtual textures

ajtribick:
Will be useful in case we get round to implementing unloading

### DIFF
--- a/src/celengine/virtualtex.cpp
+++ b/src/celengine/virtualtex.cpp
@@ -7,298 +7,40 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-#include <cmath>
+
+#include "virtualtex.h"
+
 #include <cassert>
-#include <cmath>
-#include <istream>
+#include <cstdio>
 #include <fstream>
-#include <string>
+#include <optional>
+#include <system_error>
 #include <utility>
+
 #include <fmt/format.h>
-#include <celcompat/filesystem.h>
+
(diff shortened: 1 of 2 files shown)


src/celengine/virtualtex.cpp
```diff
@@ -244,147 +318,65 @@ void VirtualTexture::populateTileTree()
...
-                {
-                    if (u >= 0 && v >= 0 && u < uLimit && v < vLimit)
-                    {
-                        // Found a tile, so add it to the quadtree
-                        Tile* tile = new Tile();
-                        addTileToTree(tile, maxLevel, (unsigned int) u, (unsigned int) v);
-                    }
-                }
-            }
+            if (!d.is_regular_file(ec))
```
ajtribick, line 334: Hopefully no-one is using symlinks in their virtual textures, not sure if we would want to support that.

src/celengine/virtualtex.cpp
```diff
@@ -244,147 +318,65 @@ void VirtualTexture::populateTileTree()
     // Count the number of resolution levels present
     unsigned int maxLevel = 0;
 
-    // Crash potential if the tile prefix contains a %, so disallow it
-    string pattern;
-    if (tilePrefix.find('%') == string::npos)
```
ajtribick, line 249: We check the prefix outside the sscanf call, so no need to do this check

375gnu (approved):
please update commit title to be more precise

PR #1885, opened 2023-09-08
Update jquery to 3.7.1

ajtribick:
Fix Dependabot warning by upgrading jquery.

Ideally though we should transition to HTML5


375gnu:
Does it work? When I tried to do the same, a lot of errors were thrown to the browser console.

munix9:
Probably it also needs a more up-to-date "animatedcollapse.js":

http://www.dynamicdrive.com/dynamicindex17/animatedcollapse.htm
http://www.dynamicdrive.com/dynamicindex17/animatedcollapse.js

PR #1888, opened 2023-09-09
Replace constellation.h prefix search with radix tree

ajtribick:
- Precompute radix tree data and store in constexpr array to avoid heap allocation
- Minor refactoring of the interface


375gnu (approved):

375gnu:
The less heap is used the better! )

PR #1890, opened 2023-09-11
Repair guided tour

dave-kaye:
Repair guided tour so view of object is at expected distance.  Update tour destinations file.


dave-kaye:
Update win dialog size on new commit

PR #1892, opened 2023-09-12
Repair guided tour

dave-kaye:
Repair guided tour so view of object is at expected distance.


dave-kaye:
Interfaces win32 and GTK have a menu item "Tour Guide"; Qt does not.  This item displays a selectable list of celestial objects. In version 1.7 the links do not show the objects. The problem was traced to improper distance settings of the gotoSelection code. The distances were not as requested, allowing the view to be too distant, sometimes so close as to be inside the object, and sometimes a negative distance that interfered with other Celestia navigational methods.

The code reads a guide.cel file that allows distances to be in kilometers, astronomical units or light years. The code had a comment that all distances were converted into light years. Logging results showed the conversion was correctly performed. Other Celestia goto codes all convert distances to kilometers. All the codes feed into one common routine for the actual view movement.  I cannot find any light year to kilometer conversion in the tour guide codes. Many of the selected items displayed as if the numbers they were sent were at kilometer distances, despite being sent light year distances.

This code changes the converted distance units to kilometers for the tour guide code. It has been tested on GTK3. It is part of the core Celestia code, so should function the same for the win32 interface. (Presently, the win32 interface and GTK interface give identical results for tour guide selections - all matching errors.)

The actual tour is performed by reading the file guide.cel in the install directory. This file has not been updated in 16 years. This work updates the file to add several items/areas of interest that have been active since then. It also removes several items that are not supported by the present core data files.  Specifically, Albireo is described as a double star, but displays as a single star. The Hyades and Pleiades have more bright stars in photographs than are being displayed, so a check of the data files is in order - but they remain in the tour. Proxima Centauri cannot be seen as part of the Alpha Centauri star system. Additional items of interest can be added to the tour, but the data files do not support them. The work was confined to items in the common data files, not addon data which the user may not have.

Additional information:

- The guide and favorites files are *.cel files, in-spite of the fact they do not follow the scripting structure.  Celestia will attempt to read them as cel files in other sections of the code. Different file extensions or changing the file structure to something like xml and then changing file names should be considered.

- The guide does not allow for initial positions before going to the object. Thus the object will appear different depending on the path taken to get to it, and the actual time of the simulation. It may be advantageous in the future to remove the guide and substitute scripts that give better control. But improved display of text information is needed before then. Perhaps a subroutine to create a text box directly in the openGL window is possible. The tour guide does have an advantage in that it is readily user customizable, without learning difficult scripting languages.

375gnu:
@dave-kaye you don't need to recreate PRs, use the same. we can squash commits while merging, so the history will be clean

dave-kaye:
Tried to reuse PR, but update caused all changes from main celestia master to load into my directory as user modifications in addition to my modifications.  Uncertain as to what will happen if I used that fork and if it would submit the main celestia master modifications again with mine , I bailed out and started fresh.

Purpose of change is win32 is fixed dialog window, gtk is dynamic.  The guide info was being cut off by win32 when space ran out.  Don't have Windows for compiling, so waited until github artifacts to check out win32 and found issue.  Enlarging window and decreasing verbiage on longest info sections.

dave-kaye:
Changes made now show complete info in Windows.

dave-kaye:
Updated code.  Found reference to tour guide in Seldon Ball website on webpage that I believe was made at least before 2007 - but it is still active.  Stated that units "km" and "au" are read, all others are assumed to be light years.  The initial code did not take this into account, This code now converts the distance to km if no units are present or if an invalid unit, i.e. "miles", is present - assuming the distance is in light years.  This matches the instruction webpage.  It will prevent any user built guide.cel files from viewing from wrong distance if they fail to explicitly state the distance units.  If both distance and units are absent, it uses the Celestia default distance calculations - I did not touch that section of the code.  Unfortunately, easiest way to work was to keep nested ifs, so SonarCloud will error.  Otherwise it will probably require a rewrite of the original code logic.

Put new code in forked branch.  Hope this means you can squash merge on single PR if you accept the code.  Did not follow github suggestion to update branch which appears to make everything cleaner and work better.

375gnu (changes requested):

dave-kaye:
Yes,  Need to get km response out of the input string.  No action needed for it.  This allows all the other inputs of bad units and misc stuff to flow into ELSE statement and assumed to be light years for conversion - per the user instructions that anything other than km and au are assumed to be light years.

I know it looks dumb, but it works.

src/celestia/destination.cpp
```diff
@@ -57,23 +57,36 @@ DestinationList* ReadDestinationList(std::istream& in)
...
+                dest->target = *target;
+                if (const std::string* description = destParams->getString("Description"); description != nullptr)
+                    dest->description = *description;
+                if (auto distance = destParams->getNumber<double>("Distance"); distance.has_value())
+                    dest->distance = *distance;
+                // Convert distance to goto distance of kilometers
+                // User default unit of distance is the light year
+                if (const std::string* distanceUnits = destParams->getString("DistanceUnits"); distanceUnits != nullptr)
+                {
+                    if (!compareIgnoringCase(*distanceUnits, "ly"))
```
375gnu, line 76:
```suggestion
                    if (compareIgnoringCase(*distanceUnits, "ly") == 0)
```

src/celestia/destination.cpp
```diff
@@ -57,23 +57,36 @@ DestinationList* ReadDestinationList(std::istream& in)
...
+                        dest->distance = astro::lightYearsToKilometers(dest->distance);
+                    else if (!compareIgnoringCase(*distanceUnits, "au"))
```
375gnu, line 78:
```suggestion
                    else if (compareIgnoringCase(*distanceUnits, "au") == 0)
```

src/celestia/destination.cpp
```diff
@@ -57,23 +57,36 @@ DestinationList* ReadDestinationList(std::istream& in)
...
+                        dest->distance = astro::AUtoKilometers(dest->distance);
+                    else if (!compareIgnoringCase(*distanceUnits, "km"));
+                    else
+                        dest->distance = astro::lightYearsToKilometers(dest->distance);
```
375gnu, line 82:
```suggestion
                    else if (compareIgnoringCase(*distanceUnits, "km") ! = 0)
                        dest->distance = astro::lightYearsToKilometers(dest->distance);
```

375gnu (changes requested):

375gnu:
@dave-kaye instead of empty `if`
```
if (foo); else dosmth();
```

you should use inverted condition:
```
if (!foo) dosmth();
```

src/celestia/destination.cpp
375gnu: don't recreate PR, use "Commit suggestion"

src/celestia/destination.cpp
```diff
@@ -69,13 +69,21 @@ DestinationList* ReadDestinationList(std::istream& in)
...
-                // Default unit of distance is the kilometer
+                // Convert distance to goto distance of kilometers
+                // User default unit of distance is the light year
                 if (const std::string* distanceUnits = destParams->getString("DistanceUnits"); distanceUnits != nullptr)
                 {
                     if (!compareIgnoringCase(*distanceUnits, "ly"))
                         dest->distance = astro::lightYearsToKilometers(dest->distance);
                     else if (!compareIgnoringCase(*distanceUnits, "au"))
                         dest->distance = astro::AUtoKilometers(dest->distance);
+                    else if (!compareIgnoringCase(*distanceUnits, "km"));
```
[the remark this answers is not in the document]
dave-kaye: changes committed

dave-kaye:
Committed changes.

for

If then
else

Figured out the inversion after I got home and never got back to make the changes.  Thanx.

dave-kaye:
@375gnu Guess it didn't like your suggested changes - all compiles failed.  I think on line 80 you put ! = and left a space between ! and = which may be throwing compiler off.  I cannot fix, if you can then update immediately and don't involve me if possible.  I will probably be gone for several days this weekend.  I don't see anything else that would be triggering the error message.

dave-kaye:
They say any day you learn something is not a wasted day.  So today is not wasted.  I learned that committed suggestions will show back on the original forked branch.

So I can remove space between ! and =.   Now compiling is successful.

375gnu:
@dave-kaye let's split this PR into two. 1st is for code and updated distance in guide.cel, please also update locale/guide_*.cel. and 2nd is for guide.cel content. we should also update translations, so we cannot simply merge it as-is.

dave-kaye:
Removed changes from guide.cel, replaced with original version.  Only code fixes are now active.  Hopefully github web interface is keeping the guide.cel intact and will not modify or change label on original file commit from 16 years ago.

Distance changes are minor on listed objects, not worth modifying all the files.

375gnu (approved):

PR #1894, opened 2023-09-12
Use gperf for legacy scripts lookup table

ajtribick:
Switch lookup table to static lookup/perfect hashing.

Do we still need .travis.yml?


ajtribick:
Note: this requires changes to the Sonar workflow which won't be picked up until it's on the master branch

375gnu:
We can remove both .appveyor.yml and .travis.yml. We use only github action and circleci (for freebsd).

375gnu (approved):

PR #1899, opened 2023-09-15
Surround objc calls with autoreleasepool

levinli303:
This ensures resources allocated in objc are correctly freed when the method is called not in Apple's system runloop.


375gnu (approved):
u know what u'r doing )

PR #1909, opened 2023-09-22
Update guided tour script

dave-kaye:
Update tour destinations file.

### DIFF
--- a/guide.cel
+++ b/guide.cel
@@ -1,51 +1,47 @@
 {
 	Name   "Jupiter"
 	Target "Sol/Jupiter"
-	Description "Jupiter is the largest planet in our solar system and the fifth from the sun.  Like the other large outer planets, Jupiter is a gas giant, with no solid surface.  The Great Red Spot is the biggest and longest-lived of the many storms in Jupiter's turbulent atmosphere; this Earth-sized tempest has been around for at least 300 years."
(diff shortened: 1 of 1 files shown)


dave-kaye:
Tour Guide is a menu item in GTK, win32 and user documentation.  It allows various celestial objects to be displayed at the user's request.

The actual tour is performed by reading the file guide.cel in the install directory. This file has not been updated in 16 years. This work updates the file to add several items/areas of interest that have been active since then. It also removes several items that are not supported by the present core data files.  Specifically, Albireo is described as a double star, but displays as a single star. The Hyades and Pleiades have more bright stars in photographs than is being displayed, so a check of the data files is in order - but they remain in the tour. Proxima Centauri cannot be seen as part of the Alpha Centauri star system. Additional items of interest can be added to the tour, but the data files do not support them. The work was confined to items in the common data files, not addon data which the user may not have.

This is a suggested edit of the tour guide.  Other potential objects (when properly documented in common data files) are:

TRAPPIST-1f (earth-sized exoplanet in habitable zone of star)

PSR B1257+12 B and C (first exoplanets discovered)

Churyumov-Gerasimenko/67P (comet was location of first spacecraft to land on comet - Rosetta/Philae; comet has unique structure and will require model)

This is a draft/suggestion if others want to edit or add/subtract items.

Note:  The locale files are 16 years old and appear to be hand translations - as do all the files in the locale directory.  Some effort would be needed in translating if they are to be kept up-to-date.  If file is changed, should be single commit to prevent constant translation requests.

dave-kaye:
The multiple commits are due to testing the object distance on multiple systems and user interfaces.  They all perform somewhat differently.  This is the best compromise I have found.  It just affects how large the object is displayed, and zoom features can be readily used to change the distance/object size.

Please merge/flatten all the commits if this request is added to the Celestia source.

PR #1912, opened 2023-09-25
Use smart pointers in SolarSystem/SolarSystemCatalog

ajtribick:
- Clarify ownership of solar system objects
- Prefer using the getSolarSystem(const Star*) overload
- Use lower_bound/emplace_hint when creating solar systems to avoid duplicate lookup

### DIFF
--- a/src/celengine/universe.cpp
+++ b/src/celengine/universe.cpp
@@ -672,10 +672,9 @@ Universe::getSolarSystem(const Star* star) const
 
     auto starNum = star->getIndex();
     auto iter = solarSystemCatalog->find(starNum);
-    if (iter != solarSystemCatalog->end())
-        return iter->second;
-
-    return nullptr;
+    return iter == solarSystemCatalog->end()
+        ? nullptr
+        : iter->second.get();
 }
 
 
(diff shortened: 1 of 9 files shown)


src/celengine/universe.cpp
```diff
@@ -717,16 +716,13 @@ Universe::getSolarSystem(const Selection& sel) const
...
-                               value_type(star->getIndex(),
-                                          solarSystem));
+    auto starNum = star->getIndex();
+    auto iter = solarSystemCatalog->lower_bound(starNum);
+    if (iter != solarSystemCatalog->end() && iter->first == starNum)
+        return iter->second.get();
 
-    return solarSystem;
+    iter = solarSystemCatalog->emplace_hint(iter, starNum, std::make_unique<SolarSystem>(star));
+    return iter->second.get();
```
375gnu, line 725: please comment this change in pr description

375gnu (approved):

src/celengine/universe.cpp
ajtribick: done

PR #1921, opened 2023-10-02
Use smart pointers in the body.h classes

ajtribick:
- Better traceability of resource ownership
- Ensure planetary systems get cleaned up on destruction
- Prevent creating rings if inner/outer radii are not specified
- Use transform views to return locations, reference marks and alternate surface names

### DIFF
--- a/src/celengine/body.cpp
+++ b/src/celengine/body.cpp
@@ -944,20 +944,6 @@ void Body::addAlternateSurface(const string& name, std::unique_ptr<Surface>&& al
 }
 
 
-vector<string>* Body::getAlternateSurfaceNames() const
-{
-    vector<string>* names = new vector<string>();
-    if (altSurfaces)
-    {
-        std::transform(altSurfaces->begin(), altSurfaces->end(),
-                       std::back_inserter(*names),
-                       [](const auto& s) { return s.first; });
-    }
-
-    return names;
-}
-
-
 void Body::addLocation(std::unique_ptr<Location>&& loc)
 {
     if (!loc)
@@ -967,12 +953,6 @@ void Body::addLocation(std::unique_ptr<Location>&& loc)
         locations = std::make_unique<std::vector<std::unique_ptr<Location>>>();
     loc->setParentBody(this);
     locations->push_back(std::move(loc));
-}
-
-
-const std::vector<std::unique_ptr<Location>>* Body::getLocations() const
-{
-    return locations.get();
 }
 
 
@@ -1003,6 +983,9 @@ void Body::computeLocations()
         return;
 
     locationsComputed = true;
+
+    if (locations == nullptr)
+        return;
 
     // No work to do if there's no mesh, or if the mesh cannot be loaded
     if (geometry == InvalidResource)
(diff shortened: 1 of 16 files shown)


src/celengine/body.h
```diff
@@ -306,11 +304,11 @@ class Body
     void getLifespan(double&, double&) const;
 
     Surface* getAlternateSurface(const std::string&) const;
-    void addAlternateSurface(const std::string&, Surface*);
+    void addAlternateSurface(const std::string&, std::unique_ptr<Surface>&&);
     std::vector<std::string>* getAlternateSurfaceNames() const;
 
-    std::vector<Location*>* getLocations() const;
-    void addLocation(Location*);
+    const std::vector<std::unique_ptr<Location>>* getLocations() const;
```
375gnu, line 310: maybe `util::array_view` is better suitable here

src/celengine/body.cpp
```diff
@@ -1075,48 +1054,46 @@ Body::addReferenceMark(ReferenceMark* refMark)
...
+        ? nullptr
+        : iter->get();
 }
 
 
 /*! Get the list of reference marks associated with this body. May return
  *  nullptr if there are no reference marks.
  */
-const list<ReferenceMark*>*
+const std::list<std::unique_ptr<ReferenceMark>>*
```
375gnu, line 1093: exposing internal implementation sounds like an anti-pattern. perhaps it's better to to return a  new list containing raw pointers

src/celengine/render.cpp
```diff
@@ -1928,86 +1928,90 @@ Renderer::locationsToAnnotations(const Body& body,
...
+            hit && t < 1.0)
+        {
+            continue;
         }
+
+        // Calculate the intersection of the eye-to-label ray with the plane perpendicular to
+        // the view normal that touches the front of the object's bounding sphere
+        double planetZ = viewNormal.dot(bodyCenter) - boundingRadius;
+        if (planetZ < -nearDist * 1.001)
+            planetZ = -nearDist * 1.001;
```
375gnu, line 1992:
```suggestion
        double planetZ = std::max(viewNormal.dot(bodyCenter) - boundingRadius, -nearDist * 1.001);
```

375gnu (changes requested):

src/celengine/body.cpp
ajtribick: Another possibility I was considering here is to bring in range-v3 and expose this as a transform view on the inner collection. The actual type name would probably still expose the implementation details, but the consuming code would not have to interact with the unique_ptr. (And the dependency could go away in the distant future when we can upgrade to C++20)

ajtribick: Actually scratch that, ranges-v3 does not build with exceptions disabled, and the author has stated he is not interested in supporting non-standard C++ dialects.

src/celengine/body.h
ajtribick: Replaced with a transform view that returns raw pointers rather than smart pointers.

src/celengine/body.cpp
ajtribick: Added a transform view implementation. Bonus is that the `getAlternateSurfaceNames` method no longer needs to allocate a vector every time it is called.

ajtribick:
I plan to squash the first two commits before merging this.

375gnu (approved):

PR #1923, opened 2023-10-05
Disable -Weffc++ by default

ajtribick:
It creates too much noise in the build process, making it hard to find compilation errors

Can re-enable with USE_WEFFCPP


375gnu (changes requested):
better hide it behind a cmake option (ENABLE_WEFF or smth like this)

375gnu (approved):

PR #1938, opened 2023-10-15
[celx] Implement object:setatmosphere with a table instead of 22 numbers

375gnu:
\+ small cleanup in object:mark.

I doubt that anybody really used method with 22 arguments, so I hope it's safe to change API.

### DIFF
--- a/src/celscript/lua/celx_object.cpp
+++ b/src/celscript/lua/celx_object.cpp
@@ -781,25 +781,18 @@ static int object_mark(lua_State* l)
     if (colorString != nullptr)
         Color::parse(colorString, markColor);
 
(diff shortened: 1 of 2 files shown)


375gnu:
I also don't like non-const `Body::getAtmosphere()`. I'd prefer `const` version only and instead of direct edit of a body's atmosphere do thee following:
```c++
Atmosphere atmosphere;
if (body->getAtmosphere() != nullptr)
    atmosphere = *body->getAtmosphere();
...
body->setAtmosphere(atmosphere);
```

Any thoughts?

ajtribick:
I did originally have a version that replaced the atmosphere instead of editing it, but I'm not sure if anyone's using this to do animations. It's also not clear to me why this method doesn't allow creating an atmosphere if one doesn't exist.

ajtribick (approved):

PR #1943, opened 2023-10-18
Rename astro namespace to celestia::astro

ajtribick:
- Tidy up includes in astro.h/astro.cpp
- Replace snprintf with fmt in astro.cpp

### DIFF
--- a/src/celengine/astro.cpp
+++ b/src/celengine/astro.cpp
@@ -8,29 +8,42 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-#include <algorithm>
-#include <cmath>
-#include <ostream>
-#include <utility>
-#include <ctime>
 #include "astro.h"
-#include "univcoord.h"
-#include <celcompat/numbers.h>
-#include <celutil/gettext.h>
+
+#include <algorithm>
+#include <array>
+#include <cassert>
+#include <cstddef>
+#include <cstdio>
+#include <cstdlib>
+#include <ctime>
+#include <locale>
+#include <string_view>
+
+#if defined(__GNUC__) && !defined(_WIN32)
+#include <fmt/chrono.h>
+#endif
+#include <fmt/format.h>
+
 #include <celmath/geomutil.h>
+#include <celutil/gettext.h>
+
+using namespace std::string_view_literals;
 
 
-using namespace Eigen;
-using namespace std;
-using namespace celmath;
+namespace celestia::astro
+{
 
-static const Quaterniond ECLIPTIC_TO_EQUATORIAL_ROTATION = XRotation(-astro::J2000Obliquity);
-static const Matrix3d ECLIPTIC_TO_EQUATORIAL_MATRIX = ECLIPTIC_TO_EQUATORIAL_ROTATION.toRotationMatrix();
+namespace
+{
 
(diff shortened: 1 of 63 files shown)


src/celengine/astro.cpp
```diff
@@ -321,15 +373,19 @@ astro::Date::Date(double jd)
     tzname = "UTC";
 }
 
-const char* astro::Date::toCStr(Format format) const
+
+const char*
+Date::toCStr(Format format) const
 {
-    static char date[255];
+    static std::array<char, 255> date;
```
375gnu, line 380: if you're in mood could you make this thread safe? I hope one day we can start using more then 1 core.

src/celengine/boundaries.cpp
```diff
@@ -74,8 +74,8 @@ ReadBoundaries(std::istream& in)
             currentChain.clear();
         }
 
-        currentChain.emplace_back(astro::equatorialToCelestialCart(ra, dec,
-                                                                   BoundariesDrawDistance));
+        currentChain.emplace_back(celestia::astro::equatorialToCelestialCart(ra, dec,
```
375gnu, line 77: better consistently use `namespace astro = celestia::astro;`, then one day when all code will be in `celestia` namespace no changes will be required.

src/celengine/astro.cpp
ajtribick: Yep, makes sense. Usages don't seem to be on any particularly critical paths, so might as well just return a `std::string`

src/celengine/astro.cpp
```diff
@@ -364,35 +419,35 @@ const char* astro::Date::toCStr(Format format) const
         break;
     }
 
-    strftime(date, sizeof(date), strftime_format, &cal_time);
+    auto length = std::strftime(result.data(), result.size(), strftime_format, &cal_time);
```
375gnu, line 422: could you check if we can use `fmt` here as well?

ajtribick: Have put something in but am currently on Windows, so need to do some testing with this code path active.

ajtribick:
I'm not prepared to mess with the internals of the JD/calendar conversions right now, hence sprinkling `//NOSONAR` all over it.

test/unit/kepler_test.cpp
```diff
@@ -9,6 +9,9 @@
 
 #include <doctest.h>
 
+namespace astro = celestia::astro;
+
+
```
375gnu, line 14: let's use only empty line as a separator

375gnu (changes requested):

375gnu (approved):

src/celengine/astro.cpp
```diff
@@ -321,15 +401,14 @@ astro::Date::Date(double jd)
...
 
+std::string
+Date::toString(Format format) const
+{
     if (format == ISO8601)
     {
-        snprintf(date, sizeof(date), "%04d-%02d-%02dT%02d:%02d:%08.5fZ",
-                 year, month, day, hour, minute, seconds);
-        return date;
+        return fmt::format("{:04}-{:02}-{:02}T{:02}:{:02}:{:02}:{:08.5f}Z"sv,
```
levinli303, line 410: @ajtribick we missed an argument here

PR #1952, opened 2023-10-26
New rendering for starfield

375gnu:


shaders/star_vert.glsl
```diff
@@ -1,11 +1,45 @@
...
+varying vec3 v_max_tetha_hk; // 24
+varying float pointSize; // 28
+
+uniform vec2 viewportSize;
+
+attribute vec4 in_Position;
+attribute vec3 in_Color;
+attribute float in_PointSize; // scaled brightness measured in Vegas
+
+float psf_max_theta(float br)
```
Askaniy, line 15: The function needs to be removed, it's outdated

shaders/star_vert.glsl
```diff
@@ -1,11 +1,45 @@
...
+{
+    return 0.012 + 1.488 / (sqrt(br_limit * 1000000.0 / (11.0 * br)) + 1.0);
+}
```
375gnu, line 18:
```suggestion
```

shaders/star_vert.glsl
```diff
@@ -1,11 +1,41 @@
...
+    float check = check_vec.x + check_vec.y + check_vec.z;
+    if (check == 0.0)
+    {
+        pointSize = 1.0;
+        v_tetha_k = vec3(-1.0);
+    }
+    else
+    {
+        float max_br = sqrt(max(color0.r, max(color0.g, color0.b)));
+        float max_theta = 0.33435822702992773 * sqrt(max_br); // glare radius
```
Askaniy, line 31: should be 0.2 * sqrt(max_br)

375gnu, line 31:
```suggestion
        float max_theta = 0.33435822702992773 * 0.2 * sqrt(max_br); // glare radius
```

375gnu: now stars are square

Askaniy: no, `float max_theta = 0.2 * sqrt(max_br); // glare radius`

shaders/star_vert.glsl
```diff
@@ -1,11 +1,41 @@
...
+    float check = check_vec.x + check_vec.y + check_vec.z;
+    if (check == 0.0)
+    {
+        pointSize = 1.0;
+        v_tetha_k = vec3(-1.0);
+    }
+    else
+    {
+        float max_br = sqrt(max(color0.r, max(color0.g, color0.b)));
+        float max_theta = 0.33435822702992773 * 0.2 * sqrt(max_br); // glare radius
```
Askaniy, line 31: `float max_theta = 0.2 * sqrt(max_br); // glare radius`

375gnu, line 31:
```suggestion
        float max_theta = 0.2 * sqrt(max_br); // glare radius
```

Askaniy:
Summary:
- Polishing the new star shader;
- Complete redesign of the light calculation methods;
- Automag was removed with the faintest magnitude system;
- Star styles were removed;
- [Classic/Enhanced/Legacy star colors were removed](https://github.com/CelestiaProject/Celestia/issues/2122);
- Sun white point is default;
- Astronomical constants updated to the latest standards;
- Merged [limb darkening](https://github.com/CelestiaProject/Celestia/pull/2047) branch.

src/celengine/render.cpp
```diff
@@ -3166,7 +3170,7 @@ void Renderer::buildRenderLists(const Vector3d& astrocentricObserverPos,
             // Compute the irradiance
             // Instead of summing the reflected light from all nearby stars,
             // we just consider the one with the highest irradiance.
-            float irradiance = 1.175494e-38f; // minimum positive float value
+            float irradiance = std::numeric_limits<float>::min();
```
Askaniy, line 3173: (I'm not sure this code would be uses after all, but) I mean here minimum *positive* value, now it would return 1.17549e-38

src/celastro/astro.h
```diff
@@ -25,8 +25,36 @@
...
+
+// IAU 2015 Resolution B3
+constexpr inline float SOLAR_IRRADIANCE   = 1361.0f; // W / m^2
+constexpr inline float SOLAR_POWER        = 3.828e26f; // W
+
+// Bessel (1979) for Johnson V filter
+constexpr inline float VEGAN_IRRADIANCE   = 3.640e-11f; // W / m^2
+
+// Auxiliary magnitude conversion factor
+constexpr inline float LN_MAG             = 5.0f / std::log(100.0f);
```
ajtribick, line 53: `std::log` is not constexpr until C++26. It works in GCC as an extension, but this is not supported on other compilers (Clang, MSVC, ...).

```suggestion
constexpr inline float LN_MAG             = 1.0857362f; // 5/ln(100)
```

ajtribick:
I can fix the conflicts with the master branch. Let me know if you prefer me to do this with a merge commit or rebasing this branch onto current master.

Askaniy:
I'd appreciate it. Maybe, rebasing this branch onto current master? But that's more up to you

Askaniy:
Let's count current bugs I created and left 🥲
(For rendering box testing, I turned on the red filler at line 47 in `star_frag.glsl`)

There are two stages of rendering a star, depending on its individual visible brightness: "9 px" mode (3x3 box size) and "glow" mode (variable box size). For simplicity, the separating value is chosen to be 1 Vega brightness at exposure 1. The separation is in `star_vert.glsl` at line 46.

You can notice that in the glow mode, the glow doesn't completely fill the box. Actually fills, but only in internal Float values. sRGB gamma correction is assumed in the glow design, which will “pull” it to the box edges.

## Bug 1
Star field have never been seen using both modes at once, depending on the brightness of a particular star. If you don't change the code (`if (all(lessThan(scaled_color, vec3(1.0))))`), the glow mode is used. If you change to `if (true)`, the 9 px mode is used.
изображение
That's how 4 px mode looked like. Stars feel square, that's why I moved it to 9 px instead.
изображение

## Bug 2
The 9 px mode was broken since my last testing in the summer 2024. When rotating the camera, the 1 pixel stars rarely appear and then disappear again:
изображение

## Bug 3
The wrong brightness values are obviously being supplied to the input. The glow mode worked correctly before the exposure transition was made:
Screenshot_20240701_114940

## Bug 4
After ruthless code reduction, barycenters began to emit. I'm guessing it was an old optimization whose limits were accidentally removed. I think it's best to find and remove the barycenter glow all the way through.
изображение

## Bug 5
Limb darkening was added, but the star surfaces have stopped rendering. It could be a typo or caused by deleting something extra.

## Bug 6
There is a small “bug” with exposure setting in the interface: I failed to make integer output when updating a variable, i.e. now `Exposure: 16.00` is printed instead of `Exposure: 16`. I rolled back my experiments

## Notes

Based on the Discord discussions, it would be both more realistic and easier to program the overall process of rendering both stars, planets, and spacecraft if they are in point mode. Without special brightness enhancement of planets, for example. For the sake of simplifying visual search, you can set different thresholds for displaying names. Now there are no star styles, which may allow simplifying the logic globally. I also removed the glow and atmosphere of the stars, while adding changes from the limb-darkening branch to optimize the logic of everything planned at once. To keep the Sun looking bright, I suggest increasing the maximum size of the glow to the size of the screen. At typical exposures, only the Sun will be bright, and it will switch to a starry background with no jumps when removed.

Also, maximum glow box size is discussed to be `min(min(screen_hight, screen_width), max_point_size)`, depending on the max point size that openGL supports.

Askaniy:
Chara updated astronomical constants (#2371) and made Sun white point the default one (#2372), to delineate the branch

AstroChara:
Since it is somehow attached to this PR, I would like to propose that we cancel the complete deletion of the "Classic" star rendering mode; instead, we recycle its code infrastructure to create a "true colour" mode, where each star type's colour is based on colour from TrueColorTools — so light yellow late M-dwarfs and pink/magenta T-dwarfs, for example.

AstroChara:
On a more general note, we have seen how combining multiple completely different tasks into one creates this mess which is very difficult to understand and by now taking over three years to complete.

I think we should look into taking apart this PR, putting each indivisible component of this PR into their own PR, where they can be debugged and reviewed and incorporated independently. Given the scale of this PR, it would be a lot of effort, but it should make things far easier and far more mobile in the long run.

We can also remove elements already in the program like updating constants or setting default star colour mode.

375gnu:
@Askaniy do you still have your original python files? it lost them, i'd like to play with them with different exposures.

Askaniy:
@375gnu do you mean https://github.com/Askaniy/CelestiaStarRenderer or some older version of it?

375gnu:
Yes, this one. Thanks!

ObmijSelaw:
greetings, I was recommended on Discord to come here to let you all know about my efforts in getting sRGB gamma correction working in tandem with this new star rendering method

I have been working on it as a "fork", but I am keeping it as closely updated to master upstream so it can be (hopefully) easily integrated when all bugs are squashed and the moment is right

[link to my "fork" here on github](
https://github.com/ObmijSelaw/Celestia/tree/srgb+psf-mk1)

thus far, I have integrated code up to commit 2ca7268 from askaniys-stars branch

Now, to recap, @Askaniy said on Discord that this star rendering system should be able to "discriminate" between the 9px rendering mode (i understand this is for low-brightness stars) and glow mode (for brighter stars), however, this is not currently working. 

i confirm in my builds that this is indeed broken and seemingly only glow mode works, if the stars get dimmed below a certain brightness threshold, they disappear altogether instead of becoming dim. Only very bright stars are rendered

so far i have not touched the code beyond said commit, so i have not integrated the commits replacing faintest-magnitude with exposure, nor deleting legacy star colors/styles, etc.

I am thinking the priority should be figuring out what's preventing this "dual mode" code from working, what do you say?
Screenshot_20260205_031826
Screenshot_20260205_031807

Askaniy:
[I just finished complete redesign of the star rendering method.](https://github.com/Askaniy/CelestiaStarRenderer) It's both easier to implement and closer to the original PSF. And has been tested a lot.

Last time, the method required separating the modes within the shader. And that didn't work out. Now there's no need to deal with that: the new method could be split into two separate shaders, one for the point and one for eye PSF. See the general logic in [`RenderEngine.draw_source()`](https://github.com/Askaniy/CelestiaStarRenderer/blob/master/CSR.py#L53)

AstroChara:
> [I just finished complete redesign of the star rendering method.](https://github.com/Askaniy/CelestiaStarRenderer) It's both easier to implement and closer to the original PSF. And has been tested a lot.
> 
> Last time, the method required separating the modes within the shader. And that didn't work out. Now there's no need to deal with that: the new method could be split into two separate shaders, one for the point and one for eye PSF. See the general logic in [`RenderEngine.draw_source()`](https://github.com/Askaniy/CelestiaStarRenderer/blob/master/CSR.py#L53)

I wonder if we should just make a new PR, and we can simply reuse the important bits in this PR (like exposure implementation) and discard the outdated stuff like limb darkening or removing legacy colours (whose infrastructure can instead be reused).

Askaniy:
Making a new PR is what I think could work.

> we can simply reuse the important bits in this PR (like exposure implementation)

But exposure I implemented is exactly the thing that is broken in this PR.

I propose adding the stars first, and exposure system after. We would need to convert `FaintestMag` to linear irradiance (maybe it's better to use SI units), and then convert the irradiance to the radiance scale that the shader uses (where radiance=1 is the scale of transition between point and eye PSF modes).

AstroChara:
> I propose adding the stars first, and exposure system after. We would need to convert `FaintestMag` to linear irradiance (maybe it's better to use SI units), and then convert the irradiance to the radiance scale that the shader uses (where radiance=1 is the scale of transition between point and eye PSF modes).

I believe that works. If we add just the PSFs, we will need the star rendering itself, the removal of the current star styles (can we just comment out the infrastructure to be reused later?), and add the conversion of magnitudes to irradiance, as you said.

PR #1953, opened 2023-10-26
Refactor sampled orbit and orientation classes

ajtribick:
- Store sample times and rotations in separate vectors to reduce padding
- Do not permit out-of-order timestamps in sample files
- Do not permit empty sampled trajectories or rotations
- Pre-rotate the sampled trajectory positions and velocities


375gnu:
> * Pre-rotate the sampled trajectory positions and velocities

I'm barely familiar with orbits code, so how this affect other codebase?

ajtribick:
Essentially the sampled orbits were doing the (x,y,z) -> (x, z, -y) operation every time the orbits were sampled. The change just does this when the orbits are constructed. This matches what the sampled orientations were doing and simplifies the evaluation functions a little (in particular, it makes it easier to do early returns for various edge cases, since there's no need to manipulate the coordinates. I guess it leads to slightly less computation but the effect should be negligible.

375gnu (approved):

PR #1955, opened 2023-10-27
[cmake] Allow enabling LTO for Qt frontend and libcelestia.so

375gnu:
Main code and Qt5 frontend benefit from LTO. Gtk+ and SDL frontends increase their sizes, i think it's due to statically linked libfmt.

| Component | Without LTO | With LTO |
|-----------------|------------------|---------------|
| libcelestia | 4367840 | 3777176 |
| celestia-qt5 | 1207152 | 1112024 |
| celestia-gtk | 656136 | 772136 |
| celestia-sdl | 439704 | 588576 |

g++ is 7.5.0. Other compilers may give other numbers.

With g++12.2 and 1.10.1:
| Component | Without LTO | With LTO |
|-----------------|------------------|---------------|
| libcelestia | 4253848 | 4547936 |
| celestia-qt6 | 875240 | 825368 |


375gnu:
MSVC doesn't like something.

ajtribick:
Tried this locally, it works with the Ninja generator but not the Visual Studio one. Need to figure out why but supplying `-G "Ninja"` in the cmake parameters in ci.yml might fix it.

ajtribick:
Ok further investigation - Ninja doesn't work either, but it was passing a build configuration of RelWithDebInfo while MSVC was translating this to Release, so Ninja didn't use LTO.

The reason appears to be that WINDOWS_EXPORT_ALL_SYMBOLS is not compatible with LTO for DLLs - we'd probably have to mark the relevant functions with the appropriate __dllspec attributes, or compile the celestia library as static rather than shared.

ajtribick (approved):

PR #1958, opened 2023-10-29
Split (some parts of) celestiacore

ajtribick:
The original idea I had here was to update formatnum.cpp to use ICU. However this would involve storing ICU format objects somewhere and I didn't want to add yet more stuff to the mess that is celestiacore.cpp.

So here goes with moving some of the stuff in celestiacore.cpp out into their own classes: here I focus on the overlay, which also pulls with it the text input and the handling of views.

This is a first step towards reducing the amount of stuff in celestiacore, in the future it should be possible to remove some of the proxy methods and just expose the various component classes directly.

Regarding the Sonar complaints, this is already large enough without trying to fix all the old code that's accumulated here.

Quick tests of Qt5 and Windows front-ends indicates that this seems to work.

Will squash the commits before merging but for now leaving as-is in case of the need to debug.

### DIFF
--- a/src/celestia/hud.cpp
+++ b/src/celestia/hud.cpp
@@ -35,6 +35,7 @@
 #include <celengine/universe.h>
 #include <celmath/geomutil.h>
 #include <celmath/mathlib.h>
+#include <celutil/flag.h>
 #include <celutil/formatnum.h>
 #include <celutil/gettext.h>
 #include <celutil/utf8.h>
@@ -81,30 +82,21 @@ SigDigitNum(double v, int digits)
 std::string
 KelvinToStr(float value, int digits, TemperatureScale temperatureScale)
 {
-    const char* unitTemplate = "";
-
     switch (temperatureScale)
     {
         case TemperatureScale::Celsius:
-            value = KelvinToCelsius(value);
-            unitTemplate = "{} °C";
-            break;
+            return fmt::format("{} °C", SigDigitNum(KelvinToCelsius(value), digits));
         case TemperatureScale::Fahrenheit:
-            value = KelvinToFahrenheit(value);
-            unitTemplate = "{} °F";
-            break;
+            return fmt::format("{} °F", SigDigitNum(KelvinToFahrenheit(value), digits));
         default: // TemperatureScale::Kelvin
-            unitTemplate = "{} K";
-            break;
+            return fmt::format("{} K", SigDigitNum(value, digits));
     }
-    return fmt::format(unitTemplate, SigDigitNum(value, digits));
 }
 
 std::string
 DistanceLyToStr(double distance, int digits, MeasurementSystem measurement)
 {
-    const char* units = "";
-
+    const char* units;
     if (std::abs(distance) >= astro::parsecsToLightYears(1e+6))
     {
         units = _("Mpc");
@@ -126,7 +118,7 @@ DistanceLyToStr(double distance, int digits, MeasurementSystem measurement)
     }
     else if (measurement == MeasurementSystem::Imperial)
     {
-        if (abs(distance) > astro::kilometersToLightYears(OneMiInKm))
+        if (std::abs(distance) > astro::kilometersToLightYears(OneMiInKm))
         {
             units = _("mi");
             distance = astro::lightYearsToKilometers(distance) / OneMiInKm;
@@ -137,18 +129,15 @@ DistanceLyToStr(double distance, int digits, MeasurementSystem measurement)
             distance = astro::lightYearsToKilometers(distance) / OneFtInKm;
         }
     }
+    else if (std::abs(distance) > astro::kilometersToLightYears(1.0f))
+    {
+        units = _("km");
+        distance = astro::lightYearsToKilometers(distance);
+    }
     else
     {
-        if (std::abs(distance) > astro::kilometersToLightYears(1.0f))
-        {
-            units = _("km");
-            distance = astro::lightYearsToKilometers(distance);
-        }
-        else
-        {
-            units = _("m");
-            distance = astro::lightYearsToKilometers(distance) * 1000.0f;
-        }
+        units = _("m");
+        distance = astro::lightYearsToKilometers(distance) * 1000.0f;
     }
 
     return fmt::format("{} {}", SigDigitNum(distance, digits), units);
@@ -262,7 +251,8 @@ displaySpeed(Overlay& overlay, float speed, MeasurementSystem measurement)
 std::string
 angleToStr(double angle)
 {
-    int degrees, minutes;
+    int degrees;
+    int minutes;
     double seconds;
     astro::decimalToDegMinSec(angle, degrees, minutes, seconds);
 
@@ -297,7 +287,8 @@ displayApparentDiameter(Overlay& overlay, double radius, double distance)
 void
 displayDeclination(Overlay& overlay, double angle)
 {
-    int degrees, minutes;
+    int degrees;
+    int minutes;
     double seconds;
     astro::decimalToDegMinSec(angle, degrees, minutes, seconds);
 
@@ -309,7 +300,8 @@ displayDeclination(Overlay& overlay, double angle)
 void
 displayRightAscension(Overlay& overlay, double angle)
 {
-    int hours, minutes;
+    int hours;
+    int minutes;
     double seconds;
     astro::decimalToHourMinSec(angle, hours, minutes, seconds);
 
@@ -391,9 +383,9 @@ displayPlanetocentricCoords(Overlay& overlay,
         Eigen::Quaterniond q = body.getEclipticToEquatorial(astro::J2000);
         bool retrograde = (q * Eigen::Vector3d::UnitY()).y() < 0.0;
 
-        if ((latitude < 0.0) ^ retrograde)
+        if ((latitude < 0.0) != retrograde)
             nsHemi = 'S';
-        else if ((latitude > 0.0) ^ retrograde)
+        else if ((latitude > 0.0) != retrograde)
             nsHemi = 'N';
 
         if (retrograde)
@@ -416,13 +408,13 @@ displayPlanetocentricCoords(Overlay& overlay,
 void
 displayStarInfo(Overlay& overlay,
                 int detail,
-                Star& star,
+                const Star& star,
                 const Universe& universe,
                 double distance,
-                MeasurementSystem measurement,
-                TemperatureScale temperatureScale)
+                const HudSettings& hudSettings)
 {
-    overlay.printf(_("Distance: %s\n"), DistanceLyToStr(distance, 5, measurement));
+    overlay.printf(_("Distance: %s\n"),
+                   DistanceLyToStr(distance, 5, hudSettings.measurementSystem));
 
     if (!star.getVisibility())
     {
@@ -448,7 +440,7 @@ displayStarInfo(Overlay& overlay,
             break;
         default:
             star_class = star.getSpectralType();
-        };
+        }
         overlay.printf(_("Class: %s\n"), star_class);
 
         displayApparentDiameter(overlay, star.getRadius(),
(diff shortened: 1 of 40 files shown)


ajtribick:
I think we still need a number formatter: the locale-specific formats provided by fmt don't give the significant figure control that we are currently using (e.g. formatting 12345 as 12000 to 3 decimal places), and output current codepage rather than UTF-8 on Windows.

src/celestia/viewmanager.cpp
```diff
@@ -0,0 +1,342 @@
...
+ViewManager::deleteView(Simulation* sim, View* v)
+{
+    if (v == nullptr)
+        v = *m_activeView;
+
+    if (v->isRootView())
+        return false;
+
+    //Erase view and parent view from views
+    for (auto i = m_views.begin(); i != m_views.end(); )
```
375gnu, line 321: can we use std::remove_if here?
```c++
m_views.erase(std::remove_if(m_views.begin(), m_views.end(), [v](auto i) { return i == v || i == v.parent; })
```

no tested :)

src/celestia/viewmanager.cpp
```diff
@@ -0,0 +1,342 @@
...
+    if (av == nullptr)
+        av = *m_activeView;
+
+    if (!av->isSplittable(type))
+        return ViewSplitResult::NotSplittable;
+
+    Observer* o = sim->addObserver();
+
+    // Make the new observer a copy of the old one
+    // TODO: This works, but an assignment operator for Observer
```
375gnu, line 254: could you address this as well?

src/celestia/viewmanager.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+namespace
+{
+
+bool
+pointOutsideView(const View* view,
+                 const WindowMetrics& metrics,
+                 float x, float y)
+{
+    return x+2 < view->x * metrics.width ||
```
375gnu, line 34: i believe we should use a constant instead of 2

src/celestia/timeinfo.h
```diff
@@ -0,0 +1,25 @@
...
+
+namespace celestia
+{
+
+struct TimeInfo
+{
+    static constexpr double MinimumTimeRate = 1.0e-15;
+
+    double currentTime{ 0.0 };
+    double fps{ 0.0 };
```
375gnu, line 20: do we really double precision for fps?

src/celestia/timeinfo.h
```diff
@@ -0,0 +1,25 @@
...
+    int timeZoneBias{ 0 }; // Diff in secs between local time and GMT
+    bool lightTravelFlag { false };
```
375gnu, line 22:
```suggestion
    bool lightTravelFlag{ false };
```

src/celestia/textinput.cpp
```diff
@@ -0,0 +1,169 @@
...
+    else if (c == '\177') // BackTab
+    {
+        if (m_completionIdx > 0)
+            m_completionIdx--;
+        else if (m_completionIdx == 0)
+            m_completionIdx = m_completion.size() - 1;
+        else if (m_completion.size() > 0)
+            m_completionIdx = m_completion.size() - 1;
+        if (m_completionIdx >= 0) {
+            std::string::size_type pos = m_text.rfind('/', m_text.length());
```
375gnu, line 111:
```suggestion
        if (m_completionIdx >= 0)
        {
            auto pos = m_text.rfind('/');
```

src/celestia/textinput.cpp
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+    else if (c == '\011') // TAB
+    {
+        if (m_completionIdx + 1 < (int) m_completion.size())
+            m_completionIdx++;
+        else if ((int) m_completion.size() > 0 && m_completionIdx + 1 == (int) m_completion.size())
+            m_completionIdx = 0;
+        if (m_completionIdx >= 0) {
+            std::string::size_type pos = m_text.rfind('/', m_text.length());
```
375gnu, line 90:
```suggestion
        if (m_completionIdx >= 0)
        {
            auto pos = m_text.rfind('/');
```

src/celestia/textinput.h
```diff
@@ -0,0 +1,51 @@
...
+
+class TextInput
+{
+public:
+    std::string_view getTypedText() const;
+    util::array_view<std::string> getCompletion() const;
+    int getCompletionIndex() const;
+
+    CharEnteredResult charEntered(const Simulation*, std::string_view, bool withLocations);
+    void appendText(const Simulation*, std::string_view, bool withLocations);
```
375gnu, line 42: this name is confusing

src/celestia/textinput.cpp
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+    else if (c == '\177') // BackTab
+    {
+        if (m_completionIdx > 0)
+            m_completionIdx--;
+        else if (m_completionIdx == 0)
+            m_completionIdx = m_completion.size() - 1;
+        else if (m_completion.size() > 0)
+            m_completionIdx = m_completion.size() - 1;
```
375gnu, line 109:
```suggestion
        else if (m_completionIdx == 0 || m_completion.size() > 0)
            m_completionIdx = m_completion.size() - 1;
```

src/celestia/textinput.cpp
```diff
@@ -0,0 +1,169 @@
...
+            std::string::size_type pos = m_text.rfind('/', m_text.length());
+            if (pos != std::string::npos)
+            {
+                m_text.resize(pos + 1);
+                m_text.append(m_completion[m_completionIdx]);
+            }
+            else
+            {
+                m_text = m_completion[m_completionIdx];
+            }
```
375gnu, line 120:
```suggestion
            if (pos != std::string::npos)
                m_text.resize(pos + 1);
            m_text = m_completion[m_completionIdx];
```

src/celestia/textinput.cpp
```diff
@@ -0,0 +1,169 @@
...
+        std::string::size_type pos = m_text.rfind('/', m_text.length());
+        if (pos != string::npos)
+        {
+            m_text.resize(pos + 1);
+            m_text.append(m_completion[0]);
+        }
+        else
+        {
+            m_text = m_completion[0];
+        }
```
375gnu, line 155:
```suggestion
        if (pos != string::npos)
            m_text.resize(pos + 1);
        m_text = m_completion[0];
```

src/celestia/textinput.cpp
```diff
@@ -0,0 +1,169 @@
...
+            std::string::size_type pos = m_text.rfind('/', m_text.length());
+            if (pos != std::string::npos)
+            {
+                m_text.resize(pos + 1);
+                m_text.append(m_completion[m_completionIdx]);
+            }
+            else
+            {
+                m_text = m_completion[m_completionIdx];
+            }
```
375gnu, line 99:
```suggestion
            if (pos != std::string::npos)
                m_text.resize(pos + 1);
            m_text = m_completion[m_completionIdx];
```

src/celestia/textinput.cpp
ajtribick: That doesn't do the same thing... think it would be:

```cpp
m_text.resize(pos == std::string::npos ? 0 : pos + 1);
m_text.append(m_completion[m_completionIdx]);
```

src/celestia/textinput.h
ajtribick: It appends text to the typed text.

src/celestia/viewmanager.cpp
ajtribick: We do have copy constructor/assignment operators, so probably this comment is just outdated? Will remove it.

src/celestia/viewmanager.cpp
```diff
@@ -0,0 +1,342 @@
...
+    if (it == m_views.end())
+        return;
+
+    m_activeView = it;
+    sim->setActiveObserver((*m_activeView)->observer);
+}
+
+
+bool
+ViewManager::deleteView(Simulation* sim, View* v)
```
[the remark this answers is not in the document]
ajtribick: Not possible due to the call to `View::remove()`, which does a whole bunch of tree updating and requires a non-const `View*`. The view classes are generally a mess but I'd rather sort that out in a separate PR.

src/celestia/timeinfo.h
ajtribick: Yep. https://github.com/CelestiaProject/Celestia/blob/c74cd97422538479a508aae9a580357e3c73b277/src/celestia/celestiacore.h#L531

src/celestia/viewmanager.cpp
ajtribick: made the method duplicate the active observer, since nothing else was calling it.

ajtribick:
Leaving the update of the view system to use smart pointers for a future pull request - would likely result in a fairly large set of changes to simulation.cpp/.h as a result.

src/celestia/textprintposition.h
```diff
@@ -1,42 +1,45 @@
...
+        Relative           = 0x00,
+        Absolute           = 0x01,
+        PositionType       = 0x01,
+        HorizontalStart    = 0x00,
+        HorizontalMiddle   = 0x02,
+        HorizontalEnd      = 0x04,
+        HorizontalOffset   = 0x06,
+        VerticalBottom     = 0x00,
+        VerticalMiddle     = 0x08,
+        VerticalTop        = 0x0f,
```
375gnu, line 31:
```suggestion
        VerticalTop        = 0x10,
```

375gnu: could you leave a comment on this enum usage?

src/celestia/textinput.h
375gnu: It looks like it adds object names, so maybe rename it into `appendName`?

src/celestia/hud.cpp
```diff
@@ -0,0 +1,1226 @@
...
+displayMass(Overlay& overlay, float mass, MeasurementSystem measurement)
+{
+    if (mass < 0.001f)
+    {
+        if (measurement == MeasurementSystem::Imperial)
+            overlay.printf(_("Mass: %.6g lb\n"), mass * astro::EarthMass / (float) OneLbInKg);
+        else
+            overlay.printf(_("Mass: %.6g kg\n"), mass * astro::EarthMass);
+    }
+    else if (mass > 50)
```
375gnu, line 192:
```suggestion
    else if (mass > 50.0f)
```

src/celestia/hud.cpp
```diff
@@ -0,0 +1,1226 @@
...
+    {
+        overlay.printf(_("Absolute magnitude: %.1f\n"), absMag);
+    }
+}
+
+void
+displayRADec(Overlay& overlay, const Eigen::Vector3d& v)
+{
+    double phi = std::atan2(v.x(), v.z()) - celestia::numbers::pi / 2;
+    if (phi < 0)
```
375gnu, line 332:
```suggestion
    if (phi < 0.0)
```

src/celestia/hud.cpp
```diff
@@ -0,0 +1,1226 @@
...
+        phi = phi + 2 * celestia::numbers::pi;
```
375gnu, line 333:
```suggestion
        phi = phi + 2.0 * celestia::numbers::pi;
```

src/celestia/hud.cpp
```diff
@@ -0,0 +1,1226 @@
...
+
+    double theta = std::atan2(std::hypot(v.x(), v.z()), v.y());
+    if (theta > 0)
+        theta = celestia::numbers::pi / 2 - theta;
```
375gnu, line 337:
```suggestion
        theta = celestia::numbers::pi * 0.5 - theta;
```

src/celestia/hud.cpp
```diff
@@ -0,0 +1,1226 @@
...
+    else
+        theta = -celestia::numbers::pi / 2 - theta;
```
375gnu, line 339:
```suggestion
        theta = -celestia::numbers::pi * 0.5 - theta;
```

src/celestia/hud.cpp
```diff
@@ -0,0 +1,1226 @@
...
+
+void displayDSOinfo(Overlay& overlay,
+                    const DeepSkyObject& dso,
+                    double distance,
+                    MeasurementSystem measurement)
+{
+    overlay.print(dso.getDescription());
+    overlay.print("\n");
+
+    if (distance >= 0)
```
375gnu, line 490:
```suggestion
    if (distance >= 0.0)
```

src/celestia/hud.cpp
```diff
@@ -0,0 +1,1226 @@
...
+    if (body == nullptr)
+        return;
+
+    Eigen::Vector3f locPos = location.getPosition();
+    Eigen::Vector3d lonLatAlt = body->cartesianToPlanetocentric(locPos.cast<double>());
+    displayPlanetocentricCoords(overlay, *body,
+                                lonLatAlt.x(), lonLatAlt.y(), lonLatAlt.z(), false, measurement);
+}
+
+std::string
```
375gnu, line 613:
```suggestion
std::string /* TODO: replace with std::string_view */
```

375gnu (changes requested):

src/celestia/hud.h
```diff
@@ -0,0 +1,193 @@
...
+    void textEnterMode(TextEnterMode mode);
+
+    void setOverlay(std::unique_ptr<Overlay>&&);
+    void setWindowSize(int, int);
+    void setTextAlignment(LayoutDirection);
+
+    int getTextWidth(std::string_view) const;
+    const std::shared_ptr<TextureFont>& font() const;
+    const std::shared_ptr<TextureFont>& titleFont() const;
+    bool trySetFont(const std::shared_ptr<TextureFont>&);
```
375gnu, line 141: why the hud should care about value of `f`? i think it's a job of the caller, so i propose jusr `setFont` & `setTitleFont`

375gnu (changes requested):

src/celestia/textinput.h
ajtribick: It's also what's used for pasting text in the Qt frontend.

src/celestia/hud.cpp
ajtribick: Simple `string_view` doesn't work here because the star branch can generate a new string if the star has only catalog numbers. Would probably have to be something along the lines of a `std::variant<std::string, std::string_view>`

375gnu (approved):

PR #1962, opened 2023-11-03
Add Georgian translation

375gnu:


ajtribick (approved):
Can't really review the content of the translations, but seems ok from a file infrastructure perspective.

PR #1965, opened 2023-11-04
[win] Do not process ka.po for Windows front-end

ajtribick:
Currently we can't support Georgian in the Windows front-end since there is no 8-bit code page to support it. Maybe if we change the Windows front-end to use the Unicode wide character APIs instead we can re-enable this.

Qt should still work here.

For now this should unbreak the build, created issue #1966 for the Windows front-end rewrite.


375gnu (approved):
oh, i forgot about this  file completelly

PR #1972, opened 2023-11-06
Disable geometry shaders on Navi with radeonsi Mesa driver

375gnu:
There is a bug in the driver, so only top-left triangle is rendered.

Also disable location labels shown by default cause it's annoying (restore 1.6 behavior).

### DIFF
--- a/src/celengine/glsupport.cpp
+++ b/src/celengine/glsupport.cpp
@@ -1,5 +1,6 @@
 #include "glsupport.h"
 #include <algorithm>
+#include <cstring>
 
 namespace celestia::gl
 {
@@ -24,6 +25,8 @@ CELAPI GLint maxTextureAnisotropy          = 0;
 namespace
 {
 
+bool EnableGeomShaders = true;
+
 inline bool has_extension(const char *name) noexcept
 {
     return epoxy_has_gl_extension(name);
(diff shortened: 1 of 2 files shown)


src/celengine/glsupport.cpp
```diff
@@ -35,7 +38,38 @@ bool check_extension(util::array_view<std::string> list, const char *name) noexc
            && has_extension(name);
 }
 
-bool EnableGeomShaders = true;
+void enable_workarouds()
```
ajtribick, line 41: typo: workarounds

ajtribick (changes requested):

src/celengine/glsupport.cpp
375gnu, line 41:
```suggestion
void enable_workarounds()
```

src/celengine/glsupport.cpp
```diff
@@ -71,6 +105,8 @@ bool init(util::array_view<std::string> ignore) noexcept
     if (gl::EXT_texture_filter_anisotropic)
         glGetIntegerv(GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT, &maxTextureAnisotropy);
 
+    enable_workarouds();
```
375gnu, line 108:
```suggestion
    enable_workarounds();
```

ajtribick (approved):

Issue #1975, opened 2023-11-07
Gaps in ring


levinli303:
**Describe the bug**
See screenshot

**To Reproduce**
1. Download this [add-on](https://celestia.mobi/resources/item?item=C276F25A-DFE6-7F33-10D1-135C7DA4AB72)
2. Go to URL: cel://Follow/Sol:Pioneer%2011/1979-09-01T16:37:11.00780Z?x=uRn60gM&y=bjF0tQ&z=GAMGhgI&ow=0.8676776&ox=0.16016424&oy=-0.4495414&oz=0.13926755&select=Sol:Saturn&fov=28.218485&ts=100&ltd=0&p=1&rf=70571923&nrf=191&lm=2114&tsrc=0&ver=3
3. See error

**Expected behavior**
Rings have no gap

**Screenshots**
<img width="1136" alt="Screenshot 2023-11-07 at 09 32 14" src="https://github.com/CelestiaProject/Celestia/assets/17924287/e1b3141b-b6e7-4176-9c63-ff8ffd90b749">

375gnu:
When spacecraft display is off there are no issues.

375gnu:
When spacecraft ON near plain distance is 0.00448116, when they are OFF it's 70.1225.

PR #1976, opened 2023-11-09
Add rings radius to renderable body radius

375gnu:
When we have a satellite close to the planet then its rings and the planet
itself is not visible (it has centerZ > 0) but the rings are visible
Celestia incorrectly sorts render list for solar system object because
it uses objects' farZ which is computed as centerZ - radius. So when
the satellite is too close and the planet is "behind" such computation
gives a result that the satellite is further then the planet (centerZ for
the satellite < 0, but for the planet > 0) As a result depth partitions
are incorrectly computed and we have a close object (the satellite) and
a distant object (the planet) in the same depth partition. And this
partition has huge farZ/nearZ (> 1e6). Such huge difference between
the near and the far plans gives rendering artifacts.

To fix this we should take into account rings radius when calculating
render item radius.

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -3245,6 +3245,8 @@ void Renderer::addRenderListEntries(RenderListEntry& rle,
             rle.isOpaque = true;
         }
         rle.radius = body.getRadius();
+        if (body.getRings() != nullptr)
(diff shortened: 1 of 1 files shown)


375gnu:
Closes: #1975

levinli303 (approved):

PR #1979, opened 2023-11-14
Allow font in hud to be null

levinli303:


375gnu (changes requested):
this should be handled by the caller

375gnu (approved):

PR #1981, opened 2023-11-14
Use fmt to generate skygrid labels

ajtribick:


src/celengine/skygrid.cpp
```diff
@@ -260,20 +263,20 @@ SkyGrid::longitudeLabel(int longitude, int longitudeStep) const
...
-    }
+    std::string result = fmt::format("{}{}", longitude / baseUnit, baseUnitSymbol);
+    if (longitudeStep % baseUnit == 0)
+        return result;
+
+    auto it = fmt::format_to(std::back_inserter(result), " {:02}{}", (longitude / MIN) % 60, minuteSymbol);
+    if (longitudeStep % MIN == 0)
+        return result;
+
+    it = fmt::format_to(it, " {:02}", (longitude / SEC) % 60);
```
375gnu, line 274: maybe better use `std::back_inserter(result)` as well? so the code is easier to understand and less dependent on the previous lines

ajtribick: Ok, that's how I did it originally...

375gnu (approved):

PR #1986, opened 2023-11-18
Disallow zero SemiAxes and Oblateness values

375gnu:


src/celengine/solarsys.cpp
```diff
@@ -847,7 +861,16 @@ Body* CreateBody(const std::string& name,
     {
         auto oblateness = planetData->getNumber<float>("Oblateness");
         if (oblateness.has_value())
-            body->setSemiAxes(body->getRadius() * Eigen::Vector3f(1.0f, 1.0f - *oblateness, 1.0f));
+        {
+            if (*oblateness > 0.0f && *oblateness < 1.0f)
```
ajtribick, line 865: Zero oblateness is a valid value corresponding to a sphere. If we disallow it, we prevent modifying an object from oblate to spherical.

ajtribick (changes requested):

JiliTheSpaceboy:
Since I asked about this on Discord: I was hoping that the check would be placed just on density calculation part so we can still make flat objects and not break the program. That said I am not sure where else in the program such a zero volume object would cause problems...

src/celengine/solarsys.cpp
375gnu: makes sense

375gnu, line 865:
```suggestion
            if (*oblateness >= 0.0f && *oblateness < 1.0f)
```

375gnu:
> so we can still make flat objects and not break the program.

What object do you want to represent with such flat pseudo-spheres?

JiliTheSpaceboy:
> > so we can still make flat objects and not break the program.
> 
> What objects do you want to represent with such flat pseudo-spheres?

You can probably create a simple sunshade out of it perhaps. I figured that if it doesn't break anything else, it would be better to allow it.

ajtribick (approved):

PR #1987, opened 2023-11-18
Move celimage files into celestia::engine and rename PixelFormat formats

375gnu:
+ unrelated bug fix for HUD

it seems to me that sonar is false-positive here.

### DIFF
--- a/src/celscript/lua/celx_internal.h
+++ b/src/celscript/lua/celx_internal.h
@@ -45,6 +45,11 @@ enum
     Celx_Category = 13
 };
 
+template<typename T> int celxClassId(T)
+{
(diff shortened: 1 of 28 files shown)


src/celscript/lua/celx_internal.h
```diff
@@ -45,6 +45,10 @@ enum
     Celx_Category = 13
 };
 
+template<typename T> int celxClassId(T)
```
375gnu, line 48: i have no idea how it worked, but with Image moved to celestia::engine g++ is no more able find a proper definition.

ajtribick (approved):

PR #1994, opened 2023-11-23
Fix crash loading DDS

levinli303:


src/celimage/dds.cpp
```diff
@@ -256,12 +256,12 @@ Image* LoadDDSImage(const fs::path& filename)
...
                 if (tmp != nullptr)
                 {
-                    pixels = std::make_unique<std::uint32_t>(ddsd.width * ddsd.height);
+                    pixels = std::make_unique<std::uint32_t[]>(ddsd.width * ddsd.height);
                     // crop
                     for (std::uint32_t y = 0; y < ddsd.height; y++)
                     {
-                        std::memcpy(reinterpret_cast<char*>(pixels.get() + y * ddsd.width * 4),
-                                    reinterpret_cast<char*>(tmp.get() + y * nw * 4),
+                        std::memcpy(reinterpret_cast<char*>(pixels.get() + y * ddsd.width),
```
levinli303, line 263: The old code was

```memcpy((char *)pixels + y * ddsd.width * 4, (char *)tmp + y * nw * 4, ddsd.width * 4);```

the pointer cast was performed earlier so we don't need to * 4 in the new code.

ajtribick, line 263: I think we can remove the `reinterpret_cast` here (and on the line below) as the conversion to `void*` is implicit.

levinli303: removed `reinterpret_cast`

ajtribick (approved):

PR #1995, opened 2023-11-29
Include size and screen dpi in cache key for fonts

levinli303:
screen dpi can change in runtime and it influence font sizes so we need to cache it too.


src/celttf/truetypefont.cpp
```diff
@@ -693,7 +693,28 @@ ParseFontName(const fs::path &filename, int &index, int &size)
...
 
-using FontCache = std::map<std::pair<fs::path, int>, std::weak_ptr<TextureFont>>;
+struct FontCacheKey
+{
+    fs::path filename;
+    int index;
+    int size;
+    int screenDpi;
+
+    bool operator ==(const FontCacheKey &other) const
```
375gnu, line 703:
```suggestion
    bool operator==(const FontCacheKey &other) const
```

375gnu (changes requested):
please fix whitespace

src/celttf/truetypefont.cpp
levinli303: fixed

375gnu (approved):

PR #2001, opened 2023-12-04
Simplify winutil.h API

ajtribick:
Remove some unused methods, and move the remaining ones to the celestia::util namespace.

### DIFF
--- a/src/celutil/winutil.cpp
+++ b/src/celutil/winutil.cpp
@@ -10,71 +10,74 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-#include <windows.h>
 #include "winutil.h"
 
-using namespace std;
+#include <Windows.h>
 
-const char* CurrentCP()
+namespace celestia::util
 {
-    static bool set = false;
-    static char cp[20] = "CP";
-    if (!set)
-    {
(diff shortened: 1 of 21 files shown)


src/celutil/winutil.cpp
```diff
@@ -10,71 +10,74 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-#include <windows.h>
 #include "winutil.h"
 
-using namespace std;
+#include <Windows.h>
```
375gnu, line 15: this will not work with mingw crosscompiler
```suggestion
#include <windows.h>
```

375gnu (changes requested):

src/celutil/winutil.cpp
ajtribick: We have a mixture of `<windows.h>` and `Windows.h` in the source code, so MinGW cross-compilation doesn't seem to be something we support.

375gnu: I even posted brief manual for this - https://github.com/CelestiaProject/Celestia/wiki/crosscompiling-using-mingw. Maybe it's quite outdated

375gnu (approved):

PR #2009, opened 2023-12-12
Get number format information from ICU

levinli303:
symbol and attributes for number formatting based on std::locale seems to be inaccurate on certain platforms.

for example: de_DE locale on Mac says grouping is not used. value returned from grouping() is 0x7F.

### DIFF
--- a/src/celutil/formatnum.cpp
+++ b/src/celutil/formatnum.cpp
@@ -29,7 +29,6 @@
 #include <unicode/decimfmt.h>
 #include <unicode/ustring.h>
 #endif
-#include <celutil/gettext.h>
 #endif
 
 using namespace std::string_view_literals;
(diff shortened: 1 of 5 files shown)


src/celutil/formatnum.cpp
```diff
@@ -146,8 +159,66 @@ parseExponential(fmt::memory_buffer& buffer, ParseExpResult& result)
...
+    return U_SUCCESS(status);
+}
+#endif
+
 } // end unnamed namespace
 
+#ifdef USE_ICU
+NumberFormatter::NumberFormatter()
+{
+    const char *orig = N_("LANGUAGE");
```
ajtribick, line 195: I don't think we should use this mechanism - should pass `nullptr` as the locale to specify the default locale.

src/celutil/formatnum.cpp
```diff
@@ -146,8 +159,66 @@ parseExponential(fmt::memory_buffer& buffer, ParseExpResult& result)
...
+    UErrorCode status = U_ZERO_ERROR;
+    UNumberFormat* numFormat = unum_open(UNUM_DECIMAL, nullptr, 0, lang == orig ? "en" : lang, nullptr, &status);
+    if (U_FAILURE(status))
+        return;
+
+    getNumberSymbol(numFormat, UNUM_DECIMAL_SEPARATOR_SYMBOL, m_decimal);
+
+    // attribute == -1 means the attribute is missing
+    int32_t groupingUsed = unum_getAttribute(numFormat, UNUM_GROUPING_USED);
+    int32_t groupingSize = unum_getAttribute(numFormat, UNUM_GROUPING_SIZE);
```
ajtribick, line 207:
```suggestion
    std::int32_t groupingUsed = unum_getAttribute(numFormat, UNUM_GROUPING_USED);
    std::int32_t groupingSize = unum_getAttribute(numFormat, UNUM_GROUPING_SIZE);
```

src/celutil/formatnum.cpp
```diff
@@ -146,8 +159,66 @@ parseExponential(fmt::memory_buffer& buffer, ParseExpResult& result)
...
+    if (groupingUsed != -1 && groupingUsed != 0 &&
+        groupingSize != -1 && groupingSize != 0 &&
+        getNumberSymbol(numFormat, UNUM_GROUPING_SEPARATOR_SYMBOL, m_thousands))
+    {
+        m_grouping.push_back(static_cast<char>(groupingSize));
+
+        int32_t secondaryGroupingSize = unum_getAttribute(numFormat, UNUM_SECONDARY_GROUPING_SIZE);
```
ajtribick, line 214:
```suggestion
        std::int32_t secondaryGroupingSize = unum_getAttribute(numFormat, UNUM_SECONDARY_GROUPING_SIZE);
```

ajtribick (changes requested):

src/celutil/formatnum.cpp
levinli303: changed to use nullptr instead

src/celutil/formatnum.cpp
```diff
@@ -192,26 +191,23 @@ bool getNumberSymbol(const UNumberFormat *numFormat, UNumberFormatSymbol symbol,
 #ifdef USE_ICU
 NumberFormatter::NumberFormatter()
 {
-    const char *orig = N_("LANGUAGE");
-    const char *lang = _(orig);
-
     UErrorCode status = U_ZERO_ERROR;
-    UNumberFormat* numFormat = unum_open(UNUM_DECIMAL, nullptr, 0, lang == orig ? "en" : lang, nullptr, &status);
+    UNumberFormat* numFormat = unum_open(UNUM_DECIMAL, nullptr, 0, nullptr, nullptr, &status);
```
375gnu, line 195: Idea was to have 
```c++
NumberFormatter::NumberFormatter(const char *locale)
{
 UErrorCode status = U_ZERO_ERROR;
UNumberFormat* numFormat = unum_open(UNUM_DECIMAL, locale != nullptr ? locale : "en", 0, nullptr, nullptr, &status);
...
```

375gnu (changes requested):

src/celutil/formatnum.cpp
levinli303: @ajtribick using ICU's default locale does have some a downside. The one that i notice is that on Windows, ICU's default locale does not respect the "LANG" environment variable while gettext respects it.

So probably still a better idea to use "LANGUAGE" from .po file? The only downside for I can think of using this is that we cannot adapt to some differences in locales like between en_GB and en_US for date formatting since en is passed in.

ajtribick: The LANG environment variable isn't the usual way to do language changes on Windows, so that isn't necessarily fatal. If we need LANG support on Windows, then would probably make sense to do a call to [uloc_setDefault](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/uloc_8h.html#aa86430509b310792443d05e5bda643b3) at startup if this environment variable is set.

levinli303: then I'm keeping this part of code as is @375gnu . it will be up to frontends to decide whether uloc_setDefault should be called or not.

375gnu: okay

ajtribick (approved):

375gnu (approved):

Issue #559, opened 2019-12-05
Endless loop in cmod::Mesh::pick()


375gnu:
With large meshes (like cartrite's 3dmars) `cmod::Mesh::pick(const Eigen::Vector3d&, const Eigen::Vector3d&, cmod::Mesh::PickResult*)` goes into endless loop.

UnknowableCoder:
Sorry, I was looking through this repo just so see if I could contribute to something and I spotted this bug and wanted to see if I could help. Took a look at the code in celmodel/mesh.cpp, is there not a problem with i0 not being updated when primType == TriFan?

In lines 432 to 441, we have:

```
                else // primType == TriFan
                {
                    index += 1;
                    if (index < nIndices)
                    {
                        index += 1;
                        i1 = i2;
                        i2 = group->indices[index];
                    }
                }
```

Shouldn't the second `index += 1;` (line 437) be some update of i0?

I'm not entirely sure of all the specifics of the mesh storage formats, but it seems unreasonable to my (ignorant) understanding that `i0` does not change in this codepath, and the increment of `index` on the condition `index < nIndices` seems like an error (if `index == nIndices - 1`, we'd end up accessing `group->indices[nIndices];`).

375gnu:
[This image](https://upload.wikimedia.org/wikipedia/commons/5/59/Trianglefan.png) shows Triangle Fan. `A` is `i0`, `B`, `C`, `D`, etc are `i1` and `i2`. So `i0` is common and no need to update it.

> and the increment of `index` on the condition `index < nIndices` seems like an error (if `index == nIndices - 1`, we'd end up accessing `group->indices[nIndices]`

Yes, this code is very fragile and potentially can cause UB.

UnknowableCoder:
Thank you very much for the explanation and sorry for any inconvenience...

PR #2012, opened 2023-12-14
possible fix for #559

eehikki:
check for out-of-boundary was added to TriFan branch in ```Mesh::pick(const Eigen::Vector3d& rayOrigin, const Eigen::Vector3d& rayDirection, PickResult* result)``` method from mesh.cpp; this PR is intended to fix [#559](https://github.com/CelestiaProject/Celestia/issues/559) issue

### DIFF
--- a/src/celmodel/mesh.cpp
+++ b/src/celmodel/mesh.cpp
@@ -502,9 +502,12 @@ Mesh::pick(const Eigen::Vector3d& rayOrigin, const Eigen::Vector3d& rayDirection
                 else // primType == TriFan
                 {
                     index += 1;
-                    if (index < nIndices)
+                    if (index < nIndices - 1)
                     {
                         index += 1;
+                        i1 = i2;
+                        i2 = group.indices[index];
+                    } else if (index == nIndices - 1) {
                         i1 = i2;
                         i2 = group.indices[index];
                     }
--- a/src/celmodel/mesh.cpp
+++ b/src/celmodel/mesh.cpp
@@ -508,7 +508,8 @@ Mesh::pick(const Eigen::Vector3d& rayOrigin, const Eigen::Vector3d& rayDirection
                         i1 = i2;
                         i2 = group.indices[index];
                     }
-                    else if (index == nIndices - 1) {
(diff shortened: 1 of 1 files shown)


src/celmodel/mesh.cpp
```diff
@@ -502,11 +502,14 @@ Mesh::pick(const Eigen::Vector3d& rayOrigin, const Eigen::Vector3d& rayDirection
                 else // primType == TriFan
                 {
                     index += 1;
-                    if (index < nIndices)
+                    if (index < nIndices - 1)
                     {
                         index += 1;
                         i1 = i2;
                         i2 = group.indices[index];
+                    } else if (index == nIndices - 1) {
```
375gnu, line 510:
```suggestion
                    }
                    else if (index == nIndices - 1) {
```
Please adhere to the current code style.

375gnu:
Let us test this a bit. And thanks for contribution

src/celmodel/mesh.cpp
```diff
@@ -502,12 +502,16 @@ Mesh::pick(const Eigen::Vector3d& rayOrigin, const Eigen::Vector3d& rayDirection
...
                 {
                     index += 1;
-                    if (index < nIndices)
+                    if (index < nIndices - 1)
                     {
                         index += 1;
                         i1 = i2;
                         i2 = group.indices[index];
                     }
+                    else if (index == nIndices - 1) {
```
375gnu, line 511:
```suggestion
                    else if (index == nIndices - 1)
                    {
```

375gnu:
Unfortunately this change is wrong. 1) model used in #559 uses triangle list but this PR amends code working with triangle fan. 2) original code is broken but this PR doesn't fix this, see #2013 .

eehikki:
Can you give me download link for model that causes [#559](https://github.com/CelestiaProject/Celestia/issues/559)? I want to reproduce this behavior for myself

375gnu:
https://celestiaproject.space/forum/download/file.php?id=10657&filename=3dmars-h344.zip

PR #2015, opened 2023-12-15
Fix some issues found by sonar

375gnu:
* nullptr deref
* resource leak

### DIFF
--- a/src/celestia/view.cpp
+++ b/src/celestia/view.cpp
@@ -21,16 +21,16 @@ using celestia::util::GetLogger;
 namespace celestia
 {
 
-View::View(View::Type _type,
-           Observer *_observer,
-           float _x, float _y,
-           float _width, float _height) :
-    type(_type),
-    observer(_observer),
-    x(_x),
-    y(_y),
-    width(_width),
-    height(_height)
+View::View(View::Type type,
+           Observer *observer,
+           float x, float y,
+           float width, float height) :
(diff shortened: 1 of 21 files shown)


src/celestia/view.cpp
```diff
@@ -124,15 +113,10 @@ View::walkTreeResizeDelta(View* v, float delta, bool check)
             return false;
         if (check)
             return true;
-        v->height = (float) newSize;
+        v->height = newSize;
+        v->y = p->y + (v->y - p->y) * ratio;
         if (sign == 1)
```
levinli303, line 118: based on the old code here it should be "sign != 1"

src/celestia/view.cpp
```diff
@@ -142,15 +126,10 @@ View::walkTreeResizeDelta(View* v, float delta, bool check)
             return false;
         if (check)
             return true;
-        v->width = (float) newSize;
+        v->width = newSize;
+        v->x = p->x + (v->x - p->x) * ratio;
         if (sign == 1)
```
levinli303, line 131: same here.

levinli303 (changes requested):

src/celestia/view.cpp
375gnu: oh, when i stop doing dump mistakes?

levinli303 (approved):

Issue #1966, opened 2023-11-04
Switch the Windows front-end to Unicode APIs


ajtribick:
Currently the Windows front-end is using ANSI APIs, this limits the ability to support certain languages (e.g. Georgian, recently added).

If we still want to keep the Windows front-end around, we should update it to use the wide-character APIs instead.

375gnu:
From https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page:

Win32 APIs often support both -A and -W variants.

-A variants recognize the ANSI code page configured on the system and support char*, while -W variants operate in UTF-16 and support WCHAR.

Until recently, Windows has emphasized "Unicode" -W variants over -A APIs. However, recent releases have used the ANSI code page and -A APIs as a means to introduce UTF-8 support to apps. If the ANSI code page is configured for UTF-8, then -A APIs typically operate in UTF-8. This model has the benefit of supporting existing code built with -A APIs without any code changes.

375gnu:
-W API will also cause problems with gettext.

ajtribick:
Using the UTF-8 codepage would be the simplest option, although I believe this requires a minimum of Windows 10 version 1903.

ajtribick:
I have some work in progress on converting the win32 UI. It turns out that gettext isn't too much of an issue since we're already having to convert from UTF8 to current codepage.

PR #2017, opened 2023-12-17
Support wide-character APIs in the Windows front-end

ajtribick:
Relates to #1966

- Add support library to handle TCHAR strings/string views
- Add support library to parse floating point values from TCHAR data
- Move classes to celestia::win32 namespace
- Get month names from locales
- Split up winmain
- Reduce usages of global variables
- Prefer wide APIs for file-handling operations even in ANSI builds

The next stage would be to switch the .rc files themselves to use UTF-16 rather than codepages, but I want to do that as a more general update to the translation process.


src/celestia/scriptmenu.cpp
```diff
@@ -35,47 +44,60 @@ static void process(const fs::path& p, vector<ScriptMenuItem>* menuItems)
...
+    ScriptMenuItem& item = menuItems.emplace_back();
+    item.title = p.filename().string();
+    item.filename = p;
 
-        string firstLine(firstLineBuf, count);
-        auto titlePos = firstLine.find(TitleTag);
+    // Read the first line, handling various newline conventions
+    std::array<char, 512> buffer;
+    in.read(buffer.data(), buffer.size());
+    std::string_view line;
```
375gnu, line 58: move this declaration closer to usage

src/celestia/scriptmenu.cpp
```diff
@@ -35,47 +44,60 @@ static void process(const fs::path& p, vector<ScriptMenuItem>* menuItems)
...
-            item.title = firstLine.substr(titlePos);
-        }
-        else
-        {
-            // No title tag--just use the filename
-            item.title = p.filename().string();
-        }
-        menuItems->push_back(item);
-    }
+    // Only consider the first line
```
375gnu, line 64: it seems that code tries to implement https://en.cppreference.com/w/cpp/io/basic_istream/getline

src/celestia/win32/odmenu.cpp
```diff
@@ -99,63 +192,62 @@ bool ODMenu::Init(HWND hOwnerWnd, HMENU hMenu)
...
-    int i;
     MENUITEMINFO miInfo;
-
-    i = 0;
     miInfo.cbSize = sizeof(MENUITEMINFO);
     miInfo.fMask = MIIM_SUBMENU | MIIM_DATA;
-    while(GetMenuItemInfo(hMenu, i, TRUE, &miInfo))
+
+    int i = 0;
+    while (GetMenuItemInfo(hMenu, i, TRUE, &miInfo))
```
375gnu, line 233: for ?

375gnu (approved):

PR #2019, opened 2023-12-17
[gtk] fix issues with the eclipse dialog

375gnu:
* Body::getName() returns a new std::string, not a reference, so planet and satellite end up having the same value with g++, but in general case they point to garbage.
 * eclipseGoto function expects that entry 11 has a pointer to a receiver.

### DIFF
--- a/src/cel3ds/3dsread.cpp
+++ b/src/cel3ds/3dsread.cpp
@@ -442,7 +442,7 @@ bool readMeshMatrix(std::istream& in, std::int32_t contentSize, M3DTriangleMesh&
         return false;
     }
 
-    float elements[12];
+    std::array<float, 12> elements;
     for (std::size_t i = 0; i < 12; ++i)
     {
         if (!util::readLE<float>(in, elements[i]))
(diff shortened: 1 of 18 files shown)


ajtribick (approved):

src/celastro/date.h
```diff
@@ -84,10 +90,30 @@ class Date
...
+    char tzname[8]{ 'U', 'T', 'C', '\0' }; // timezone name
     double seconds{ 0.0 };
 };
 
+inline Date operator-(const Date &a, const Date &b)
+{
+    return Date(static_cast<double>(a) - static_cast<double>(b));
+}
+
+inline Date operator+(const Date &a, const Date &b)
```
ajtribick, line 102: I'm not sure about this operator - probably would make more sense if the right-hand side were a time interval type rather than a date.

src/celastro/date.h
```diff
@@ -84,10 +90,30 @@ class Date
...
+{
+    return Date(static_cast<double>(a) + static_cast<double>(b));
+}
+
+inline Date operator/(const Date &a, double b)
```
ajtribick, line 107: Divison and multiplication are definitely things that would make sense for time intervals rather than dates... better to not conflate the two.

src/celastro/date.h
```diff
@@ -84,10 +90,30 @@ class Date
     int minute{ 0 };
     int wday{ 0 };               // week day, 0 Sunday to 6 Saturday
     int utc_offset{ 0 };         // offset from utc in seconds
-    std::string tzname{ "UTC" }; // timezone name
+    char tzname[8]{ 'U', 'T', 'C', '\0' }; // timezone name
```
ajtribick, line 93: use `std::array<char, 8>`?

src/celastro/date.cpp
```diff
@@ -470,7 +463,7 @@ TDBtoLocal(double tdb)
     d.wday = localt.tm_wday;
 #if defined(__GNUC__) && !defined(_WIN32)
     d.utc_offset = static_cast<int>(localt.tm_gmtoff);
-    d.tzname = localt.tm_zone;
+    strncpy(d.tzname, localt.tm_zone, sizeof(d.tzname)-1);
```
ajtribick, line 466: can't remember if the initializer will zero out the array when it is created, maybe need to also set the last element to 0 just in case.

375gnu:
i decided to remove Date changes for now

ajtribick (approved):

Issue #2016, opened 2023-12-16
Normalize object's point color


JiliTheSpaceboy:
Currently, the color of the point source directly uses the Color parameter. This means that if the Color parameter is defined to have Value of less than 1 (such as [ 0.3 0.3 0.3 ], the object's point would appear dark as well. This is not the case in real life, where measurement of point source only yields 'normalized' colors. Therefore, I believe the program should normalize the Color parameter when drawing the point. This would allow us to use BlendTexture with any arbitrary color to tint a single color-calibrated texture to obtain a realistic-colored texture, instead of having to manually calibrate the brightness of the texture, make multiple copies of the same texture with diffferent brightnesses, and calculate normalized colors. (If the color is set to [ 0 0 0 ], then the normalized color should be set to [ 1 1 1 ].)

There could also be a 'toggle' of this function to enable/disable normalizing. The normalization should be **ON by default** as it is how real life objects work.

EDIT 1: The normalized colors should also apply on planetshine, as albedo already controls the brightness of the reflected light, in a similar manner to point source brightness.

pedro-fixingstuff:
An alternative to having a toggle parameter (which was suggested as a workaround for some old add-ons which use pitch black colors to make objects invisible as points) is to not apply the normalization if _and only if_ the color is pitch black [ 0 0 0 ]. After all, in theory, such an object should not glow at all...

PR #2021, opened 2023-12-19
Normalize point source color

JiliTheSpaceboy:
See Issue #2016

Pedro noted that old addons use [ 0 0 0 ] in order for object to not get rendered as points; this is handled by the condition at the beginning, which excludes such objects (and so they do not get rendered anyway).

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -2968,17 +2968,16 @@ void Renderer::renderPlanet(Body& body,
 
     if (body.isVisibleAsPoint())
     {
-        if (float maxCoeff = body.getSurface().color.toVector3().maxCoeff(); maxCoeff > 0.0f)
+        if (float maxCoeff = body.getSurface().color.toVector3().maxCoeff(); maxCoeff > 0.0f) // ignore [ 0 0 0 ]; used by old addons to make objects not get rendered as point
         {
-    {
-        renderObjectAsPoint(pos,
-                            body.getRadius(),
-                            appMag,
-                            discSizeInPixels,
-                            body.getSurface().color / maxCoeff,
-                            false, false, m);
+            renderObjectAsPoint(pos,
+                                body.getRadius(),
+                                appMag,
+                                discSizeInPixels,
+                                body.getSurface().color / maxCoeff, // normalize point color; 'darkness' is handled by size of point determined by GeomAlbedo.
+                                false, false, m);
+        }
     }
-}
 
 
 void Renderer::renderStar(const Star& star,
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -2978,6 +2978,7 @@ void Renderer::renderPlanet(Body& body,
                                 false, false, m);
         }
     }
+}
 
 
 void Renderer::renderStar(const Star& star,
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -2966,13 +2966,13 @@ void Renderer::renderPlanet(Body& body,
         }
     }
 
-    if (body.isVisibleAsPoint())
+    if (body.isVisibleAsPoint() && body.getSurface().color.toVector3().maxCoeff() > 0.0f)
     {
         renderObjectAsPoint(pos,
                             body.getRadius(),
                             appMag,
                             discSizeInPixels,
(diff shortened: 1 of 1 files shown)


src/celengine/render.cpp
```diff
@@ -2966,13 +2966,13 @@ void Renderer::renderPlanet(Body& body,
...
 
-    if (body.isVisibleAsPoint())
+    if (body.isVisibleAsPoint() && body.getSurface().color.toVector3().maxCoeff() > 0.0f)
     {
         renderObjectAsPoint(pos,
                             body.getRadius(),
                             appMag,
                             discSizeInPixels,
-                            body.getSurface().color,
+                            body.getSurface().color / body.getSurface().color.maxCoeff(),
```
ajtribick, line 2975:
```suggestion
                            body.getSurface().color / body.getSurface().color.toVector3().maxCoeff(),
```

src/celengine/render.cpp
```diff
@@ -2966,13 +2966,13 @@ void Renderer::renderPlanet(Body& body,
         }
     }
 
-    if (body.isVisibleAsPoint())
+    if (body.isVisibleAsPoint() && body.getSurface().color.toVector3().maxCoeff() > 0.0f)
```
375gnu, line 2969:
```suggestion
    if (body.isVisibleAsPoint())
    {
        if (float maxCoeff = body.getSurface().color.toVector3().maxCoeff(); maxCoeff > 0.0f)
        {
```

src/celengine/render.cpp
```diff
@@ -2966,13 +2966,13 @@ void Renderer::renderPlanet(Body& body,
...
     {
         renderObjectAsPoint(pos,
                             body.getRadius(),
                             appMag,
                             discSizeInPixels,
-                            body.getSurface().color,
+                            body.getSurface().color / body.getSurface().color.toVector3().maxCoeff(),
```
375gnu, line 2975:
```suggestion
                            body.getSurface().color / maxCoeff,
```

src/celengine/render.cpp
JiliTheSpaceboy: That one is meant to avoid causing the program to divide by 0, but maybe there is a better way to do this.

375gnu (changes requested):
the change should also contain a comments
1. "many addons use black color to blah-blah-blah so we skip such objects here"
2. "normalize color because blah-blah-blah"

JiliTheSpaceboy:
> the change should also contain comments
> 
> 1. "many addons use black color to blah-blah-blah so we skip such objects here"
> 2. "normalize color because blah-blah-blah"

True, will do that real quick

src/celengine/render.cpp
```diff
@@ -2968,14 +2968,16 @@ void Renderer::renderPlanet(Body& body,
...
+        {
+            renderObjectAsPoint(pos,
+                                body.getRadius(),
+                                appMag,
+                                discSizeInPixels,
+                                body.getSurface().color / maxCoeff, // normalize point color; 'darkness' is handled by size of point determined by GeomAlbedo.
+                                false, false, m);
+        }
     }
-}
```
375gnu, line 2978: return this bracket back

375gnu (changes requested):

src/celengine/render.cpp
```diff
@@ -2968,12 +2968,15 @@ void Renderer::renderPlanet(Body& body,
...
-                            discSizeInPixels,
-                            body.getSurface().color,
-                            false, false, m);
+        if (float maxCoeff = body.getSurface().color.toVector3().maxCoeff(); maxCoeff > 0.0f) // ignore [ 0 0 0 ]; used by old addons to make objects not get rendered as point
+        {
+            renderObjectAsPoint(pos,
+                                body.getRadius(),
+                                appMag,
+                                discSizeInPixels,
+                                body.getSurface().color / maxCoeff, // normalize point color; 'darkness' is handled by size of point determined by GeomAlbedo.
```
375gnu, line 2977:
```suggestion
                                body.getSurface().color * (1.0f / maxCoeff), // normalize point color; 'darkness' is handled by size of point determined by GeomAlbedo.
```

375gnu (approved):

PR #2022, opened 2023-12-20
Make mouse drag work with pick rays

### LINKED ISSUE #1344: Different sensitivity along the axes on elongated screens
**Describe the bug**
The sensitivity of object rotation / FOV movement depends on which axis to move the cursor/finger. It seems now the sensitivity along the axis is proportional to the size of the screen along the axis.

**Expected behavior**
I expect cursor/finger movement and field of view movement to be equal.

**Screenshots**
https://user-images.githubusercontent.com/44873271/151985175-796dc0fa-65bf-4d3b-96be-514440141081.mp4

**Desktop:**
 - OS: at least Windows and Android
 - Frontend: at least Qt, native Win, UWP
 - Version: at least 1.6, 1.7

levinli303:
Fixes #1344 the view moves exactly with the mouse cursor.


src/celengine/observer.cpp
```diff
@@ -46,6 +50,26 @@ static Vector3d slerp(double t, const Vector3d& v0, const Vector3d& v1)
     return (cos(theta * t) * u + sin(theta * t) * v) * math::lerp(t, r0, r1);
 }
 
+Eigen::Vector3d getNearIntersectionPointOrClosetPoint(const Eigen::ParametrizedLine<double, 3>& ray,
+                                                  const math::Sphered& sphere)
```
375gnu, line 54:
```suggestion
Eigen::Vector3d getNearIntersectionPointOrClosetPoint(const Eigen::ParametrizedLine<double, 3>& ray,
                                                      const math::Sphered& sphere)
```

src/celengine/observer.cpp
```diff
@@ -909,6 +933,40 @@ void Observer::orbit(const Selection& selection, const Quaternionf& q)
...
+ *  a direction to another
+ */
+void Observer::orbit(const Selection& selection, const Eigen::Vector3f &from, const Eigen::Vector3f &to)
+{
+    Selection center = frame->getRefObject();
+    if (center.empty() && !selection.empty())
+        center = selection;
+
+    if (center.empty())
+        return;
```
375gnu, line 946:
```suggestion
    if (center.empty())
    {
        if (selection.empty())
            return;
        center = selection;
    }
```

src/celestia/celestiacore.cpp
```diff
@@ -2216,6 +2208,22 @@ void CelestiaCore::renderOverlay()
...
 
 
+Eigen::Vector3f CelestiaCore::getPickRay(float x, float y, const celestia::View *view)
+{
+    float pickX;
+    float pickY;
+    float aspectRatio = static_cast<float>(metrics.width) / static_cast<float>(metrics.height);
+    view->mapWindowToView(x / static_cast<float>(metrics.width),
+                                               y / static_cast<float>(metrics.height),
+                                               pickX, pickY);
```
375gnu, line 2218: invalid indentation. tabs?

levinli303: just copy paste. fixed.

375gnu:
It behaves... weird. I'll try to record a video to show.

375gnu:
Unfortunately, video recording is broken. Dragging with the left mouse button is ok, but with the right one planets rotate different angles (small angle, then large, and so on), with random direction (i drag the mouse right and the planet rotates right, then up, then left, etc).

levinli303:
> but with the right one planets rotate different angles (small angle, then large, and so on), with random direction (i drag the mouse right and the planet rotates right, then up, then left, etc).

this is expected if your drag starts or ends in a location that's not part of the surface of the sphere, because this approach try to keep surface follow the mouse. if the start or end location is not on the surface, this try to find the nearest point to match which will result in a somewhat behavior.

I updated the code to
if the dragging starts from the surface, use the new behavior making the intersection point follow mouse. and when the mouse is not on the surface anymore, there will be no movement. 
if the dragging starts from not the surface, the old behavior is kept.

375gnu:
It's still not smooth and there is no "endless" rotation when dragging starts from the surface.

levinli303:
> It's still not smooth and there is no "endless" rotation when dragging starts from the surface.

hmm on my end it seems smooth. for endless rotation I checked Google Earth as an example, when your drag leaves the surface there is some inertia at first but then stops responding to mouse moves as well.

I'm happy to keep the right drag behind a setting if you think it is necessary to keep the old behavior by default on the frontends hosted here.

375gnu:
Definitely we must not break desktop experience even if it helps mobile users. So hiding it behind `#ifdef` can be good idea.

levinli303:
added an option in CMakeLists "ENABLE_ACCURATE_DRAGGING"

375gnu:
Please change the option name, because on desktop it's absolutely not accurate.

levinli303:
> Please change the option name, because on desktop it's absolutely not accurate.

changed to ENABLE_RAY_BASED_DRAGGING

375gnu (approved):

PR #2024, opened 2023-12-23
Add support for zooming with a focus

levinli303:
Add an option ENABLE_FOCUS_ZOOMING to support zooming (shift + left drag) with a focus by changing orientation during zooming so that the focus location (where the drag started) is kept where it was.

Also adding support for mobile pinch gesture based on this. Pinch on mobile is basically a focus and scale so it can be easily translated in to zooming with focus.

### DIFF
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -489,16 +489,7 @@ void CelestiaCore::mouseButtonUp(float x, float y, int button)
         }
         else if (button == RightButton)
         {
-            float pickX, pickY;
-            float aspectRatio = static_cast<float>(metrics.width) / static_cast<float>(metrics.height);
-            viewManager->activeView()->mapWindowToView(x / static_cast<float>(metrics.width),
-                                                       y / static_cast<float>(metrics.height),
-                                                       pickX, pickY);
-            pickX *= aspectRatio;
-            if (isViewportEffectUsed)
-                viewportEffect->distortXY(pickX, pickY);
-
-            Vector3f pickRay = renderer->getProjectionMode()->getPickRay(pickX, pickY, viewManager->activeView()->getObserver()->getZoom());
+            Eigen::Vector3f pickRay = getPickRay(x, y, viewManager->activeView());
 
             Selection sel = sim->pickObject(pickRay, renderer->getRenderFlags(), obsPickTolerance);
             if (!sel.empty())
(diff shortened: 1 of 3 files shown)


375gnu:
Could you provide a test build for mobile users?

levinli303:
> Could you provide a test build for mobile users?

I intend to provide beta testing via Google Play/App Store after this change is merged.

src/celestia/celestiacore.h
```diff
@@ -401,11 +402,18 @@ class CelestiaCore // : public Watchable<CelestiaCore>
     void setLayoutDirection(celestia::LayoutDirection);
 
 private:
+    struct MouseLocation
```
375gnu, line 405: Actually, this type is not required, please use `Eigen::Vector2f` instead.

src/celestia/celestiacore.cpp
```diff
@@ -2267,15 +2254,59 @@ Eigen::Vector3f CelestiaCore::getPickRay(float x, float y, const celestia::View
...
+    observer->setFOV(newFOV);
+    setZoomFromFOV();
+
+    if (oldPickRay.has_value())
+    {
+        Eigen::Vector3f newPickRay = getPickRay(focus.value().x, focus.value().y, view);
+        observer->rotate(Eigen::Quaternionf::FromTwoVectors(oldPickRay.value(), newPickRay));
+    }
+
+    if ((renderer->getRenderFlags() & Renderer::ShowAutoMag))
```
375gnu, line 2303:
```suggestion
    if ((renderer->getRenderFlags() & Renderer::ShowAutoMag) != 0)
```

src/celestia/celestiacore.cpp
```diff
@@ -2267,15 +2254,59 @@ Eigen::Vector3f CelestiaCore::getPickRay(float x, float y, const celestia::View
...
+
+    projectionMode->setSize(windowWidth, windowHeight);
+    return pickRay;
+}
+
+void CelestiaCore::updateFOV(float newFOV, std::optional<MouseLocation> focus, const celestia::View *view)
+{
+    float minFOV = renderer->getProjectionMode()->getMinimumFOV();
+    float maxFOV = renderer->getProjectionMode()->getMaximumFOV();
+    Observer *observer = view->getObserver();
```
375gnu, line 2283: move this definition closer to its usage

375gnu (changes requested):

src/celestia/celestiacore.cpp
```diff
@@ -753,9 +751,7 @@ void CelestiaCore::pinchUpdate(float focusX, float focusY, float scale)
     const View *view = viewManager->activeView();
     float fov = view->getObserver()->getFOV();
 
-    MouseLocation focus;
-    focus.x = focusX;
-    focus.y = focusY;
+    Eigen::Vector2f focus { focusX, focusY };
```
375gnu, line 754: please use `{}` only in class declarations.

375gnu (approved):

PR #2028, opened 2023-12-26
Support pinch to change distance

levinli303:
Based on the feedback from mobile users, adding support for pinching to change distance (FOV zoom has a limit) to match old zooming behavior in the default Object Mode.

### DIFF
--- a/src/celengine/observer.cpp
+++ b/src/celengine/observer.cpp
@@ -977,41 +977,38 @@ bool Observer::orbit(const Selection& selection, const Eigen::Vector3f &from, co
  */
 void Observer::changeOrbitDistance(const Selection& selection, float d)
 {
+    scaleOrbitDistance(selection, std::exp(-d));
+}
+
+
+void Observer::scaleOrbitDistance(const Selection& selection, float scale)
+{
     Selection center = frame->getRefObject();
-    if (center.empty() && !selection.empty())
+    if (center.empty())
     {
+        if (selection.empty())
+            return;
+
         center = selection;
         setFrame(frame->getCoordinateSystem(), center);
     }
 
-    if (!center.empty())
-    {
-        UniversalCoord focusPosition = center.getPosition(getTime());
+    UniversalCoord focusPosition = center.getPosition(getTime());
 
-        double size = center.radius();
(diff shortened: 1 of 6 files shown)


src/celengine/observer.cpp
```diff
@@ -976,42 +976,39 @@ bool Observer::orbit(const Selection& selection, const Eigen::Vector3f &from, co
...
-        if (currentDistance < minOrbitDistance)
-            minOrbitDistance = currentDistance * 0.5;
+    double span = currentDistance - minOrbitDistance;
+    double newDistance = minOrbitDistance + span / static_cast<double>(scale);
+    v = v * (newDistance / currentDistance);
 
-        if (currentDistance >= minOrbitDistance && naturalOrbitDistance != 0)
-        {
-            double r = (currentDistance - minOrbitDistance) / naturalOrbitDistance;
-            double newDistance = minOrbitDistance + naturalOrbitDistance * exp(log(r) + d);
```
levinli303, line 1008: `naturalOrbitDistance` is not really used so deleting it. It is just `double newDistance = minOrbitDistance + r * exp(d)` here.

src/celengine/observer.cpp
```diff
@@ -976,42 +976,39 @@ bool Observer::orbit(const Selection& selection, const Eigen::Vector3f &from, co
...
+    if (currentDistance < minOrbitDistance)
+        minOrbitDistance = currentDistance * 0.5;
 
-        if (currentDistance < minOrbitDistance)
-            minOrbitDistance = currentDistance * 0.5;
+    double span = currentDistance - minOrbitDistance;
+    double newDistance = minOrbitDistance + span / static_cast<double>(scale);
+    v = v * (newDistance / currentDistance);
 
-        if (currentDistance >= minOrbitDistance && naturalOrbitDistance != 0)
```
levinli303, line 1005: The only change to existing behavior is in the old code if radius == 0 (by checking naturalOrbitDistance != 0), distance would not be changed at all (which in my opinion does not make a lot of sense). In the new code, distance would still be scaled.

375gnu (approved):

PR #2030, opened 2023-12-26
Rewrite translation scripts

### LINKED ISSUE #1966: Switch the Windows front-end to Unicode APIs

ajtribick:
- Replace Perl/Makefiles with Python
- Use UTF-16 in Windows resource files
- Enable generation of Georgian language resource file

Closes #1966

Python scripts formatted using [Black](https://black.readthedocs.io/en/stable/)


375gnu (approved):
Bye-bye, Perl! (crying)

data-man:
> Replace Perl/Makefiles with Python

I'm wondering why not Lua. :)

PR #2032, opened 2023-12-26
Fix build with older g++

375gnu:


src/celestia/loadstars.cpp
```diff
@@ -9,9 +9,9 @@
 
 #include "loadstars.h"
 
-#include <filesystem>
```
375gnu, line 12: stupid `clangd`!

ajtribick (approved):

PR #2033, opened 2023-12-26
Move dateformatter to celutil

ajtribick:
Fits better with celutil than celengine.

Also use unique_ptr with custom deleter to manage ICU objects


src/celutil/dateformatter.cpp
```diff
@@ -96,24 +90,26 @@ UDateFormat *DateFormatter::getFormatter(bool local, astro::Date::Format format)
...
         dateStyle = UDAT_PATTERN;
         timeStyle = UDAT_PATTERN;
         break;
     }
 
     UErrorCode error = U_ZERO_ERROR;
-    auto formatter = udat_open(timeStyle, dateStyle, nullptr, local ? nullptr : u"UTC", -1, pattern, -1, &error);
+    formatters[index] = UniqueDateFormat(udat_open(timeStyle, dateStyle, nullptr,
+                                                   local ? nullptr : u"UTC", -1,
+                                                   pattern.data(), static_cast<std::int32_t>(pattern.size()),
```
375gnu, line 107: string_view is not recommended to use with C API because it can contain no '\0'. Let's stay on the safe size and use UChar *

375gnu (changes requested):

src/celutil/dateformatter.cpp
ajtribick: The function explicitly provides a length parameter for the case of non-null terminated strings.

https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/udat_8h.html#a27dd4504fd957101d616aaeaf9e8ec9c

"The number of characters in the pattern, or -1 if null-terminated."

ajtribick: The reason they provide a length parameter is explicitly to handle the case of non-null-terminated strings, so makes sense to use it where possible.

But we need to pass `nullptr` for one of the cases (I'm somewhat surprised the standard allows assigning `std::nullptr_t` to a `basic_string_view`), so makes sense to use pointer here

375gnu (approved):

PR #2047, opened 2023-12-30
Limb darkening

### LINKED ISSUE #1830: Non-sprite limb darkening implementation
The current [limb darkening](https://en.wikipedia.org/wiki/Limb_darkening) implementation seems to work by placing the sprite in front of the star, which increases the brightness in the center:
изображение
Compared to photographs, it is noticeably unnatural. Two solutions have been proposed on the Discord:
1. Use the empirical formula as in section 2.4 of [this article](https://arxiv.org/pdf/2101.06254.pdf).
2. Or, more correctly, make the effect dependent on the angle.

CubicApocalypse about the second way:
> I'd think angle-based would be relatively easy, using the same view angle/Fresnel effect that basically every game ever uses for pretty much anything
> the old `HazeColor` effect worked the same way, more or less

TTarrants:
> Fortunately the limb-darkening laws' coefficients are basically temperature (and metallicity) dependant, so it would be feasible to add them in a nice, "realistic" way in Celestia. M dwarfs having more pronounced limb-darkening than early-type stars, etc.

375gnu:
published for testing and getting response.

image

With atmospheres it looks… weird.

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -125,8 +125,6 @@ static constexpr float MaxFarNearRatio      = 2000000.0f;
 
 static constexpr float MinRelativeOccluderRadius = 0.005f;
 
-static constexpr float CoronaHeight = 0.2f;
-
 // Size at which the orbit cache will be flushed of old orbit paths
 static constexpr unsigned int OrbitCacheCullThreshold = 200;
 // Age in frames at which unused orbit paths may be eliminated from the cache
@@ -1817,6 +1815,9 @@ static void renderSphereUnlit(const RenderInfo& ri,
         textures.push_back(ri.overlayTex);
     }
 
+    if (ri.isStar)
+        shadprop.lightModel = LightingModel::StarModel;
+
     // Get a shader for the current rendering configuration
     auto* prog = r->getShaderManager().getShader(shadprop);
     if (prog == nullptr)
@@ -1827,6 +1828,8 @@ static void renderSphereUnlit(const RenderInfo& ri,
     prog->textureOffset = 0.0f;
     prog->ambientColor = ri.color.toVector3();
     prog->opacity = 1.0f;
+    if (ri.isStar)
+        prog->eyePosition = ri.eyePos_obj;
 
     Renderer::PipelineState ps;
     ps.depthMask = true;
@@ -2399,6 +2402,7 @@ void Renderer::renderObject(const Vector3f& pos,
         }
         else
         {
+            ri.isStar = obj.isStar;
             renderSphereUnlit(ri, viewFrustum, planetMVP, this, m_lodSphere.get());
         }
     }
@@ -2987,19 +2991,21 @@ void Renderer::renderStar(const Star& star,
         surface.appearanceFlags |= Surface::ApplyBaseTexture;
         surface.appearanceFlags |= Surface::Emissive;
 
+        rp.isStar = true;
         rp.surface = &surface;
         rp.rings = nullptr;
         rp.radius = star.getRadius();
         rp.semiAxes = star.getEllipsoidSemiAxes();
         rp.geometry = star.getGeometry();
 
+#if 0 // disabled for limb darkening
         Atmosphere atmosphere;
 
         // Use atmosphere effect to give stars a fuzzy fringe
         if (star.hasCorona() && rp.geometry == engine::GeometryHandle::Invalid)
         {
             Color atmColor(color.red() * 0.5f, color.green() * 0.5f, color.blue() * 0.5f);
-            atmosphere.height = radius * CoronaHeight;
+            atmosphere.height = radius * 0.2f /* CoronaHeight */;
             atmosphere.lowerColor = atmColor;
             atmosphere.upperColor = atmColor;
             atmosphere.skyColor = atmColor;
@@ -3010,6 +3016,9 @@ void Renderer::renderStar(const Star& star,
         {
             rp.atmosphere = nullptr;
         }
+#else
+        rp.atmosphere = nullptr;
+#endif
 
         rp.orientation = star.getRotationModel()->orientationAtTime(observer.getTime()).cast<float>();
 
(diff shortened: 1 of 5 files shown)


src/celengine/shadermanager.cpp
```diff
@@ -2905,14 +2910,13 @@ ShaderManager::buildAtmosphereFragmentShader(const ShaderProperties& props)
 
     // Sum the contributions from each light source
     source += "vec3 color = vec3(0.0);\n";
-    source += "vec3 V = normalize(eyeDir);\n";
```
375gnu, line 2908: `eyeDir` is already normalized since we implemented per-pixel shading.

src/celengine/renderinfo.h
```diff
@@ -21,12 +21,14 @@ class Texture;
 
 struct RenderInfo
 {
-    Color color{ 1.0f, 1.0f, 1.0f };
```
375gnu, line 24: reodering is to decrease the struct size

src/celengine/render.h
```diff
@@ -460,12 +460,15 @@ class Renderer
         Surface* surface{ nullptr };
         const Atmosphere* atmosphere{ nullptr };
         RingSystem* rings{ nullptr };
+        LightingState::EclipseShadowVector* eclipseShadows{ nullptr };
+
+        Eigen::Quaternionf orientation{ Eigen::Quaternionf::Identity() };
```
375gnu, line 465: reodering is to decrease the struct size

Askaniy:
Is it just me, or is there a sprite (or something similar) in the center of the star that increases the brightness and causes some colors to overexpose?

Askaniy:
I also think that atmospheres, as a workaround for the glow effect, should be removed

375gnu:
> Is it just me, or is there a sprite (or something similar) in the center of the star that increases the brightness and causes some colors to overexpose?

Yes, there is a big quad (sprite) if front of the Sun.

375gnu:
That's how it looks with stars as scaled disks (in this mode the quad size is much smaller):
image

375gnu:
btw, in old HDR code atmosphere is disabled:
```c++
#ifndef USE_HDR
        Atmosphere atmosphere;

        // Use atmosphere effect to give stars a fuzzy fringe
        if (star.hasCorona() && rp.geometry == InvalidResource)
        {
            Color atmColor(color.red() * 0.5f, color.green() * 0.5f, color.blue() * 0.5f);
            atmosphere.height = radius * CoronaHeight;
            atmosphere.lowerColor = atmColor;
            atmosphere.upperColor = atmColor;
            atmosphere.skyColor = atmColor;

            rp.atmosphere = &atmosphere;
        }
        else
#endif
        rp.atmosphere = NULL;
```

JiliTheSpaceboy:
> That's how it looks with stars as scaled disks (in this mode the quad size is much smaller):
> image

What would this look like if used with a normal brightness star texture?

375gnu:
> What would this look like if used with a normal brightness star texture?


image

JiliTheSpaceboy:
> > What would this look like if used with a normal brightness star texture?
> 
> 
> 
> 
> 
> image
> 
> 
> 
> or even
> 
> image
> 
> 

Yup that looks lovely. Feels like we can actually use normal brightness textures in default now (and maybe add some kind support for outdated darkened textures if needed).

ajtribick (approved):

AstroChara:
Looks like this is mostly done; I believe we mainly have to get new star textures to replace the current textures which have been darkened (and which have license issues)?

375gnu:
> Looks like this is mostly done; I believe we mainly have to get new star textures to replace the current textures which have been darkened (and which have license issues)?

if i remember correctly, yes, textures are main blocker

src/celengine/shadermanager.cpp
```diff
@@ -2093,6 +2093,11 @@ ShaderManager::buildFragmentShader(const ShaderProperties& props)
         source += "gl_FragColor.rgb = gl_FragColor.rgb * scatterEx + scatterColor;\n";
     }
 
+    if (props.lightModel == LightingModel::StarModel)
+    {
+        source += "gl_FragColor.rgb = gl_FragColor.rgb - vec3(1.0 - NV) * vec3(0.56, 0.61, 0.72);\n";
```
pedro-fixingstuff, line 2098: Might be a good idea to write down how the coefficients were obtained.

AstroChara:
The new star textures are getting ready. I believe we'll want to resolve the remaining PR conflicts now.

src/celengine/render.cpp
pedro-fixingstuff, line 149: This variable's only other use is for the computation of the cull radius (line 4819). Since the corona's no longer being rendered, there's no reason to account it on culling, so it can be removed?

AstroChara: I imagine we might eventually want to re-implement the corona again. I wonder if it is possible to have it without drawing the glow on top of the star’s disc (since that’s the root of the problem)?

pedro-fixingstuff: Instead of drawing the glow sprite in front of stars, it can be drawn at the same position by abandoning the calculateQuadCenter function. I experimented with that, but some clipping issues appear, if a star is rendered off the display's center, probably due to it not being normal to the observer.

375gnu: @pedro-fixingstuff do you remember what variable you was talking about?

pedro-fixingstuff: `CoronaHeight`

src/celengine/shadermanager.cpp
375gnu: In the dock that I used I can find only 0.56:

> Limb-darkening is much greater in the violet and near ultraviolet than in the red. For example, at
a wavelength of 600 nm, u = 0.56, whereas at 320 nm u = 0.95.

But the source of 0.61 and 0.72 is lost. Maybe it's somewhere in discord / forum discussions.

src/celengine/render.cpp
pedro-fixingstuff: Addressed by #2567

src/celengine/shadermanager.cpp
AstroChara: I believe it might be a good idea to address this in another PR.

pedro-fixingstuff: For now we can leave this as-is. Figuring out which coeffs to use is a whole project of its own, and the ultimate solution might be to implement a lookup table as a function of temperature (and other parameters?).

pedro-fixingstuff (approved):

PR #2048, opened 2023-12-30
Fix too wide bitmask in Url which caused planets to be shown always

375gnu:
Plus cleanup without functional changes

With `master`:
1.  Start Celestia, so the Earth is visible
1. Disable all bodies in display options, the Earth is invisible
1. Copy URL
1. Paste it back
1. The Earth is visible again

This was caused by too wide RF_MASK which set bit 26 used in 1.7 to set planet visibility in a 1.6-compatible way.

### DIFF
--- a/src/celengine/render.h
+++ b/src/celengine/render.h
@@ -206,7 +206,6 @@ class Renderer
                                   ShowMinorMoons        |
                                   ShowAsteroids         |
                                   ShowComets            |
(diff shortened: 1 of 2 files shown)


ajtribick (approved):

375gnu:
I've realized it requires one more change

ajtribick (approved):

Issue #2053, opened 2024-01-02
Texture fonts in Lua userdata should have a __gc method


ajtribick:
Need to set the `__gc` method in the userdata metatable to destroy the `shared_ptr`, otherwise the reference count will never get decreased.

PR #2057, opened 2024-01-04
[celx] implement  __gc method for fonts

375gnu:
Closes: #2053

### DIFF
--- a/src/celscript/lua/celx_misc.cpp
+++ b/src/celscript/lua/celx_misc.cpp
@@ -271,6 +271,14 @@ static int font_tostring(lua_State* l)
     return celx.push("[Font]");
 }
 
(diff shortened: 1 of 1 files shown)


src/celscript/lua/celx_misc.cpp
```diff
@@ -271,13 +271,22 @@ static int font_tostring(lua_State* l)
     return celx.push("[Font]");
 }
 
+static int font_gc(lua_State* l)
+{
+    CelxLua celx(l);
+    auto *font = celx.getThis<std::shared_ptr<TextureFont>>();
+    *font = nullptr;
```
ajtribick, line 278:
```suggestion
    *font->~shared_ptr();
```

ajtribick (approved):

PR #2058, opened 2024-01-04
Various updates to celx_celestia

ajtribick:
- Prefer .find() over .count()/indexer when checking flag maps
- Use lua_setlstring to avoid unnecessary string allocations
- Avoid some cases of longjmp skipping string destructors


ajtribick:
Most of the duplicated stuff reflects what was there already. Not fixing it here.

375gnu (approved):

PR #2060, opened 2024-01-07
Parse paths as UTF-8

ajtribick:
The path constructor from `char` strings uses the system's current narrow character encoding, which on Windows will depend on the locale. Therefore we should construct paths from UTF-8 via `u8path`.

Also validate filenames in a cross-platform manner, to avoid illegal path sequences, Windows reserved device names, or breaking out of the directory structure.


375gnu (approved):
though i'm not nappy in dragging woe limitations to normal systems

PR #2062, opened 2024-01-12
Use smart pointers to track orbits and rotation models

ajtribick:
More changes that may eventually let us do re-loads of data...

- Cache custom orbits
- Use separate samples and trajectory caches to reduce duplication of samples at different interpolation/precision
- Use custom classes for trajectory/rotation model caches as resource handles aren't used


375gnu (approved):

375gnu:
Any numbers to show performance difference with the old and new code?

PR #2073, opened 2024-01-15
[qt] Fix startup crash with --url parameter

### LINKED ISSUE #2072: Strating celestia-qt with URL crashes the program
```
0x00007ffff700dd81 in CelestiaCore::setStartURL(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /redacted//libcelestia.so.1.7
(gdb) bt
#0  0x00007ffff700dd81 in CelestiaCore::setStartURL(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
    at /redacted/libcelestia.so.1.7
#1  0x000055555558b95d in CelestiaAppWindow::init(CelestiaCommandLineOptions const&) ()
#2  0x000055555557ec62 in main ()

```

ajtribick:
- Reorganize startup to avoid clobbering FOV when using startup URL
- Make several appwin/glwidget methods private, mark overrides

Fixes #2072

### DIFF
--- a/src/celestia/qt/qtappwin.cpp
+++ b/src/celestia/qt/qtappwin.cpp
@@ -10,7 +10,6 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
-
 #include "qtappwin.h"
 
 #include <map>
@@ -106,7 +105,6 @@
 #define CONFIG_DATA_DIR "./"
 #endif
 
-
 namespace
 {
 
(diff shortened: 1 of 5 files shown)


src/celestia/qt/qtappwin.cpp
```diff
@@ -479,6 +469,11 @@ CelestiaAppWindow::init(const CelestiaCommandLineOptions& options)
     timer->start();
 }
 
+void
+CelestiaAppWindow::startAppCore()
```
375gnu, line 473: just `start`?

ajtribick: Since there's already a constructor and an `init` method in terms of things that begin the object, I prefer to be slightly more explicit about precisely what this thing is starting.

375gnu (approved):

PR #2075, opened 2024-01-15
[qt] Update deep sky browser

ajtribick:
- Only show Type column for galaxies (irrelevant otherwise)
- Replace multiset with heap algorithms when populating the DSO list
- When building with Qt6, use QRegularExpression instead of legacy QRegExp
- Remove some unused methods from the DSO database

### DIFF
--- a/src/celengine/dsodb.cpp
+++ b/src/celengine/dsodb.cpp
@@ -310,13 +310,6 @@ bool DSODatabase::load(std::istream& in, const fs::path& resourcePath)
 }
 
 
-bool DSODatabase::loadBinary(std::istream&)
-{
-    // TODO: define a binary dso file format
-    return true;
-}
-
-
 void DSODatabase::finish()
 {
     buildOctree();
--- a/src/celengine/dsodb.h
+++ b/src/celengine/dsodb.h
@@ -38,7 +38,6 @@ class DSODatabase
     DSODatabase() = default;
     ~DSODatabase();
 
-
     DeepSkyObject* getDSO(const std::uint32_t) const;
     std::uint32_t size() const;
 
@@ -65,10 +64,7 @@ class DSODatabase
     void setNameDatabase(std::unique_ptr<DSONameDatabase>&&);
 
     bool load(std::istream&, const fs::path& resourcePath = fs::path());
-    bool loadBinary(std::istream&);
     void finish();
-
-    static DSODatabase* read(std::istream&);
 
     float getAverageAbsoluteMagnitude() const;
 
--- a/src/celestia/qt/qtdeepskybrowser.cpp
+++ b/src/celestia/qt/qtdeepskybrowser.cpp
@@ -14,8 +14,8 @@
 
 #include <algorithm>
 #include <cstring>
-#include <set>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include <Eigen/Core>
@@ -37,7 +37,11 @@
 #include <QModelIndexList>
 #include <QPushButton>
 #include <QRadioButton>
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
 #include <QRegExp>
+#else
+#include <QRegularExpression>
+#endif
 #include <QString>
 #include <QTreeView>
 #include <QVariant>
@@ -47,7 +51,6 @@
 #include <celengine/deepskyobj.h>
 #include <celengine/dsodb.h>
 #include <celengine/marker.h>
-#include <celengine/render.h>
 #include <celengine/selection.h>
 #include <celengine/simulation.h>
 #include <celengine/universe.h>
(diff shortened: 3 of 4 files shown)


dave-kaye:
The open and globular clusters do have types associated with them.  It is just that Celestia has not enumerated the types.

For the globular clusters:

> Classification of globular clusters from physical, structural parameters has been introduced in the 1920s by Harlow Shapley and Helen Sawyer Hogg ([Shapley, Sawyer (1927)](http://www.messier.seds.org/glob.html#shapley1927); [Shapley 1930](http://www.messier.seds.org/glob.html#shapley1930), II.5, pp. 11-14). They introduced Concentration Classes (sometimes referred to as Shapley-Sawyer Concentration Classes), named by numbers, typically denoted Roman, from I-XII, or 1-12, with I denoting the most concentrated or densest structure, XII the least concentrated. These concentration classes are given in these pages for each cluster. Quick overview for the 29 Messier Globular Clusters:
> 
> Class I (1): M75
> Class II (2): M2, M80
> Class III (1): M54
> Class IV (4): M15, M28, M62, M92
> Class V (7): M5, M13, M30, M53, M69, M70, M79
> Class VI (1): M3
> Class VII (2): M10, M22
> Class VIII (3): M9, M14, M19
> Class IX (3): M4, M12, M72
> Class X (3): M56, M68, M107
> Class X-XI (1): M71
> Class XI (1): M55
> Class XII (0)
http://www.messier.seds.org/glob.html 

For the open clusters:

> Open clusters are often typized according to a simple scheme which goes back to [Harlow Shapley](http://www.messier.seds.org/xtra/Bios/shapley.html), which describes richness and concentration roughly ([Shapley 1930](http://www.messier.seds.org/open.html#shapley1930)):
> 
> a Field Irregularities
> b Star Associations
> c Very loose and irregular clusters
> d Loose clusters
> e Intermediately rich compact clusters
> f Fairly rich compact clusters
> g Considerably rich and concentrated compact clusters
> 
> Another important and more sophisticated scheme was introduced by [R.J. Trumpler](http://www.messier.seds.org/xtra/Bios/trumpler.html) ([Trumpler 1930](http://www.messier.seds.org/open.html#trumpler1930)). This scheme consists of three parts, characterizing the cluster's degree of concentration, the range of brightness of its stars, and the richness, as follows:
> Concentration
> I Detached; strong concentration toward center
> II Detached; weak concentration toward center
> III Detached; no concentration toward center
> IV Not well detached from surrounding star field
> Range in Brightness
> 1 Small range in brightness
> 2 Moderate range in brightness
> 3 Large range in brightness
> Richness
> p Poor: Less than 50 stars
> m Moderately rich: 50 to 100 stars
> r Rich: More than 100 stars
> A "n" following the Trumpler class indicates that there is nebulosity associated with the cluster.
http://www.messier.seds.org/open.html

Perhaps someone who is more proficient in astronomy than I should look at classifying the clusters in the Celestia database rather than removing the type listing.  The references contain information on the types of numerous clusters.  But I don't know which ones would be the accepted type when multiple designations are given.  While I have not looked at the code recently, I seem to remember that it was possible to put the types into present data files and the code would need little/no revisions.  The open and globular are main identification parameters and are used to separate them.  Galaxies have a "type" parameter, which can be made to work for the clusters following the code base of the galaxies as an example.

ajtribick:
Those classifications are currently not in the Celestia data model. Currently all the field does is expose a non-translated constant string for all objects, except for galaxies. If we decide to add them to the Celestia data model in future, we can update the deep sky browser accordingly.

ajtribick:
This is also in line with the handling in the celx object:getinfo method, which only exposes this field for galaxies (under the name hubbleType)

dave-kaye:
You are adding a boolean `showType` to set the number of columns in the deep sky browser.  This can be readily updated if cluster types are added in the future.  The key is whether the modifications that you are doing to the browser filters will lock the browser into only being able to filter galaxy types and not other object types.

For the browser output in Qt, the types appear to be several enumerated arrays for galaxies.  These structures could be duplicated for clusters.  There are several main small routines (getType, setType, getDescription, etc.) that interact with these arrays in the galaxy object.  They exist in the cluster codes and you are not modifying them - so long as you do not remove them as unused code, it should be possible to mimic these galaxy codes for the cluster objects.  Hopefully these would be the only modifications necessary and the generic deep sky object codes would operate successfully.  I don't know about the celx codes and the datafile reading codes.

Steinike's work on ngc/ic is used as the starting point for the galaxy data files.  The work also contains some Concentration and Trumpler classifications for clusters.  It would be a starting point for updating the data files.  @GurrenLagannTSS recently updated the cluster data files and may have some opinions or available data.  Celestia has separate data files for galaxies, open clusters and globular clusters, so working on one would not affect the others. 

A possiblity is to make an issue to add Concentration Class types to the Globular Cluster object and a second issue to add Trumpler Classification types to the Open Cluster object.  Then mark them as future wishlist and they can be tracked.  Alternatively, it could be added to Celestia github discussion or Celestia forum for opinions and suggestions.

Of course, this would be of a lower priority and only as time/desire is available.

ajtribick:
> The key is whether the modifications that you are doing to the browser filters will lock the browser into only being able to filter galaxy types and not other object types.

No it doesn't. If someone wants to update the code, then they can do so. It is a very simple condition to change. At the moment, the field it is picking up is a hard-coded constant for other DSO types - it's not linked to anything in the dsc file. You would have to change the code elsewhere to make it do something useful.

GurrenLagannTSS:
> Steinike's work on ngc/ic is used as the starting point for the galaxy data files. The work also contains some Concentration and Trumpler classifications for clusters. It would be a starting point for updating the data files. @GurrenLagannTSS recently updated the cluster data files and may have some opinions or available data. Celestia has separate data files for galaxies, open clusters and globular clusters, so working on one would not affect the others.
>
> A possiblity is to make an issue to add Concentration Class types to the Globular Cluster object and a second issue to add Trumpler Classification types to the Open Cluster object. Then mark them as future wishlist and they can be tracked. Alternatively, it could be added to Celestia github discussion or Celestia forum for opinions and suggestions.

Hello, cluster updater here. The Shapley-Sawyer Concentration and Trumpler classifications would personally be nice to have in default. _However_, it seems hard to calculate SSCC and Trumpler classes without a good definition on what terms like "concentrated" and "[small/large] range" actually mean.

375gnu (approved):

PR #2078, opened 2024-01-25
Move NEAR_DIST, FAR_DIST to ProjectionMode

levinli303:
In set up of some rendering system, nearZ and farZ are provided by the system, so adding a method to support this usage.

### DIFF
--- a/src/celengine/projectionmode.h
+++ b/src/celengine/projectionmode.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include <tuple>
 #include <Eigen/Core>
 
 class ShaderManager;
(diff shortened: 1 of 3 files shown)


src/celengine/projectionmode.h
```diff
@@ -30,6 +30,7 @@ class ProjectionMode
     virtual ~ProjectionMode() = default;
 
     virtual Eigen::Matrix4f getProjectionMatrix(float nearZ, float farZ, float zoom) const = 0;
+    virtual void getDefaultDepthRange(float& nearZ, float& farZ) const;
```
375gnu, line 33:
```suggestion
    virtual std::tuple<float, float> getDefaultDepthRange() const;
```

src/celengine/render.cpp
```diff
@@ -1518,7 +1515,9 @@ void Renderer::render(const Observer& observer,
 
     // Set up the projection and modelview matrices.
     // We'll usethem for positioning star and planet labels.
-    buildProjectionMatrix(m_projMatrix, NEAR_DIST, FAR_DIST, observer.getZoom());
+    float nearZ, farZ;
+    projectionMode->getDefaultDepthRange(nearZ, farZ);
```
375gnu, line 1519:
```suggestion
    auto [nearZ, farZ] = projectionMode->getDefaultDepthRange();
```

src/celengine/projectionmode.h
375gnu: and add a comment why this method is required

375gnu (changes requested):

375gnu (approved):

PR #2083, opened 2024-01-30
Add support for getting measurement system from ICU

levinli303:


src/celestia/hud.cpp
```diff
@@ -22,6 +22,18 @@
...
+#include <icu.h>
+#elif defined(HAVE_WIN_ICU_SEPARATE_HEADERS)
+#include <icucommon.h>
+#include <icui18n.h>
+#else
+#include <unicode/uloc.h>
+#include <unicode/ulocdata.h>
+#endif
+#endif
+
```
375gnu, line 36: can we have celutil/icu.h file instead of duplicating this block in multiple files?

375gnu (changes requested):

PR #2084, opened 2024-02-01
Fix compilation with Windows ICU

ajtribick:
- Rename celutil/icu.h to includeicu.h to avoid recursive include
- Use pragma once

### DIFF
--- a/src/celutil/includeicu.h
+++ b/src/celutil/includeicu.h
@@ -7,6 +7,8 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
+#pragma once
+
 #ifdef USE_ICU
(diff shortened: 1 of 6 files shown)


src/celutil/includeicu.h
```diff
@@ -7,6 +7,8 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
+#pragma once
+
 #ifdef USE_ICU
 #ifdef HAVE_WIN_ICU_COMBINED_HEADER
 #include <icu.h>
```
ajtribick, line 14: turns out that this actually resulted in this file including itself.

levinli303 (approved):

PR #2088, opened 2024-02-07
Update static manager handling initialization

ajtribick:


ajtribick:
tempted to disable S6018 as a bad rule - no point initializing these things if the relevant code paths don't get called.

ajtribick:
NOSONAR because the S6018 rule would result in unconditional instantiation of the managers which may or may not be required depending on the catalogue files.

375gnu (approved):
looks like something that can be auto generated.

PR #2097, opened 2024-02-08
[ci] Restrict subset of qtbase installed with vcpkg

ajtribick:
Should make Windows library updates a little faster by avoiding building parts of Qt we don't need. Also set the host triplet to x86-windows for the relevant builds to avoid it being treated as cross-compilation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,9 +55,34 @@ jobs:
     - name: Install dependencies
       shell: pwsh
       run: |
-        vcpkg --triplet=${{matrix.platform}}-windows install --recurse `
-          cspice eigen3 ffmpeg[x264] fmt freetype gettext gperf libepoxy libjpeg-turbo libpng luajit `
-          qtbase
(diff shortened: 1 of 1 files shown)


ajtribick:
Less useful than I'd hoped because x86-windows builds the x64-windows qtbase with all default features and dependencies. :(

.github/workflows/ci.yml
```diff
@@ -55,9 +55,34 @@ jobs:
...
+          'libepoxy',
+          'libjpeg-turbo',
+          'libpng',
+          'luajit',
+          'qtbase[core,opengl,widgets,freetype,harfbuzz,icu,jpeg,png]'
+        )
+
+        # We treat x86 builds as native, other builds keep host as x64-windows.
+        # Currently this is equivalent to the platform triplet, but this would
+        # differ if we enable amd64-windows builds in future
```
ajtribick, line 75: stop typing amd when you mean arm

375gnu (approved):

PR #2100, opened 2024-02-08
Fix deprecation warnings in Qt6 build

ajtribick:


src/celestia/qt/qtglwidget.cpp
```diff
@@ -52,6 +53,19 @@ constexpr float DEFAULT_VISUAL_MAGNITUDE = 8.0f;
 constexpr Renderer::StarStyle DEFAULT_STAR_STYLE = Renderer::FuzzyPointStars;
 constexpr unsigned int DEFAULT_TEXTURE_RESOLUTION = medres;
 
+std::pair<float, float>
+mousePosition(const QMouseEvent& m, qreal scale)
+{
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+    return std::make_pair(static_cast<float>(m.x() * scale),
```
375gnu, line 60: you can write just `return {...}` instead

375gnu (approved):
i hate these "modern toolkits"

PR #2102, opened 2024-02-10
Load SPICE library dynamically at runtime

### LINKED ISSUE #410: Make spice support dynamically loadable
NASA's spice library has proprietary obscure licence which prevents its inclusion into GNU/Linux distributions and BSD ports. I.e. see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=841104.

Excerpt from https://naif.jpl.nasa.gov/naif/rules.html:

> **Obtaining the SPICE Toolkit**
> To meet NASA/JPL requirements and to better ensure getting the correct products, persons wishing a copy of the SPICE Toolkit should obtain it from this web site or from an official NASA flight project distribution. (Also see "Toolkit Redistribution" below.)
> 
> **Toolkit Redistribution**
> Simple redistribution of the complete Toolkit, such as from a mirror site, is prohibited without prior clearance from NAIF. However, including the SPICE Toolkit library modules and relevant SPICE Toolkit programs and allied User Guides as part of a package supporting a customer-built SPICE-based tool is entirely appropriate.

So while we can provide spice-enabled packages, users of e.g. Debian will not have spice enabled. So the alternative approach is to make spice-enabled code dynamically loadable.

ajtribick:
Resolves #410

- Due to license restrictions, do not compile or link the CSPICE library
- On Windows, load the cspice.dll file from the plugins directory
- On Apple, search for libcspice.dylib
- On Linux, search for libcspice.so

TODO;

- [ ] Test Windows
- [ ] Test Linux
- [ ] Confirm that this works on Mac
- [ ] Update installation instructions


src/celephem/spiceinterface.cpp
```diff
@@ -27,56 +33,152 @@ namespace
...
     return true;
 }
 
+std::unique_ptr<SpiceLibrary>
+LoadSpiceLibrary()
+{
+#if defined(_WIN32)
+    HMODULE handle = LoadLibraryExW(L"plugins\\cspice.dll", NULL, LOAD_LIBRARY_SEARCH_APPLICATION_DIR);
+#elif defined(__APPLE__)
+    void* handle = dlopen("libcspice.dylib", RTLD_LAZY | RTLD_LOCAL);
```
ajtribick, line 85: does this work on Mac?

ajtribick:
Should also update installation instructions if we decide to go down this route

src/celephem/spiceinterface.cpp
```diff
@@ -27,56 +33,152 @@ namespace
...
+    }
 
     return true;
 }
 
+std::unique_ptr<SpiceLibrary>
+LoadSpiceLibrary()
+{
+#if defined(_WIN32)
+    HMODULE handle = LoadLibraryExW(L"plugins\\cspice.dll", NULL, LOAD_LIBRARY_SEARCH_APPLICATION_DIR);
```
levinli303, line 83: On UWP we need to use [LoadPackagedLibrary](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-loadpackagedlibrary) for UWP. We can check for UWP [here](https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/misc-info).

ajtribick:
I've changed the interface a bit. There is now a `celestia::SetSpiceInterface` method that allows a front-end to supply the implementation. For a third-party front-end, build with the new CMake option `USE_SPICE_LOADER` set to `OFF` to avoid compiling the default loader, and supply the function pointers however you want.

src/celephem/spiceinterface.cpp
ajtribick: I've restructured the code to allow third-party front-ends to supply their own loading implementation.

ajtribick:
And looking at it, it seems the official distribution of the SPICE kernels only builds the static library also. Unfortunate.

PR #2109, opened 2024-02-14
Add support for backing framebuffer object with renderbuffer

levinli303:
Most likely viewport effects do not need to read from depth attachment so we can use render buffers when available.

### DIFF
--- a/src/celengine/framebuffer.h
+++ b/src/celengine/framebuffer.h
@@ -15,20 +15,22 @@
 class FramebufferObject
 {
  public:
-    enum
+    enum class AttachmentType
     {
-        ColorAttachment = 0x1,
-        DepthAttachment = 0x2
+        None,
+        Texture,
+        Renderbuffer,
     };
+
     FramebufferObject() = delete;
-    FramebufferObject(GLuint width, GLuint height, unsigned int attachments);
+    FramebufferObject(GLuint width, GLuint height, AttachmentType colorAttachment, AttachmentType depthAttachment);
     FramebufferObject(const FramebufferObject&) = delete;
     FramebufferObject(FramebufferObject&&) noexcept;
     FramebufferObject& operator=(const FramebufferObject&) = delete;
     FramebufferObject& operator=(FramebufferObject&&) noexcept;
     ~FramebufferObject();
 
-    static inline bool isSupported();
+    static bool isSupported();
     bool isValid() const;
     GLuint width() const
     {
@@ -40,8 +42,11 @@ class FramebufferObject
         return m_height;
     }
 
-    GLuint colorTexture() const;
-    GLuint depthTexture() const;
+    GLuint colorAttachment() const;
+    GLuint depthAttachment() const;
+
+    AttachmentType colorAttachmentType() const;
+    AttachmentType depthAttachmentType() const;
 
     bool bind();
     bool unbind(GLint oldfboId);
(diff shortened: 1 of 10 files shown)


375gnu:
let me have some time to check this

375gnu:
@levinli303 do you have any numbers to show benefit?

levinli303:
> @levinli303 do you have any numbers to show benefit?

I got time to test today, and sadly it seems to negatively impact the render time at least on iPad (to switch depth attachment from texture to renderbuffer in viewport effect)

For the demo script,
the average frame time (without viewport effect) is around 6.45ms
the average frame time (with viewport effect renderbuffer as depth attachment) is around 6.57ms
the average frame time (with viewport effect texture as depth attachment) is around 6.5ms

375gnu:
What resolution your ipad has?

src/celengine/framebuffer.h
```diff
@@ -40,28 +42,36 @@ class FramebufferObject
         return m_height;
     }
 
-    GLuint colorTexture() const;
-    GLuint depthTexture() const;
+    GLuint colorAttachment() const;
+    GLuint depthAttachment() const;
```
375gnu, line 46: I think this is not a proper update. `*Texture()` should return texture id (0 if not attached), for Renderbuffers you should add new methods.

src/celengine/framebuffer.h
```diff
@@ -15,20 +15,22 @@
 class FramebufferObject
 {
  public:
-    enum
+    enum class AttachmentType
     {
-        ColorAttachment = 0x1,
-        DepthAttachment = 0x2
+        None,
```
375gnu, line 20: Better add new bit fields

src/celestia/view.cpp
```diff
@@ -261,16 +260,15 @@ View::drawBorder(Overlay* overlay, int gWidth, int gHeight, const Color &color,
 
 
 void
-View::updateFBO(int gWidth, int gHeight)
+View::updateFBO(int gWidth, int gHeight, FramebufferObject::AttachmentType colorAttachment, FramebufferObject::AttachmentType depthAttachment)
 {
     auto newWidth = static_cast<GLuint>(width * gWidth);
     auto newHeight = static_cast<GLuint>(height * gHeight);
-    if (fbo && fbo.get()->width() == newWidth && fbo.get()->height() == newHeight)
+    if (fbo && fbo->width() == newWidth && fbo->height() == newHeight && fbo->colorAttachmentType() == colorAttachment && fbo->depthAttachmentType() == depthAttachment)
```
375gnu, line 267: no need for additional checks. we can use either textures or renderbuffers but not switch them.

375gnu (changes requested):

levinli303:
> What resolution your ipad has?

2388 x 1668.

src/celengine/framebuffer.h
levinli303: what do you mean here? AttachmentType cannot be a union of multiple.

375gnu:
> > What resolution your ipad has?
> 
> 2388 x 1668.

The it should use 15MB less, not 30. Am I correct?

levinli303:
> The it should use 15MB less, not 30. Am I correct?

my bad. it seems `GL_DEPTH_COMPONENT24` should be used instead of `GL_DEPTH_COMPONENT` in `glRenderbufferStorage`. and in my previous test, `glCheckFramebufferStatus` has been returning `GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT` thus the framebuffer is not really used for viewport effect.

After correcting this, the memory usage and render time of the two were very close.

Issue #2114, opened 2024-02-18
setlocale is not thread safe, causing crash due to race


levinli303:
setlocale is not thread safe, some platforms will call methods related to locale on their own thread, leading a race condition.

```
Thread 6 Crashed:
0   libsystem_c.dylib                    0x00007ff81a5d352f localeconv_l + 323
1   libsystem_c.dylib                    0x00007ff81a5cf0cb __vfprintf + 112
2   libsystem_c.dylib                    0x00007ff81a5fcff8 _vsnprintf + 256
3   libsystem_c.dylib                    0x00007ff81a5daf63 vsnprintf_l + 40
4   libsystem_c.dylib                    0x00007ff81a5db019 snprintf + 155
5   SkyLight                             0x00007ff81feda931 (anonymous namespace)::EventLog::getLogString() const + 276
6   SkyLight                             0x00007ff81fedbd10 (anonymous namespace)::MouseEventLog::getLogString() const + 79
7   SkyLight                             0x00007ff81feda02d CGSEventLogEvent + 360
8   SkyLight                             0x00007ff820103d61 SLSGetNextEventRecordInternal + 331
9   SkyLight                             0x00007ff81feca6db SLEventCreateNextEvent + 8
10  HIToolbox                            0x00007ff825105922 PullEventsFromWindowServerOnConnection(unsigned int, unsigned char, __CFMachPortBoost*) + 44
11  HIToolbox                            0x00007ff8251058c2 MessageHandler(__CFMachPort*, void*, long, void*) + 47
12  CoreFoundation                       0x00007ff81a83887c 0x7ff81a78e000 + 698492
13  CoreFoundation                       0x00007ff81a80a527 0x7ff81a78e000 + 509223
14  CoreFoundation                       0x00007ff81a80a45e 0x7ff81a78e000 + 509022
15  CoreFoundation                       0x00007ff81a8090d6 0x7ff81a78e000 + 504022
16  CoreFoundation                       0x00007ff81a808082 0x7ff81a78e000 + 499842
17  AppKit                               0x00007ff81df75aac _NSEventThread + 121
18  libsystem_pthread.dylib              0x00007ff81a72e202 _pthread_start + 98
19  libsystem_pthread.dylib              0x00007ff81a729bab thread_start + 14

Thread 7:
0   libsystem_kernel.dylib               0x00007ff81a6f01e2 __ulock_wait + 10
[...]

PR #2115, opened 2024-02-18
Remove setlocale for fmt calls

levinli303:


ajtribick (approved):

src/celestia/celestiacore.cpp
```diff
@@ -1206,26 +1209,27 @@ void CelestiaCore::charEntered(const char *c_p, int modifiers)
             Vector3d v = sim->getSelection().getPosition(sim->getTime()).offsetFromKm(sim->getObserver().getPosition());
             int hours, mins;
             float secs;
-            string buf;
+            std::string buf;
+            auto loc = std::locale("");
```
375gnu, line 1213: please make this a CelestiaCore member so it's initialized only once

375gnu (changes requested):

375gnu (approved):

375gnu:
i think we should rework how we handle locales, because we need LC_NUMERIC=C only for reading text catalogs and to work with URLs.

levinli303:
link to #2114

PR #2133, opened 2024-02-28
Read config from writable data path

ColinKinloch:
This MR adds `${XDG_DATA_HOME}/Celestia/celestia.cfg` to the list of locations in which celestia looks for configs.
This option is preferable to `~/.celestia.cfg` for flapak as that would require the whole home directory to be visible inside the sandbox.

`XDG_DATA_HOME` is generally `~/.local/share` however in a flatpak it's `~/.var/app/${APP_ID}/data/`.

### DIFF
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -2304,6 +2304,10 @@ bool CelestiaCore::initSimulation(const fs::path& configFileName,
         localConfigFile = PathExp("~/.celestia-1.7.cfg");
         if (!localConfigFile.empty())
             hasConfig |= ReadCelestiaConfig(localConfigFile, *config);
+
+        localConfigFile = PathExp("~/.celestia/celestia.cfg");
+        if (!localConfigFile.empty())
+            hasConfig |= ReadCelestiaConfig(localConfigFile, *config);
     }
 
     if (!hasConfig)


375gnu:
We already have 3 locations to read. Can it be hidden behind `#ifdef` and a CMake option?

ColinKinloch:
If you're worried about the number of locations an alternative would be replacing the `~/.celestia.1.7.cfg`, then for the stable release of 1.7 using the home directory dotfiles could be deprecated. I assume that most people using that location are technical enough to handle it.

My thinking was to mirror the layout of the usually non user editable `/usr/share/celestia` directory, in the user editable `XDG_HOME_DATA`. With the cfg alongside an `extras` directory etc..

I think adding ifdef's and CMake options is the most complex option.

375gnu:
> using the home directory dotfiles could be deprecated

Please don't force FreeDesktop's GULAG rules.

ajtribick:
What is your objection to the XDG specification? For Linux this makes increasing sense.

375gnu:
Because all it only complicates devs' and users' lives and make a normal system look like windows(c)(r)(tm) with its 'AppData' directories. But there is no real problems that XDG solves. We have dot files and dot dirs which are perfectly been serving our needs for decades without complicated directory hierarchies.

If some systems or package managers want to apply different rules or limitation then it's a package maintainer job to bring the software into compliance, e.g. like we do it [here](https://github.com/CelestiaProject/Celestia/blob/master/debian/rules#L31) or Paolo apply some patches for SUSE and AppImages. So if applcation in flatpack should read from a strange location, then it's flatpack's maintainer responsibility to make a patch.

I use applications which obey XDG and it's always a pain to edit or remove their configs or data files, because they are hidden somewhere in `.local`. While with normal applications it's just `vim .foo<TAB>`

ajtribick:
So having to type an extra directory separator once in awhile is equivalent to being incarcerated in Siberia. Got it.

I strongly disagree with your position here - it at least provides a reasonably well-understood standard to abstract over the differences between various Linux distributions and related environments like Flatpak, without having to force everyone to maintain a bunch of patches, ifdefs and the like. And lazy Windows bashing is lazy - just because Windows has a certain way of doing things does not make that way of doing things automatically bad.

375gnu:
`.celestia` is fine for me

375gnu (approved):

PR #2140, opened 2024-03-02
Move some optional features out of Body class

ajtribick:
Reduce size of the Body class by moving some features that are only used on a limited subset of objects out into a separate features manager.

- Atmospheres
- Ring systems
- Alternate surfaces
- Locations
- Reference marks

Will look at some of the informational fields (the various albedos, temperature discrepancies, etc.) in the future, plus rewrite the name handling (which will also likely move the relevant fields out of the object)


375gnu:
`infoURL` is also a good candidate. Initially it was `std::string*` then I made it just `std::string`. But I would like to replace it with something like `const char*` or better `std::string_view` pointing to a common info URL storage (smth like name database but for info URLs).

ajtribick:
Yep, makes sense.
I also think we can reduce the space occupied by the various `bool` fields: `secondaryIlluminator` and `visibleAsPoint` are derivable from the classification, `primaryNameLocalized` could probably be replaced by `!localizedName.empty()`, leaving `visible` and `clickable` that could be handled as bitfields.

375gnu:
Could you provide some numbers? CPU and memory usage comparing to master?

ajtribick:
Approx 3 MiB memory use reduction with the base data set. Don't see any major differences in CPU usage.

375gnu (approved):

PR #2143, opened 2024-03-03
Reuse ring render data for all ring systems

### LINKED ISSUE #2142: Wasted data for ring systems
We create buffers/vertex objects for every new object rendered that has a ring system, which are never freed. This seems rather wasteful and could relatively easily be optimized. We actually only have 4 distinct resolutions, so we can define the vertices once per resolution for all ring systems, and pass the inner and outer radii as uniforms to the shader.

ajtribick:
Resolves #2142

We only need 4 sets of buffers/vertex objects to cover all resolutions, just reuse them (and use uniforms to handle different inner/outer radii) instead of creating the same data over and over for all ring systems.


375gnu (approved):
actual memory savings are not so large (unless you travel across many planets with rings) but this approach is much more correct

Issue #2049, opened 2023-12-31
Flatpak/Flathub build


faveoled:
Hello,

Could you publish your application to [Flathub](https://flathub.org/)? It's based on a cross-distro [Flatpak technology](https://docs.flatpak.org/en/latest/introduction.html).

Or: would you agree if someone else writes the necessary build files for you and sends them to the Flathub maintainers?

375gnu:
@faveoled please check https://github.com/CelestiaProject/Celestia/discussions/1988, Flatpaks are provided by Fedora community.

faveoled:
@375gnu Thanks, didn't know about that. But what about Flathub submission?

375gnu:
We don't have human resources to maintain additional packaging. If want to volunteer we'll be grateful.

ColinKinloch:
I've created a manifest for Celestia and a PR to get it on flathub: https://github.com/flathub/flathub/pull/4955
@375gnu is this okay with you?
Currently it uses cmake which looks like isn't in a stable release yet.

375gnu:
I'm fine with this. @ajtribick @levinli303 any objections from your side?

levinli303:
looks good to me.

RacerBG:
I don't want to open a new issue about this so I will write it here:

After trying to test the flatpak build (GTK, Qt and SDL) I was always getting a CTD, Celestia didn't even started. The problem was related to my locale settings:
`Detected locale "C" with character encoding "ANSI_X3.4-1968", which is not UTF-8. Qt depends on a UTF-8 locale, and has switched to "C.UTF-8" instead. If this causes problems, reconfigure your locale. See the locale(1) manual for more information.`

The output of localectl is:
`System Locale: LANG=en_US.UTF-8
LANGUAGE=en_US:en
VC Keymap: (unset)
X11 Layout: us,bg
X11 Model: pc105
X11 Variant: ,phonetic
X11 Options: grp:alt_shift_toggle,grp_led:scroll`

And locale shows:
`LANG=en_US.UTF-8
LANGUAGE=en_US
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=bg_BG.UTF-8
LC_TIME=bg_BG.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=bg_BG.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
[...]

PR #2144, opened 2024-03-05
Avoid throwing on invalid locale

ColinKinloch:
This solves the issue mentioned here: https://github.com/CelestiaProject/Celestia/issues/2049#issuecomment-1963008789 where Celestia crashes if the specified locale does not exist.
I can trigger it by running `LC_ALL=C.ANSI_X3.4-1968 celestia-qt6`

This MR solves the issue as `std::locale()` does not throw.
~However it changes the behavior to return either `std::locale::classic()` or the last value entered into `std::locale::global()`.~

The original behavior has been restored by calling `std::locale::global(std::locale(""))` in the case that the user selected locale is found by `setlocale`.
~I'm not sure how to do this as the project uses `-fno-exceptions`.~

https://en.cppreference.com/w/cpp/locale/locale/locale

### DIFF
--- a/src/celutil/localeutil.cpp
+++ b/src/celutil/localeutil.cpp
@@ -1,0 +1,32 @@
+// localeutil.cpp
+//
+// Copyright (C) 2024-present, Celestia Development Team.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#include <clocale>
+#include <locale>
+#include <fmt/format.h>
+
+namespace celestia::util
+{
+
+void InitLocale()
+{
+    /* try to set locale, fallback to "classic" */
+    if (setlocale(LC_ALL, ""))
+    {
+        std::locale::global(std::locale(""));
+    }
+    else
+    {
+        fmt::print("Could not find locale, falling back to classic.\n");
+    }
+    /* Force number displays into C locale. */
+    setlocale(LC_NUMERIC, "C");
+}
+
+}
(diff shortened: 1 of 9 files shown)


ColinKinloch:
That's correct we could keep the same crashing behaviour by adding locale global.
Is there a way to ignore exceptions?

ajtribick:
I think we can check whether `setlocale` returns `nullptr` - we should probably do that at application start to avoid changing the global locale which results in thread safety issues. I'm not entirely sure that we should support such broken environments where the user's preferred locale does not exist, but maybe we can provide a cleaner error message on exit.

The usages of the `format` overloads that take a `locale` should be replaced by the `NumberFormatter` class - we expect these strings to be UTF-8 but on Windows it will return a string in the ANSI code page.

ColinKinloch:
Of course libc, always simplifies things.

>  I'm not entirely sure that we should support such broken environments

I don't disagree. On the other hand avoiding exceptions doesn't seem like too bad an idea.

Would I have to set the global locale in each front end, or is  there some appropriate common location?

src/celutil/localeutil.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <locale>
+#include <fmt/format.h>
+
+namespace celestia::util
+{
+
+void InitLocale()
+{
+    /* try to set locale, fallback to "classic" */
+    if (setlocale(LC_ALL, "")) {
```
375gnu, line 20:
```suggestion
    if (setlocale(LC_ALL, ""))
    {
```

src/celutil/localeutil.cpp
```diff
@@ -0,0 +1,29 @@
...
+        std::locale::global(std::locale(""));
+    } else {
```
375gnu, line 22:
```suggestion
    }
    else
    {
```

375gnu (changes requested):
Could you also add a comment to the commit on `std::locale()` behavior?

375gnu (approved):

ajtribick (approved):

PR #2145, opened 2024-03-06
Convert Body classification to enum class

ajtribick:
Fix bug in cel script orbitflags command - parse correct map.


ajtribick:
Going to re-raise this as branch in Celestia repository to re-trigger the binary caching

PR #2146, opened 2024-03-06
Convert Body classification to enum class

ajtribick:
Fix bug in cel script orbitflags command - parse correct map.


src/celengine/body.h
```diff
@@ -96,76 +97,79 @@ class RingSystem
...
+// * Diffuse is used for gas clouds, dust plumes, and the like. They are
+// visible, but other properties are false by default. It is expected
+// that an observer will move through a diffuse object, so there's no
+// need for any sort of collision detection to be applied.
+//
+// * Stellar is a pseudo-class used only for orbit rendering.
+//
+// * Barycenter and SmallBody are not used currently. Invisible is used
+// instead of barycenter.
+enum class BodyClassification : std::uint16_t
```
375gnu, line 145: why limiting to `std::uint16_t`?

ajtribick: ah right, had done this on top of the other PR with optional features but split it out... made sense in the context of the layout there but probably going to do some changes there as well

375gnu (approved):

PR #2155, opened 2024-03-18
Pass locale around instead of setting global locale

levinli303:
I've observed weird behaviors when `std::locale::global(std::locale(""))` is called, on Windows + ANGLE, stars are not rendered.


src/celastro/date.cpp
```diff
@@ -213,13 +210,14 @@ Date::toString(Format format) const
...
+        return fmt::format(loc, "{:%c}"sv, cal_time);
     case TZName:
-        return fmt::format(std::locale(), "{:%Y %b %d %H:%M:%S %Z}"sv, cal_time);
+        return fmt::format(loc, "{:%Y %b %d %H:%M:%S %Z}"sv, cal_time);
     default:
-        return fmt::format(std::locale(), "{:%Y %b %d %H:%M:%S %z}"sv, cal_time);
+        return fmt::format(loc, "{:%Y %b %d %H:%M:%S %z}"sv, cal_time);
     }
 #else
+    static auto monthAbbreviations = MonthAbbreviations(loc);
```
375gnu, line 220: this allocates heap memory, so it should not be used this way. add a static pointer and use `new` on the 1st call

375gnu (changes requested):

src/celastro/date.cpp
```diff
@@ -213,26 +211,27 @@ Date::toString(Format format) const
...
+        return fmt::format(loc, "{:%c}"sv, cal_time);
     case TZName:
-        return fmt::format(std::locale(), "{:%Y %b %d %H:%M:%S %Z}"sv, cal_time);
+        return fmt::format(loc, "{:%Y %b %d %H:%M:%S %Z}"sv, cal_time);
     default:
-        return fmt::format(std::locale(), "{:%Y %b %d %H:%M:%S %z}"sv, cal_time);
+        return fmt::format(loc, "{:%Y %b %d %H:%M:%S %z}"sv, cal_time);
     }
 #else
+    static MonthAbbreviations* const monthAbbreviations = std::make_unique<MonthAbbreviations>(loc).release(); //NOSONAR
```
375gnu, line 221:
```suggestion
    static const MonthAbbreviations* monthAbbreviations = nullptr;
    if (monthAbbreviations == nullptr)
       monthAbbreviations  = std::make_unique<MonthAbbreviations>(loc).release(); //NOSONAR
```

375gnu (changes requested):

375gnu (approved):

PR #2163, opened 2024-03-24
Use Boost intrusive_ptr and static_vector

ajtribick:
Replaces our custom ArrayVector and IntrusivePtr implementations with the equivalents from Boost: these are better known and have the advantage of being tested outside Celestia's codebase.


375gnu:
I always hated this monstrous library due to its size. Will it be linked into a resulting binary?

ajtribick:
Right, so `Boost::boost` is the header-only component (later versions of CMake provide the alias `Boost::headers` which makes this clearer, but we can't use that with our minimum version).

Building on Windows, I get identical filesizes of celestia.dll and celestia-win.exe before and after the change, and (as expected for a header-only library) no additional boost dll in the output folder.

375gnu (approved):

PR #2165, opened 2024-03-27
Updates for intrusive_ptr refcounting

ajtribick:
- Initialize the refcounts with 1
- Use memory orderings that match the implementation of shared_ptr refcounting in libc++/Windows STL

Latter also matches the example in [Herb Sutter's `atomic<>` weapons talk](https://www.youtube.com/watch?v=KeLBd2EJLOU) (discussion starts at 1:19:51), should be a little better on ARM.


375gnu (approved):
I have no idea what';s going here but I hope you do.

PR #2168, opened 2024-04-04
Remove expression trees from shadermanager

ajtribick:
- Improved searchability of GLSL code in C++ source
- Simplifies implementation, removes need for reference counting


375gnu (approved):
i believe this had a purpose, like shader validation during Celestia compilation and allowing to change shader language, but as it's unfinished, and no one wants to finalize this work, let's use inline texts.

PR #2170, opened 2024-04-06
Split out SkyGridRenderer from SkyGrid

ajtribick:
- Use unique_ptr to hold LineRenderer rather than globals to simplify Renderer destruction
- Move types into celestia::engine and celestia::render namespaces


src/celrender/skygridrenderer.cpp
```diff
@@ -0,0 +1,654 @@
...
+{
+
+// The maximum number of parallels or meridians that will be visible
+constexpr double MAX_VISIBLE_ARCS = 10.0;
+
+// Number of line segments used to approximate one arc of the celestial sphere
+constexpr int ARC_SUBDIVISIONS = 100;
+
+// Size of the cross indicating the north and south poles
+const double POLAR_CROSS_SIZE = 0.01;
```
375gnu, line 48:
```suggestion
constexpr double POLAR_CROSS_SIZE = 0.01;
```

375gnu (approved):

PR #2173, opened 2024-04-12
Fix build issues arising from syntax error in acinclude.m4

afontenot:
The acinclude.m4 file places the closing `fi` in the wrong location, resulting in builds on some systems failing, e.g. Arch Linux: https://aur.archlinux.org/packages/celestia

The `fi` belongs where it is placed by this commit because there is no `if` inside the `AC_CACHE_VAL`. Rather, the statement is intended to close the conditional `if test -n "$LIBJPEG"`.


375gnu (approved):

375gnu:
M4 is nightmare ) thanks

PR #2187, opened 2024-05-22
Split Frustum and InfiniteFrustum

ajtribick:
Shrinks the object sizes and makes it clearer which type of object is being used.

- Use enum classes for FrustumPlane and FrustumAspect
- Update projection modes to pass matrices by reference
- Use std::array for viewport in projection routines
- Remove unused math::ProjectPerspective overload


ajtribick:
Duplication: WONTFIX - appears to be complaining because classes derived from the same parent class override the same virtual methods.

375gnu (approved):

PR #2189, opened 2024-05-26
Remove partially-implemented Gliese cross-index code

ajtribick:
No objections to re-adding Gliese support at a future time, but currently this code doesn't have any useful effects.


SevenSpheres:
I would suggest removing the `GlieseCrossIndex` line from celestia.cfg (which for some reason has been renamed to celestia.cfg.in?)

ajtribick:
Updated celestia.cfg.in

375gnu (approved):

PR #2192, opened 2024-05-29
Remove constexpr from absToAppMag/appToAbsMag

ajtribick:
Unfortunately std::log10 is not constexpr until C++26.


375gnu (approved):
you can make it constexpr for g++ only

ajtribick:
@375gnu - can you exclude the checks directory from Codacy? It's not particularly useful there and I don't think I have permissions to do this.

375gnu (approved):

375gnu:
> can you exclude the checks directory from Codacy? It's not particularly useful there and I don't think I have permissions to do this.

done

PR #2197, opened 2024-06-03
Reorganize star database code

ajtribick:
- Move StarDBBuilder to separate files
- Move cross-index catalogs into StarNameDatabase
- Use indices instead of pointers in catalogNumberIndex


ajtribick:
Fixing the Sonar critical issues would require a fairly major refactoring of the .stc parsing code, I would rather do that in a future PR.

375gnu (approved):

PR #2200, opened 2024-06-07
Reduce AtmosphereExtinctionThreshold value

pedro-fixingstuff:
Testing by **MinersHavenM43** on the Discord server confirms that this change produces the expected effect of making the cut-off at the top of atmospheres almost invisible, while causing no obvious issues.

Before:
image

After (screenshot by him):
image

### DIFF
--- a/src/celengine/atmosphere.h
+++ b/src/celengine/atmosphere.h
@@ -65,4 +65,4 @@ class Atmosphere
 // to render. The radius of the sphere is the height at which the
 // density of the atmosphere falls to the extinction threshold, i.e.
 // -H * ln(extinctionThreshold)
-constexpr inline float AtmosphereExtinctionThreshold = 0.05f;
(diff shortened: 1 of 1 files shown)


375gnu:
What about other bodies?

AstroChara:
Hm, considering that we will probably need to update the atmosphere rendering to a newer, more accurate model, perhaps this could be postponed or merged into that, whenever it happens?

AstroChara:
After considering the options, @pedro-fixingstuff and I decided to close this pull request for the following reasons:
1. Although this change fixes Titan's atmosphere, appears to have an unintended consequence on Earth, as shown here: https://discord.com/channels/558624707147071489/691413425494032474/1249504464185069662
2. The current atmosphere rendering is old and inaccurate, and should be updated to a newer rendering system in any case; issues related to this may be fixed as part of this update.

PR #2216, opened 2024-07-10
Use smart pointers in octree classes

ajtribick:
Use smart pointers in octree classes, modify the builder to use moves rather than copies to enable the DSOs to be held in `std::unique_ptr`.

The current data structure requires recursion (including during destruction) but I will fix that in a future pull request.

### DIFF
--- a/src/celengine/dsodb.cpp
+++ b/src/celengine/dsodb.cpp
@@ -12,6 +12,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <numeric>
 #include <utility>
 
 #include <celcompat/numbers.h>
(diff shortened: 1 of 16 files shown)


src/celengine/dsodb.cpp
```diff
@@ -238,46 +235,34 @@ DSODatabase::load(std::istream& in, const fs::path& resourcePath)
...
         if (obj != nullptr && obj->load(objParams, resourcePath))
         {
-            UserCategory::loadCategories(obj, *objParams, DataDisposition::Add, resourcePath.string());
+            UserCategory::loadCategories(obj.get(), *objParams, DataDisposition::Add, resourcePath.string());
 
             // Ensure that the DSO array is large enough
-            if (nDSOs == capacity)
+            if (DSOs.size() == DSOs.capacity())
             {
                 // Grow the array by 5%--this may be too little, but the
```
375gnu, line 255: Please update the comment

src/celengine/dsorenderer.cpp
```diff
@@ -55,12 +55,11 @@ brightness(float avgAbsMag, float absMag, float appMag, float brightnessCorr, fl
...
 } // anonymous namespace
 
-DSORenderer::DSORenderer() :
-    ObjectRenderer<DeepSkyObject*, double>(DSO_OCTREE_ROOT_SIZE)
+DSORenderer::DSORenderer() : ObjectRenderer(DSO_OCTREE_ROOT_SIZE)
 {
 }
 
-void DSORenderer::process(DeepSkyObject* const &dso,
+void DSORenderer::process(const std::unique_ptr<DeepSkyObject>& dso, //NOSONAR
```
375gnu, line 62: why ref to uniqptr instead of raw ptr?

src/celengine/universe.cpp
```diff
@@ -424,7 +424,7 @@ class DSOPicker : public DSOHandler
               float angle);
     ~DSOPicker() = default;
 
-    void process(DeepSkyObject* const &, double, float) override;
+    void process(const std::unique_ptr<DeepSkyObject>&, double, float) override; //NOSONAR
```
375gnu, line 427: again, why not raw pointer?

375gnu (changes requested):

src/celengine/dsorenderer.cpp
ajtribick: Because that's how the base template class `OctreeProcessor<OBJ, PREC>` defines it. In the case of the DSO octree, `OBJ` is `std::unique_ptr<DeepSkyObject>` (previously `DeepSkyObject*`), and defines the pure virtual method `process` as taking a parameter of type `const OBJ&`, which ends up being `const std::unique_ptr<DeepSkyObject>&` (previously `DeepSkyObject* const&`).

src/celengine/dsodb.cpp
ajtribick: On second thoughts, this growth strategy was designed for the case where a whole bunch of DSOs were loaded from a binary file, but that never happened. Might as well use the default vector growth strategy for now, especially since this vector only holds pointer-sized objects anyway.

375gnu (approved):

src/celengine/universe.cpp
ajtribick: as above

PR #2217, opened 2024-07-14
Split DSODatabaseBuilder and DynamicOctree into separate files

ajtribick:
Reduces the amount of code pulled in by including the octree files. This will make later changes to the octree structure easier to contain.


ajtribick:
Sonar rule S4963 is marked as deprecated. I'm not sure how to disable this in the profile.

375gnu (approved):

PR #2218, opened 2024-07-14
Build updates

ajtribick:
Various build updates:

- Default all front-ends to `OFF`
- Use vcpkg in manifest mode to automatically install dependencies
- Switch to using triplets x64-window-release (community triplet supplied with vcpkg) and x86-windows-release (created in this repository) to build the packages, rather than hacking the existing x64-windows/x86-windows triplets.
- Provide an option to use a local installation of Qt rather than building it with vcpkg
- Update Windows build instructions


375gnu (approved):
though i'm not a big fan of making vcpkg a hard dependency, it's the only currently supported way, so let's roll

PR #2221, opened 2024-07-25
Revert "Updates to build and instructions"

ajtribick:
This reverts commit 3bb3d12640b144157a6f4c40ac7e5abc4c71c9bc.

On further testing, vcpkg manifest mode seems to suffer from excessive full dependency rebuilds on things like switching branches. Switching back to classic mode.


Askaniy:
I think `Default all front-ends to OFF` was a good idea, maybe leave that change in?

375gnu (approved):

ajtribick:
@Askaniy - agreed, but I prefer to do this as a clean revert and re-do the instructions update later.

PR #2222, opened 2024-07-25
Linearize octree node storage

ajtribick:
- Separate octree types and octree node types
- Store nodes in vectors (BlockArray for the DynamicOctree)
- For static octree, order the nodes in depth-first order, together with the node to jump to when skipping the subtree
- Avoids need to perform recursive calls when processing the octree (including the destructor)

### DIFF
--- a/src/celastro/astro.h
+++ b/src/celastro/astro.h
@@ -76,10 +76,18 @@ float appMagToLum(float mag, float lyrs);
 
 template<class T>
 CELESTIA_CMATH_CONSTEXPR T
+distanceModulus(T lyrs)
+{
+    using std::log10;
+    return T(5) * log10(lyrs / LY_PER_PARSEC<T>) - T(5);
+}
+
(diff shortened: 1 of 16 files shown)


ajtribick:
The switch from creating all 8 child nodes at the same time to only creating when an object needs to be added reduces the number of nodes in the final octree:

Stars: 112745 --> 111176
DSOs: 6337 -> 4095

We also track the brightness via the objects contained within each node (and descendant nodes) rather than using the computed threshold magnitudes, this should allow for earlier culling of faint nodes when processing visible objects.

375gnu:
Any difference in loading & finding stars?

ajtribick:
Didn't see any major differences in render time processing the star octree, seems slightly faster (by 0.1-0.5 ms, so nothing really noticeable) to build the star octree.

375gnu (approved):

PR #2234, opened 2024-10-08
Add Selection info to getCompletion

### LINKED ISSUE #2231: Keyboard Location Entry - Multiple Items
This issue appears to be in the common Celestia code, thus affecting Qt, GTK, SDL and Windows interfaces.

When on the main Celestia viewing screen, pressing ENTER opens a line at the bottom of the screen allowing the keyboard entry of a celestial object and displaying a listing of objects matching the partial keyboard entry. The user can scroll through the listing, selecting one and allowing focus of the display to move to that selected object.

If there are two or more objects with the same name, the listing correctly displays the number of like-named objects in the list.  However, the link for all the listings is only to the first object.

This can be demonstrated by Juno - an asteroid in the main databases; and Juno - the spacecraft orbiting Jupiter (available at Celestia.mobi). The keyboard entry listing displays two entries labeled Juno. However, selecting either entry will only go to the asteroid Juno.  It is not possible to get to the spacecraft Juno from this listing.

While this can be overcome by using a different object browsing window (Solar System Browser) on some of the interfaces, no such method exists for the SDL interface.  It can also be overcome by the user editing one of the *.ssc files to change the name of the object.  

With the wide variety of addon authors and download locations, it is not unexpected to find duplication in object names.  Users editing *.ssc files is not a good, long-term solution.

levinli303:
This allows distinguishing between completions with same names. Closes #2231

### DIFF
--- a/src/celengine/completion.h
+++ b/src/celengine/completion.h
@@ -1,0 +1,36 @@
+// completion.h
+//
+// Copyright (C) 2024-present, the Celestia Development Team
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+
+#pragma once
+
+#include <functional>
+#include <string>
+#include <variant>
+#include <celengine/selection.h>
+
+namespace celestia::engine
+{
+
+class Completion
+{
+public:
+    Completion(const std::string& name, const std::variant<Selection, std::function<Selection()>>& selection) :
+        name(name),
+        selection(selection)
+    {}
+
+    std::string getName() const;
+    Selection getSelection() const;
+
+private:
+    std::string name;
+    std::variant<Selection, std::function<Selection()>> selection;
+};
+
+}
--- a/src/celengine/CMakeLists.txt
+++ b/src/celengine/CMakeLists.txt
@@ -11,6 +11,8 @@ set(CELENGINE_SOURCES
   boundaries.h
   category.cpp
   category.h
+  completion.cpp
+  completion.h
   console.cpp
   console.h
   constellation.cpp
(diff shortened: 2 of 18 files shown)


src/celengine/completion.h
```diff
@@ -0,0 +1,36 @@
...
+#include <functional>
+#include <string>
+#include <celengine/selection.h>
+
+namespace celestia::engine
+{
+
+struct Completion
+{
+    Completion(const std::string& name, const Selection& selection, std::function<Selection()> resolver = nullptr) :
```
ajtribick, line 21: If I'm reading this correctly, this is an either/or situation - probably should be two separate constructors, one taking a Selection, one taking a resolver.

src/celengine/completion.h
```diff
@@ -0,0 +1,36 @@
...
+        selection(selection),
+        resolver(resolver)
+    {}
+
+    const std::string& getName() const;
+    const Selection& getSelection();
+
+private:
+    std::string name;
+    Selection selection;
```
ajtribick, line 32: Possibly could be `std::variant<Selection, std::function<Selection()>>`

src/celengine/completion.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include <functional>
+#include <string>
+#include <celengine/selection.h>
+
+namespace celestia::engine
+{
+
+struct Completion
```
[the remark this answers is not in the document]
levinli303: changed to class

src/celengine/completion.h
levinli303: switched to use std::variant instead and use it in constructor as well. I originally wanted to avoid `getSelection` calling resolver repeatedly so storing them separately. but it seems in our code it will only be called once.

375gnu (approved):

PR #2245, opened 2024-10-24
Update vcpkg to latest (1.6.x)

ajtribick:
Update in case we want to build another 1.6.x release at some point


celestia.vcxproj
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildThisFileDirectory)\windows\vcpkg\scripts\buildsystems\msbuild\vcpkg.props" />
```
levinli303, line 3: is it possible to use `vcpkg integrate install` to configure vcpkg integration so we don't need to add it as a submodule?

.gitmodules
levinli303, line 1: empty file can be removed.

celestia.vcxproj
```diff
@@ -95,7 +95,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
-      <AdditionalIncludeDirectories>.\src;C:\tools\vcpkg\installed\x86-windows\include\cspice;C:\tools\vcpkg\installed\x86-windows\include\luajit;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>.\src;$(VCPKG_INSTALLATION_ROOT)\installed\x64-windows\include\cspice;$(VCPKG_INSTALLATION_ROOT)\installed\x64-windows\include\luajit;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
levinli303, line 98: I think we should use macros from https://learn.microsoft.com/en-us/vcpkg/users/buildsystems/msbuild-integration instead of VCPKG_INSTALLATION_ROOT which is only specified on GitHub runners.

ajtribick: I despise msbuild

levinli303 (approved):

PR #2246, opened 2024-10-25
Various BMP loader fixes

ajtribick:
- Fix mirrored vertical orientation
- Detect more error conditions/malformed files
- Execute fewer read operations on the stream
- Fix issue calculating the row stride (should round up to multiple of 4, not 2)
- Fix palette offset in the presence of extended info structures
- Use magenta for out-of-range palette references
- Support 1, 2, 4 bits per pixel

The vertical orientation issue seems to be a long-standing bug, I guess no-one's noticed because no-one uses .bmp images.

### DIFF
--- a/src/celimage/bmp.cpp
+++ b/src/celimage/bmp.cpp
@@ -8,33 +8,49 @@
 // as published by the Free Software Foundation; either version 2
 // of the License, or (at your option) any later version.
 
+#include <algorithm>
+#include <array>
+#include <cassert>
+#include <cstddef>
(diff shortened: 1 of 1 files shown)


ajtribick:
The upside-down image loading appears to have been an issue for a very long time, e.g. it is present in 1.6.x. Probably no-one is using BMP images so it doesn't matter much, but technically doing this could break stuff. (Several of the fixes in here should also probably be backported to 1.6.x)

375gnu (approved):

PR #2289, opened 2025-02-02
[ci] use github arm64 runner

375gnu:


.github/workflows/ci.yml
```diff
@@ -189,7 +189,7 @@ jobs:
           wayland: ON
           icu: ON
           qt6: OFF
-        - os: ubuntu-24.04
+        - os: ubuntu-24.04-arm
```
ajtribick, line 192: Looks like this replaces the ubuntu-24.04 build with the ARM one, I think we want to do both an x64 build and an arm64 build on 24.04?

375gnu: Personally I don't see any need in this as we have ubuntu 20.04 with amd64. It could be possible that something works with older ubuntu/g++ and amd64 and with newer ubuntu/g++ and arm64 but not with newer ubuntu/g++ and amd64.But what the actual probability of this? I believe it's much much lower than 1%.

ajtribick (approved):

Issue #2292, opened 2025-02-14
DDS Textures Not Displaying


dave-kaye:
Issue: DDS image formats are not displaying as textures for spacecraft models.

![Image](https://github.com/user-attachments/assets/cccb4d1d-0fba-415c-9ff6-46d45886011c)

ISS model with dds textures

![Image](https://github.com/user-attachments/assets/f9517118-fb1b-40f9-be6a-36b5e84a59a3)

ISS model after recent code 58d7712

Affects: Recent versions of Celestia.  Affects both Windows operating system compiles from Github (via downloadable artifacts) and Linux openSUSE 15.2 compiles. Affects Qt5, Gtk3, native WIN and SDL interfaces.  Therefore it resides in Celestia core code.

History: Is not present in github 97fe2c9 (Nov 3, 2024). Begins in github 62716d8 (Nov 7, 2024).  Is still present in current code, as shown in native Windows interface - see picture above.

Test Model: High quality ISS model originating with linuxman.  An early version which is suitable for testing can be found the website [http://celestiasws.free.fr/ISS/ISS.html](http://celestiasws.free.fr/ISS/ISS.html) utilizing the download link [http://celestiasws.free.fr/iss-construction.zip]( http://celestiasws.free.fr/iss-construction.zip).  The ISS is listed in menus (under Earth) as Zarya, the name of the first module of the ISS.

Additional information: The file which has the bug appears to be celimage/dds.cpp.  The file formats of *.png, *.bmp and *.jpg are working properly.  The file format *.avif is unknown. Attempts to utilize the current Win artifact does not show a model, but it is unknown if avif is enabled in the artifact.  If the old version of dds.cpp is substituted into the github 62716d8 code it compiles successfully (Linux) and the dds textures are visible as normally seen. Log outputs of the new github 62716d8 version show an error message originating in line 335 of dds.cpp.  MediaInfo program states that the textures in the ISS model are DXT1 style with 1 bit alpha/transparency channel.

PR #2293, opened 2025-02-17
[celimage] DDS Texture Display

dave-kaye:
This code reverts an if statement format in dds.cpp found in git 62716d8 to the format of a prior version.

It fixes issue #2292.


dave-kaye:
The code reverts the if statement to the previous version.  The dds textures now load and display successfully.

The if statement format of github 62716d8 was inserted into the older version of the dds.cpp file.  The file compiled successfully, but the textures will not load or display - identical to the issue with the updated dds.cpp.  It appears the !in.read(...) statement does not return the expected value, and that is why the the original code had a more elaborate if statement.

NOTE: The prior code had a "delete image" statement as part of the (if => true) actions.  The newer code removed this delete statement.  I do not know if it should be present or not.  Please add if needed.

375gnu (approved):

levinli303 (approved):

PR #2300, opened 2025-02-26
C++20 forward compatibility fixes

ajtribick:
Fixes for various issues identified when compiliing in C++20 mode:

* Clang 19 complains about certain uses of incomplete types
  - Major one is the use of `Value` in the templated `for_all` method of `AssociativeArray`. Solved by combining the headers
  - Inner types in `GalaxyRenderer` and `VertexObject` must be complete at the point the constructor of the main type is defined.
* Format strings that are not compile-time evaluated are wrapped in `fmt::runtime` - this requires bumping minimum fmt to 8.1.0
* Fix a couple of malformed format strings
* As part of `AssociativeArray`/`Value` cleanup, move the sources into the celutil folder and use the `celestia::util` namespace
* Also move the parser into celutil, and put the parser and tokenizer into the `celestia::util` namespace

A weird issue is that compiling in C++20 mode causes the star atmospheres to stop working, I am not sure why. C++17 mode still works.


src/celengine/asterism.cpp
```diff
@@ -20,20 +20,22 @@
 #include "star.h"
 #include "stardb.h"
 
+namespace util = celestia::util;
```
375gnu, line 23: why not `using celestia::util::Tokenizer`? then you'll have only 1 line change

src/celengine/category.cpp
```diff
@@ -5,9 +5,10 @@
 #include <initializer_list>
 #include <utility>
 
+#include <celutil/associativearray.h>
 #include <celutil/gettext.h>
-#include "hash.h"
-#include "value.h"
+
+namespace util = celestia::util;
```
375gnu, line 11: as above

375gnu (approved):

src/celengine/category.cpp
ajtribick: Not so useful in this file because of the small number of individual usages, have updated some of the other ones.

PR #2306, opened 2025-03-16
Remove the GTK front-end

ajtribick:
Unmaintained front-end. Alternatives are Qt for the main cross-platform front-end, or SDL for a more lightweight version with an immediate-mode GUI.


375gnu (approved):
let's go

PR #2309, opened 2025-04-01
[Qt] Nebula Type in DSO Browser

dave-kaye:
Git f5512c4 removed the type column for nebula objects from the deep sky/celestial browser. Partially finished code still remains. This existing code is updated/finished and type designations are displayed for nebula.

### DIFF
--- a/src/celestia/qt/qtdeepskybrowser.cpp
+++ b/src/celestia/qt/qtdeepskybrowser.cpp
@@ -166,11 +166,25 @@ DSOPredicate::operator()(const DeepSkyObject* dso0, const DeepSkyObject* dso1) c
     }
 }
 
-void populateDsoVector(std::vector<DeepSkyObject*>& dsos,
-                       const DSODatabase& dsodb,
-                       const DSOFilterPredicate& filter,
-                       const DSOPredicate& comparison,
-                       unsigned int nDSOs)
+// look for empty or all spaces DSO object name
+bool
+isEmptyDSOName(const DSODatabase& dsodb,
+               DeepSkyObject* dso)
+{
+    std::string name = dsodb.getDSOName(dso, true);
+    if (name.empty())
+        return true;
+    if (std::all_of(name.begin(), name.end(), [](char c) {return std::isspace(c);}))
+        return true;
+    return false;
+}
+
+void
+populateDsoVector(std::vector<DeepSkyObject*>& dsos,
+                  const DSODatabase& dsodb,
+                  const DSOFilterPredicate& filter,
+                  const DSOPredicate& comparison,
+                  unsigned int nDSOs)
 {
     const std::uint32_t size = dsodb.size();
     std::uint32_t index = 0;
@@ -187,11 +201,7 @@ void populateDsoVector(std::vector<DeepSkyObject*>& dsos,
         DeepSkyObject* dso = dsodb.getDSO(index);
         if (filter(dso))
         {
-            std::string dson = dsodb.getDSOName(dso, true);
-            if (dson.empty())
-                continue;
-
-            if (std::all_of(dson.begin(), dson.end(), [](char c) {return std::isspace(c);}))
+            if (isEmptyDSOName(dsodb, dso))
                 continue;
 
             if (dsos.size() == nDSOs)
@@ -211,11 +221,7 @@ void populateDsoVector(std::vector<DeepSkyObject*>& dsos,
         if (!filter(dso))
             continue;
 
-        std::string dson = dsodb.getDSOName(dso, true);
-        if (dson.empty())
-            continue;
-
-        if (std::all_of(dson.begin(), dson.end(), [](char c) {return std::isspace(c);}))
+        if (isEmptyDSOName(dsodb, dso))
             continue;
 
         if (comparison(dso, dsos.front()))


dave-kaye:
Attached is a file of sample nebulae to test the code with.  There are examples of nebulae that will demonstrate several of the code modifications.
 
planetary nebula:
eight-burst
ring
cat's eye
eskimo
lemon slice

supernova remnant:
firework
crab

These two sets will allow testing of the type filter box located near the bottom of the celestial browser window.  Success will be if only the type entered into the filter is displayed in the listing and all examples of that type are shown.

The cat's eye nebula has multiple cmod files associated with it in the .dsc file.  These nebula parts have no names.  Success will be if a) the no name sections do not appear in the listing; and b) the cat's eye nebula still shows all of its parts (none of the no name sections are missing) when viewed as an image in the main window.

The firework nebula does not have an absolute magnitude associated with it. Success will be if the App. mag. column is blank, and does not display a default -999 magnitude which is impossible.

All nebula should display a reasonable information screen in the upper left corner of the main object display window when accessed.

Note: The deep sky browser is dynamic. It will change the distance and magnitude values based upon where the observer is located. However, it does need to be reloaded if the observer's location is changed and the user wants to see the new values. Otherwise the initial values based upon an earth observer point are displayed.

[deepsky.zip](https://github.com/user-attachments/files/19553506/deepsky.zip)


----------------

The samples were obtained from the Celestia Mobi website, then updated with newer information from primarily the NASA and Wikipedia websites. The tags modified in the .dsc were the addition of type and absolute magnitude tags.  The "best" distance and magnitude values were selected from multiple conflicting data.  Also the clickable tag was verified for multi-cmod files, so only the primary nebula image is clickable.

The absolute magnitude was calculated from the apparent magnitude with the formula:

magnitude calc

dave-kaye:
If this code is accepted, there are several items which can be updated.

[https://en.wikibooks.org/wiki/Celestia/DSC_File]
The wikibooks entry for Celestia DSC files in the Nebula section should be updated to add the "Type" tag using the nebula types found in the nebula.h file (enum Nebula type approx lines 27-36) and nebula.cpp file (array NebulaTypeNames approx lines 37-44). The NebulaTypeNames array contains the spelling/spacing of the types for the input .dsc file.  The wikibooks entry should also add the use of the AbsMag tag.  SEE the Galaxy entry as an example.  _Note_: the general/common entry and galaxy entry shows AppMag tag as active, however testing a .dsc file by replacing a AbsMag tag with AppMag shows the code is not active/working.  The clickable tag can also be added to the DSC entry.

The models on Celestia mobi and elsewhere can have their .dsc files updated; adding the tags Type and AppMag and Clickable as warranted.  If owner/usage considerations prevent model file data from being updated, the necessary information can be posted as a comment and allow users to make the updates.

Distance and magnitude data varies widely based upon the source.  Newer Gaia DR2 and DR3 data may be of use here.  Also, the maintainers of the Celestia Origins website may have improved data.  There is also a Planetary Nebula Database on the internet, however internet safety software is blocking my access to it.  One thing I have found is many authors declare the magnitude of the nebula to be the magnitude of the central star.  Others will keep them separate.  Seperate is the better option so the magnitude of the nebula is the magnitude of the nebula object.

src/celengine/nebula.cpp
```diff
@@ -22,23 +23,47 @@
...
+                             [&](const NebulaTypeName& n) { return n.name == typeStr; });
+    if (iter != std::end(NebulaTypeNames))
+        type = iter->type;
 }
 
 std::string
 Nebula::getDescription() const
 {
-    return _("Nebula");
+    return fmt::sprintf(_("Nebula: %s"), getType());
```
ajtribick, line 66: Prefer the `fmt::format` overloads rather than `fmt::sprintf`

Also this should use translatable type names, rather than the dsc file keywords.

src/celengine/nebula.h
```diff
@@ -24,6 +24,18 @@ namespace celestia::util
...
+enum class NebulaType
+{
+    Emission           = 0,  // includes Herbig–Haro objects and misc. emission nebula not listed below
+    Reflection         = 1,  // includes misc. reflection nebula not listed below
+    Dark               = 2,
+    Planetary          = 3,
+    SupernovaRemnant   = 4,
+    HII_Region         = 5,
+    Protoplanetary     = 6,
+    NotDefined         = 7
```
ajtribick, line 36: I think it would be better to have the NotDefined case as 0

src/celestia/qt/qtdeepskybrowser.cpp
```diff
@@ -184,8 +184,13 @@ void populateDsoVector(std::vector<DeepSkyObject*>& dsos,
             return;
         }
 
-        if (DeepSkyObject* dso = dsodb.getDSO(index); filter(dso))
+        DeepSkyObject* dso = dsodb.getDSO(index);
+        if (filter(dso))
         {
+            std::string dson = dsodb.getDSOName(dso, true).c_str();
```
ajtribick, line 190: Unnecessary call to `c_str()`

src/celestia/qt/qtdeepskybrowser.cpp
```diff
@@ -203,6 +208,10 @@ void populateDsoVector(std::vector<DeepSkyObject*>& dsos,
         if (!filter(dso))
             continue;
 
+        std::string dson = dsodb.getDSOName(dso, true).c_str();
```
ajtribick, line 211: Unnecessary call to `c_str()`

src/celengine/nebula.cpp
375gnu, line 66:
```suggestion
    return fmt::format(_("Nebula: {}"), getType());
```
Don't use C-style function

src/celengine/nebula.cpp
```diff
@@ -14,6 +14,7 @@
 #include <celutil/fsutils.h>
 #include <celutil/logger.h>
 #include <celutil/gettext.h>
+#include <fmt/printf.h>
```
375gnu, line 17:
```suggestion
#include <fmt/format.h>
```

src/celengine/nebula.cpp
```diff
@@ -62,6 +87,12 @@ Nebula::getObjType() const
 bool
 Nebula::load(const util::AssociativeArray* params, const fs::path& resPath, std::string_view name)
 {
+    
+    if (const auto* typeName = params->getString("Type"); typeName == nullptr)
+        setType({});
```
ajtribick, line 92: I think we can just leave the type alone if it is not specified, since the field initializer already defaults it to `NotDefined`

src/celengine/nebula.cpp
```diff
@@ -22,23 +23,47 @@
...
 const char*
 Nebula::getType() const
 {
-    return "Nebula";
+    return NebulaTypeNames[static_cast<std::size_t>(type)].name;
 }
 
 void
-Nebula::setType(const std::string& /*typeStr*/)
+Nebula::setType(const std::string& typeStr)
```
ajtribick, line 54: Could take `std::string_view` instead of `const std::string&` here

src/celengine/nebula.cpp
```diff
@@ -62,6 +87,12 @@ Nebula::getObjType() const
 bool
 Nebula::load(const util::AssociativeArray* params, const fs::path& resPath, std::string_view name)
 {
+    
+    if (const auto* typeName = params->getString("Type"); typeName == nullptr)
+        setType({});
+    else
+        setType(*typeName);
```
375gnu, line 94:
```suggestion
    if (const auto* typeName = params->getString("Type"); typeName != nullptr)
        setType(*typeName);
```

src/celengine/nebula.cpp
```diff
@@ -22,23 +23,47 @@
 namespace engine = celestia::engine;
 namespace util = celestia::util;
 
+using namespace std::string_view_literals;
 using util::GetLogger;
 
+struct NebulaTypeName
```
ajtribick, line 29: The definitions of `NebulaTypeName` and `NebulaTypeNames` should be in an unnamed namespace

src/celestia/qt/qtdeepskybrowser.cpp
```diff
@@ -3,7 +3,7 @@
 // Copyright (C) 2008, Celestia Development Team
 // celestia-developers@lists.sourceforge.net
 //
-// Deep sky browser widget for Qt4 front-end
+// Deep sky browser widget for Qt5-Qt6 front-end
```
375gnu, line 6:
```suggestion
// Deep sky browser widget for Qt front-end
```
who knows when Qt7 arrives )

src/celestia/qt/qtdeepskybrowser.cpp
```diff
@@ -591,7 +602,7 @@ DeepSkyBrowser::slotRefreshTable()
     else
         filterPred.objectType = DeepSkyObjectType::OpenCluster;
 
-    if (filterPred.objectType == DeepSkyObjectType::Galaxy)
+    if ((filterPred.objectType == DeepSkyObjectType::Galaxy) || (filterPred.objectType == DeepSkyObjectType::Nebula))
```
ajtribick, line 605: no need for extra brackets

ajtribick (changes requested):

src/celestia/qt/qtdeepskybrowser.cpp
375gnu, line 190:
```suggestion
            std::string dson = dsodb.getDSOName(dso, true);
```
`getDSOName` returns std::string, not need to convert it

src/celestia/qt/qtdeepskybrowser.cpp
375gnu, line 211:
```suggestion
        std::string dson = dsodb.getDSOName(dso, true);
```

src/celestia/qt/qtdeepskybrowser.cpp
```diff
@@ -311,6 +320,8 @@ DeepSkyBrowser::DSOTableModel::data(const QModelIndex& index, int role) const
         return QString("%L1").arg((observerPos - dso->getPosition()).norm(), 0, 'g', 6);
     case AppMagColumn:
         {
+            if (dso->getAbsoluteMagnitude() == DSO_DEFAULT_ABS_MAGNITUDE) return QString();
```
375gnu, line 323:
```suggestion
            if (dso->getAbsoluteMagnitude() == DSO_DEFAULT_ABS_MAGNITUDE)
                return {};
```
please use one codestyle

src/celengine/nebula.h
```diff
@@ -24,6 +24,18 @@ namespace celestia::util
 class AssociativeArray;
 }
 
+enum class NebulaType
```
375gnu, line 27: return this back to `Nebula` but rename to `Type`, and then use `Nebula::Type` instead of `NebulaType`

src/celengine/nebula.cpp
```diff
@@ -22,23 +23,47 @@
...
+    return NebulaTypeNames[static_cast<std::size_t>(type)].name;
 }
 
 void
-Nebula::setType(const std::string& /*typeStr*/)
+Nebula::setType(const std::string& typeStr)
 {
+    type = NebulaType::NotDefined;
+    auto iter = std::find_if(std::begin(NebulaTypeNames), std::end(NebulaTypeNames),
+                             [&](const NebulaTypeName& n) { return n.name == typeStr; });
```
375gnu, line 58:
```suggestion
                             [&](const NebulaTypeName& n) { return compareIgnoringCase(n.name, typeStr) == 0; });
```

375gnu: don't forget `#include <celutil/stringutils.h>`

375gnu (changes requested):
i haven't tested the build, just checked the code

src/celestia/qt/qtdeepskybrowser.cpp
dave-kaye: Fixed spacing but leaving QString to guarantee Qt compatibility in future and across OSs.

src/celestia/qt/qtdeepskybrowser.cpp
dave-kaye: Fixed

src/celengine/nebula.cpp
dave-kaye: Fixed

src/celestia/qt/qtdeepskybrowser.cpp
dave-kaye: fixed

src/celestia/qt/qtdeepskybrowser.cpp
dave-kaye: fixed

src/celengine/nebula.cpp
dave-kaye: Fixed code to format.
Agree translation desirable but both nebula and galaxy will have to be changed for translation, they use the same codes.

src/celengine/nebula.cpp
dave-kaye: NebulaTypeName, NebulaTypeNames, NebulaType have same structure as galaxy versions. Multiple calling functions are virtual and cross deepskyobjects, nebula, galaxy, globular and opencluster.  Keep present code and you can decide on preferred code and implement across all DSOs to maintain uniformity in the future.

src/celengine/nebula.h
dave-kaye: NebulaTypeName, NebulaTypeNames, NebulaType have same structure as galaxy versions. Multiple calling functions are virtual and cross deepskyobjects, nebula, galaxy, globular and opencluster.  Keep present code and you can decide on preferred code and implement across all DSOs to maintain uniformity in the future.

src/celestia/qt/qtdeepskybrowser.cpp
dave-kaye: Want to keep return under Qt's control. QString() and QString("") send back nullptr and empty string which are different in Qt and potentially could change. Return{} will be evaluating a QVariant function that could have issues across multiple C++ versions and Qt versions and definately across these programs when looking at three different operating systems and multiple versions of those operating systems. It could compile and still fail while running which will be very difficult to track down. I saw several reports of random issues in return{} returns, but they are few.

Looking at it, it probably would be better to use QString("") so all good returns are strings and not a null that could cause errors in the future further downstream in QAbstractTableModel. I consider the return from this if condition to be valid, and not an error.  Just not logical in this situation. Don't know what QAbstractTableModel does internally in deciding and acting on invalid responses.

ajtribick: Looking at the docs, it seems that null/empty QString behaves identically except for the specific case of the `isNull` method, which they recommend against using. So we're probably ok with `QString()` here.

dave-kaye:
Authors have created models with multiple cmod files to draw an object in parts. The submitted code took this into account, removing parts with empty names in the *.dsc file from the DSO window display. However, I found some model builders did not use empty names, but rather blank space names. The code is updated to also remove from the DSO window display any model part with a name composed only of spaces.

This code uses a construct that was introduced in C++11.  If this is unacceptable, the names could be converted to qstrings, trim the string, then check the string for zero length to achieve the same result.

src/celengine/nebula.cpp
```diff
@@ -62,6 +88,10 @@ Nebula::getObjType() const
 bool
 Nebula::load(const util::AssociativeArray* params, const fs::path& resPath, std::string_view name)
 {
+    
+    if (const auto* typeName = params->getString("Type"); typeName != nullptr)
```
375gnu, line 92:
```suggestion
    if (const std::string* typeName = params->getString("Type"); typeName != nullptr)
```
The next line uses `std::string`, so let's keep the same style.

src/celengine/nebula.cpp
```diff
@@ -22,23 +24,47 @@
 namespace engine = celestia::engine;
 namespace util = celestia::util;
 
+using namespace std::string_view_literals;
 using util::GetLogger;
 
+struct NebulaTypeName
```
375gnu, line 30:
```suggestion
namespace
{
struct NebulaTypeName
```

src/celengine/nebula.cpp
```diff
@@ -22,23 +24,47 @@
...
+    NebulaTypeName{ " ", NebulaType::NotDefined },
+    NebulaTypeName{ "Emission", NebulaType::Emission },
+    NebulaTypeName{ "Reflection",  NebulaType::Reflection },
+    NebulaTypeName{ "Dark",  NebulaType::Dark },
+    NebulaTypeName{ "Planetary",  NebulaType::Planetary },
+    NebulaTypeName{ "SupernovaRemnant",  NebulaType::SupernovaRemnant },
+    NebulaTypeName{ "HII_Region", NebulaType::HII_Region },
+    NebulaTypeName{ "Protoplanetary", NebulaType::Protoplanetary },
+};
+
```
375gnu, line 47:
```suggestion
}

```

src/celengine/nebula.h
375gnu: your reply is not related to the requested. `NebulaType was defined inside `Nebula`, please restore this. But rename `NebulaType` to `Type`

375gnu (changes requested):

src/celengine/nebula.cpp
dave-kaye: done

375gnu (changes requested):

src/celestia/qt/qtdeepskybrowser.cpp
```diff
@@ -184,8 +184,16 @@ void populateDsoVector(std::vector<DeepSkyObject*>& dsos,
             return;
         }
 
-        if (DeepSkyObject* dso = dsodb.getDSO(index); filter(dso))
+        DeepSkyObject* dso = dsodb.getDSO(index);
+        if (filter(dso))
         {
+            std::string dson = dsodb.getDSOName(dso, true);
```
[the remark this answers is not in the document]
dave-kaye: The very first lines of code I wrote removed the blank names from the DSO load. I abandoned that code due to unexpected consequences. These names were used to add sections to the image. Removing them upon load also removed the image section. See below:

cats eye load all names
Cat's Eye Nebula loading all names


cats eye no load blank names
Cat's Eye Nebula not loading blank names

Trying to load only proper names would appear to require a complete re-write of the loading logic to allow multiple sub-images/models (mesh entries) in the single main dsc or ssc entry.  Or a variation of "component" would have to be added to *.dsc files that would allow image loading but be filtered out on the stellar body listing. Or requiring all the models be rebuilt with using only one mesh. That would be difficult since the models would need to be overlapping layers of semi-transparent materials.  

Requiring the modeler to have only proper names on all sections clutters the stellar body listing, making it unusable. That includes the bottom of the main window listing where you type a name and a display of selectable objects appears.   Using blank names allows image display and the submitted code logic removes them from the stellar body listing - decluttering the listing. That appears to be the best solution.

src/celestia/qt/qtdeepskybrowser.cpp
```diff
@@ -184,8 +184,16 @@ void populateDsoVector(std::vector<DeepSkyObject*>& dsos,
             return;
         }
 
-        if (DeepSkyObject* dso = dsodb.getDSO(index); filter(dso))
+        DeepSkyObject* dso = dsodb.getDSO(index);
```
375gnu, line 187:
```suggestion
        const DeepSkyObject* dso = dsodb.getDSO(index);
```
Not sure if it breaks something else, so please check

src/celestia/qt/qtdeepskybrowser.cpp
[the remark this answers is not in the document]
375gnu: okay, then, let's add a helper method:
```
    private bool isEmptyDSOName(const DeepSkyObject* dso) const
    {
        std::string name = dsodb.getDSOName(dso, true);
        if (name.empty())
            return true;
        if (std::all_of(name.begin(), name.end(), [](char c) {return std::isspace(c);}))
            return true;
        return false;
    }
```
And use it instead of copy-pasting

src/celestia/qt/qtdeepskybrowser.cpp
dave-kaye: Using `const` creates multiple compile errors on C++ vector usage. My version of C++ gives error message with potential suggestions then next line gives statement using suggestions could create ill-formed conversions. Best not to make change as compiler versions could react unexpectedly.

src/celestia/qt/qtdeepskybrowser.cpp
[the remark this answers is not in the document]
dave-kaye: Code modified to work and uploaded.

ajtribick (approved):

PR #2321, opened 2025-06-09
[SDL] Make ImGui menu optional

dave-kaye:
SDL has been an interface with no menus, allowing other uses for this view. A recent addition of minor menus with ImGui was made. This code adds a user settable option in the CMake system to compile SDL with/without the ImGui menu.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ option(USE_ICU            "Use ICU for UTF8 decoding for text rendering (Default
 option(USE_WIN_ICU        "Use Windows SDK's ICU implementation (Default: off)" OFF)
 option(USE_MESHOPTIMIZER  "Use meshoptimizer (Default: off)" OFF)
 option(USE_WEFFCPP        "Use the -Weffc++ option when compiling with GCC (Default: off)" OFF)
+option(USE_IMGUI          "Use ImGui for menu system on SDL frontend (Default: off)" OFF)
 option(LEGACY_OPENGL_LIBS "Use legacy OpenGL libraries instead of glvnd library (Default: off)" OFF)
 
 # Prefer GLVND or "legacy" OpenGL library (libOpenGL.so vs libGL.so)
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -140,9 +140,10 @@ sudo make install
 
 [*] `INTERFACE` must be replaced with one of "`QT5`", "`QT6`" or "`SDL`"
 
-Four interfaces are available for Celestia on Unix-like systems:
+Three interfaces are available for Celestia on Unix-like systems:
 - SDL: minimal interface, barebone Celestia core with no toolbar or menu...
        Disabled by default.
+       Optional minimal menu is available.
 - QT5: A full interface with minimal dependencies, adds a menu, a configuration
        dialog some other utilities, bookmarks... A preferred option. Enabled by
        default, No need to pass -DENABLE_QT5=ON.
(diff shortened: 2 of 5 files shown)


ajtribick:
Instead of adding yet more combinations of build configurations, I think it would be better to have the UI hidable at runtime, e.g. have an "auto-hide menubar" toggle and only render it when the mouse is in the top part of the window.

dave-kaye:
It is not just about the menu.  The SDL version is a smaller, portable program.  It compiles to 3.6 MB, but with ImGui it is 16.4 MB.

The SDL program has few dependencies.  When toolboxes such as Qt and Gtk are  used, they have many routines that are encoded into the toolbox.  ImGui uses more system libraries.  It took a long time for me to get the interface to compile.  Slight differences in my OS caused compile crashes.  It was traced back to system supplied libraries.  I wrote the CMAKE option to bypass the ImGui code for those instances that the menu is not wanted, and the size/dependencies are an issue. 

The bypass uses the old SDLmain.cpp.  If ImGui sits on top of the SDL code, then there may be a way to use that version of what mimics the old SDL code. That would decrease the maintenance.  I did write it as an option, so as not to make a tree of SDL code and a second tree of SDL-ImGui code, which would substantially increase maintenance.

ajtribick:
While the size of the celestia-sdl executable does increase, on both Linux and Windows x64 this is still substantially smaller than the celestia shared library. 

Release build (note: NOT RelWithDebInfo)

Windows: celestia.dll 16.3 MiB, celestia-sdl.exe 33.5 KiB -> 525 KiB
Linux: libcelestia.so 3.89 MiB, celestia-sdl  44.4 KiB -> 867 KiB

So yes, while it does result in a fairly significant jump in the size of the SDL front-end part, the bulk of the application size is the core library. As such, I do not see that the smaller size is worth the cost of introducing another compile-time configuration that we do not build or are likely to actively use (particularly given the inability to set preferences in the renderer).

As for pulling in more system libraries, the OpenGL backend links to the OpenGL/OpenGLES libraries, which would already be linked into the main application, and the SDL2 backend links to SDL2, which is already linked by the celestia-sdl executable.

Your comments about Qt and GTK are irrelevant here, yes they are heavyweight libraries, but ImGui uses neither.

Without knowing what you are doing to make the compilation difficult for yourself, I'm not sure why you had so much trouble compiling with ImGui, the library code itself should be in a submodule.

dave-kaye:
The issues are with included files that the ImGui code is adding that the initial SDL, Qt and Gtk interfaces did not seem to access.

An included file was causing compiler errors.  The system was using incorrect files, and adding additional path information fixed the issue.

A compiler error stating it could not compile the error messages found in an included file then appeared. It was caused by my OS not allowing direct access to the included file; access via a macro was required. Changing the include statement from the file to the macro fixed this issue.

The fixes were simple, but finding the causes was a problem.  The error messages from gcc were not very helpful.  None of these issues has appeared before - on any programs that I compile. This is my concern. If others have problems, I wrote the code to bypass the ImGui.  I have seen other projects that supposedly use Celestia within their programs, but that was some time ago, and they may not even exist anymore.  If you don't think that this is an issue, just deny the submission.  It will remain in the history, and if an issue should surface it can be revisited then.  However, I think that making a "disappear" option for the ImGui menu is a good idea and should be done.  Perhaps a command line option like -nogui would work better than a menu item.

ajtribick:
Closing this, we have too many configuration combinations already.

PR #2322, opened 2025-06-09
[Qt]  Celestial browser sort indicator

dave-kaye:
Several tables in the Celestial Browser panel are hard coded to sort ascending by distance (column 2) upon display/changes. The panel is using the default table display which places the sort indicator as descending in the first column. This code fixes this sync issue.

### DIFF
--- a/src/celestia/qt/qtcelestialbrowser.cpp
+++ b/src/celestia/qt/qtcelestialbrowser.cpp
@@ -30,6 +30,7 @@
 #include <QComboBox>
 #include <QGridLayout>
 #include <QGroupBox>
+#include <QHeaderView>
 #include <QItemSelectionModel>
 #include <QLabel>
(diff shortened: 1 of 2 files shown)


dave-kaye:
The "Stars" and "Deep Sky Objects" tables in the Celestial Browser panel are hard coded to sort ascending by distance (column 2) upon initial display/changes. The panel is using the default table display which places the sort indicator as descending in the first column.

The code has been tested with initial display of the tables; changes in filter to change the number of items listed; switching between tables; and using other columns as the sorting column. It has also been tested using drag-and-drop changing of the location of the distance column (click on the column header and move it to a different location).

Multiple combinations and sequences have been tested. It appears to accurately maintain the sorting indicator in all situations.

375gnu (approved):

PR #2325, opened 2025-06-14
Make GalaxyRenderer/GlobularRenderer own textures and OpenGL objects

ajtribick:
Ensures proper cleanup when Renderer is destroyed.


ajtribick:
I have disabled the "Detect Missing System Include Files" check in Codacy/cppcheck - this seems to be firing a bunch of false positives in recent PRs.

375gnu (approved):

PR #2327, opened 2025-06-16
Move GLShader/GLProgram onto the stack

ajtribick:
- Reduce some heap allocations
- Replace GLShaderLoader with static methods on the shader/program classes
- Use smart pointers in shadermanager
- Replace use of stringstream with fmt in shadermanager

### DIFF
--- a/src/celengine/glshader.cpp
+++ b/src/celengine/glshader.cpp
@@ -150,12 +150,6 @@ GLShader::GLShader(GLuint _id) :
 {
 }
 
-GLShader::~GLShader()
-{
-    if (id != 0)
-        glDeleteShader(id);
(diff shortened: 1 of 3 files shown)


ajtribick:
Deprecated rule [cpp:S4963](https://sonarcloud.io/organizations/celestiaproject/rules?open=cpp%3AS4963&rule_key=cpp%3AS4963) should probably be removed from the Sonar profile - don't have admin rights so can't do that.

ajtribick:
Remaining Sonar issues: 2 complexity ones (not fixing that now), 1 from deprecated rule

375gnu (approved):

Issue #2176, opened 2024-04-19
Constellation lines beyond Solar System


SpaceExplorer2008:
Proposal to reintegrate the visualization of constellation lines beyond Solar System. It would be a useful feature to give people an idea of ​​the true distances between stars in the same constellation.

AstroChara:
Implemented; now it is possible to specify fade start and fade end distances (in light-years) for the three elements of the constellation in the config file.

Disclaimer: I haven't actually tested the case where the fade start is larger than fade end; we might add handling for that later but for now please avoid that

PR #2331, opened 2025-06-21
Fix constellation line fade distance

AstroChara:
As per Issue #2176. Removed the random scaling factor applied to “dist” value, which is confusing and unintuitive. Now constellation lines should show up at interstellar distances, as preferred by many users. If fading at smaller distance is preferred then it would make sense to change the constants themselves rather than this.

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -3132,8 +3132,8 @@ void Renderer::renderBoundaries(const Universe& universe, float dist, const Matr
     //*** Boundaries fading parameters
 
     //   The two values are for distances from centre where the boundaries begins fading and finish fading, respectively.
-    const float RenderBoundariesFadeStartDist     = detailOptions.renderBoundariesFadeStartDist;
-    const float RenderBoundariesFadeEndDist     = detailOptions.renderBoundariesFadeEndDist;
+    const float RenderBoundariesFadeStartDist = detailOptions.renderBoundariesFadeStartDist;
+    const float RenderBoundariesFadeEndDist   = detailOptions.renderBoundariesFadeEndDist;
     
     if (!util::is_set(renderFlags, RenderFlags::ShowBoundaries) || boundaries == nullptr)
         return;
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -4026,8 +4026,8 @@ void Renderer::labelConstellations(const AsterismList& asterisms,
     //*** Constellation label fading parameters
 
     //   The two values are for distances from centre where the labels begins fading and finish fading, respectively.
-    const float LabelConstellationsFadeStartDist     = detailOptions.labelConstellationsFadeStartDist;
-    const float LabelConstellationsFadeEndDist     = detailOptions.labelConstellationsFadeEndDist;
+    const float LabelConstellationsFadeStartDist = detailOptions.labelConstellationsFadeStartDist;
+    const float LabelConstellationsFadeEndDist   = detailOptions.labelConstellationsFadeEndDist;
 
     for (const auto& ast : asterisms)
     {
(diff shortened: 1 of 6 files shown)


375gnu:
If for some reason you need this, then:

1. add a new boolean flag in Render
2. add a new interface option to toggle distance fading

AstroChara:
Eventually it may be a good idea to add a way to manually input custom fade distances for the three elements of constellations in the program. For now, simply turning the constellation lines off should be sufficient as it does more or less the same thing fading asterism lines outside the Solar System does.

pedro-fixingstuff:
I think exposing the fade distances as settings in celestia.cfg would be good. It makes sense to have the fade itself be a toggleable setting (with the default being enabled!).

AstroChara:
> I think exposing the fade distances as settings in celestia.cfg would be good. It makes sense to have the fade itself be a toggleable setting (with the default being enabled!).

I support this idea. It allows full customizability of the fading while not being too difficult to implement (hopefully) nor taking too much space in the setting (you need three sliders for three elements).

AstroChara:
> I think exposing the fade distances as settings in celestia.cfg would be good. It makes sense to have the fade itself be a toggleable setting (with the default being enabled!).

I have made an attempt to implement this though I cannot guarantee it would work or if I did not miss anything.

ajtribick (approved):

src/celengine/render.cpp
```diff
@@ -4013,6 +4023,12 @@ void Renderer::labelConstellations(const AsterismList& asterisms,
 {
     Vector3f observerPos = observer.getPosition().toLy().cast<float>();
 
+    //*** Constellation label fading parameters
+
+    //   The two values are for distances from centre where the labels begins fading and finish fading, respectively.
+    const float LabelConstellationsFadeStartDist     = detailOptions.labelConstellationsFadeStartDist;
+    const float LabelConstellationsFadeEndDist     = detailOptions.labelConstellationsFadeEndDist;
```
375gnu, line 4030:
```suggestion
    const float LabelConstellationsFadeStartDist = detailOptions.labelConstellationsFadeStartDist;
    const float LabelConstellationsFadeEndDist   = detailOptions.labelConstellationsFadeEndDist;
```

src/celengine/render.cpp
```diff
@@ -3125,6 +3128,13 @@ void Renderer::renderAsterisms(const Universe& universe, float dist, const Matri
 void Renderer::renderBoundaries(const Universe& universe, float dist, const Matrices& mvp)
 {
     auto boundaries = universe.getBoundaries();
+
+    //*** Boundaries fading parameters
+
+    //   The two values are for distances from centre where the boundaries begins fading and finish fading, respectively.
+    const float RenderBoundariesFadeStartDist     = detailOptions.renderBoundariesFadeStartDist;
+    const float RenderBoundariesFadeEndDist     = detailOptions.renderBoundariesFadeEndDist;
```
375gnu, line 3136:
```suggestion
    const float RenderBoundariesFadeStartDist = detailOptions.renderBoundariesFadeStartDist;
    const float RenderBoundariesFadeEndDist   = detailOptions.renderBoundariesFadeEndDist;
```

src/celengine/render.cpp
```diff
@@ -3095,6 +3092,12 @@ void Renderer::renderAsterisms(const Universe& universe, float dist, const Matri
 {
     auto *asterisms = universe.getAsterisms();
 
+    //*** Asterism fading parameters
+
+    //   The two values are for distances from centre where the asterism begins fading and finish fading, respectively.
+    const float RenderAsterismsFadeStartDist     = detailOptions.renderAsterismsFadeStartDist;
+    const float RenderAsterismsFadeEndDist     = detailOptions.renderAsterismsFadeEndDist;
```
375gnu, line 3099:
```suggestion
    const float RenderAsterismsFadeStartDist = detailOptions.renderAsterismsFadeStartDist;
    const float RenderAsterismsFadeEndDist   = detailOptions.renderAsterismsFadeEndDist;
```

375gnu (changes requested):

375gnu (approved):

375gnu:
Don't forget to use "squash and merge"

PR #2338, opened 2025-07-01
Require C++ std::filesystem

levinli303:


ajtribick:
I am in favour of this but maybe there are some Linux distros that are still using older compilers that only have the experimental filesystem library? Worth checking with @munix9 here I think.

munix9:
I will test this on [obs](https://build.opensuse.org/package/show/home:munix9:unstable/celestia) and get back to you with the result.

The test has run and it looks good, but I can't say anything about other distributions and (older) compiler versions.

levinli303:
> The test has run and it looks good, but I can't say anything about other distributions and (older) compiler versions.

gcc version >= 8 should all build fine, I think we are good

375gnu (approved):

ajtribick (approved):

PR #2340, opened 2025-07-03
Use VertexObject to draw solid rectangles

levinli303:


ajtribick:
Looks ok to me, although I'm not sure about having the parameter be `const std::optional<T>&` vs `const T*`. (I see that at least in some places, the caller actually has the optional, and there isn't a built-in API for doing this.

Maybe worth introducing a utility function along the lines of:

```C++
template<typename T>
const T*
as_ptr(const std::optional<T>& opt) noexcept
{
    return opt.has_value() ? std::addressof(*opt) : nullptr;
}
```

but maybe this is overkill.

I defer to @375gnu for the OpenGL review.

375gnu:
Hmm, i don't think that modifying VAO is a good idea. I remember I tried to implement something similar one day, but decided to postpone changes till time when Celestia require GL >= 3.0.

@levinli303 could you elaborate why this change is needed?

levinli303:
> Hmm, i don't think that modifying VAO is a good idea. I remember I tried to implement something similar one day, but decided to postpone changes till time when Celestia require GL >= 3.0.
> 
> @levinli303 could you elaborate why this change is needed?

it's not really needed for anything, just hoping to move all GL draw calls to VertexObject so making this change. my thought is that the places where we need to call drawRectangle in HUD/overlay image, we only need to change VAO when window size changes, so the actual code for changing VAO might not be called at all.

375gnu:
If I remember correctly, GL > 2 (I don't remember what exact version) doesn't allows VAO changing. So proper forward compatible implementation will require usage of several VAOs. I think we should bump required GL version to at least 3.0 to have hardware sRGB w/o pain.

PR #2347, opened 2025-07-09
Add a method in Observer to applyTransform

levinli303:
When moving the camera using gyroscope/rotation vector, I hope to use `setOrientationTransform` instead of `rotate` on observer to achieve the effect that the camera follows the device's orientation. This is 

When user enables gyroscope, we call `setOrientationTransform` to update the transform with device input, and when user disables gyroscope, we apply the transform onto the `originalOrientation` and resets the transform to identity, so that there is no change in the rendered content and when users enables gyroscope again we can start clean.


375gnu (approved):
if it works, then i'm ok

PR #2352, opened 2025-07-12
Use the starting orientation transform value throughout journey

levinli303:
This does not affect Celestia frontends here since the transform is always .identity.

The initial/final orientation of journeys should be the same, because in VR, user might look away and the orientation transform changes.

### DIFF
--- a/src/celengine/observer.cpp
+++ b/src/celengine/observer.cpp
@@ -663,7 +663,7 @@ Observer::update(double dt, double timeScale)
 
         // Spherically interpolate the orientation over the first half
         // of the journey.
-        originalOrientation = undoTransform(interpolateOrientation(journey, t));
+        originalOrientation = journey.orientationTransformInverse * interpolateOrientation(journey, t);
 
         // If the journey's complete, reset to manual control
         if (t == 1.0f)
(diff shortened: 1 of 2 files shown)


src/celengine/observer.cpp
```diff
@@ -578,14 +578,14 @@ Observer::setOriginalOrientation(const Eigen::Quaterniond& q)
     updateOrientation();
 }
 
-Eigen::Matrix3d
+Eigen::Quaterniond
```
ajtribick, line 581: Maybe it would be useful to return this as `const Eigen::Quaterniond&` to avoid forcing the copy?

src/celengine/observer.h
```diff
@@ -259,14 +259,15 @@ class Observer
...
+        Eigen::Quaterniond orientationTransformInverse; // the inverse of orientation transform when the journey starts
 
         Selection centerObject;
 
         TrajectoryType traj{ TrajectoryType::Linear };
     };
 
     void gotoJourney(const JourneyParams&);
-    // void setSimulation(Simulation* _sim) { sim = _sim; };
+    void startTraveling();
```
ajtribick, line 270: Does this need to have public visibility? In this PR it looks like it's only being used by `Observer` internal methods

levinli303: yeah can be private

PR #2358, opened 2025-07-12
Add support for a second thumbstick

levinli303:
can be potentially used for right joystick on the game controller. unlike the yaw/pitch with existing joystick which might cause roll, this one store Euler angles and has pitch clamped between ±90 so won't cause rolls, works more like a planetarium.


levinli303:
@375gnu @ajtribick  this change is ready. it does not affect the frontends here.

PR #2361, opened 2025-07-15
Warp mesh fixes

ajtribick:
- Check that the contents of the map file represent a grid with monotonically increasing x- and y-coordinates. This is slightly more lenient than the existing code that assumed uniform grid spacing without verification. A fully general implementation would need to use the triangles directly.

- Use an index buffer to hold the triangle coordinates, rather than duplicating the vertex data. This significantly reduces the amount of memory used, particularly when the vertex count is small enough to fit into a GLushort.

- As there is only one warp mesh, and changing this should be an infrequent operation (existing code loads one warp mesh at startup), do not use the resource manager machinery. This removes the possibility to use a wildcard extension, but only one extension was supported anyway.

- Use smart pointers to track resources rather than manual new/delete.


src/celestia/celestiacore.cpp
```diff
@@ -2497,10 +2497,9 @@ bool CelestiaCore::initSimulation(const std::filesystem::path& configFileName,
             }
             else
             {
-                WarpMeshManager *manager = GetWarpMeshManager();
-                WarpMesh *mesh = manager->find(manager->getHandle(WarpMeshInfo(config->paths.warpMeshFile)));
+                auto mesh = WarpMesh::load(config->paths.warpMeshFile);
```
levinli303, line 2500: need to load warp/warpMeshFile

levinli303 (commented):
I keep getting memory corruption error when I try your branch. you can test by uncommenting celestia.cfg's ProjectionMode/ViewportEffect/WarpMeshFile
Screenshot 2025-07-16 at 9 43 59 PM

src/celestia/celestiacore.cpp
ajtribick: I can't reproduce this from a clean build on Windows or Linux. I did see something similar at one point during development due to what looked like an incremental build error (this persisted when I switched back to the master branch), but clearing out the build directory and doing a clean rebuild fixed it.

levinli303: looks fixed now

PR #2364, opened 2025-07-17
Implement orbital precession

### LINKED ISSUE #760: Orbital precession
Would be good to be able to specify orbital precession in ssc files. It would probably use a `PrecessingOrbit` block, similar to `PrecessingRotation`.

pedro-fixingstuff:
This adds two new parameters for modeling of [nodal](https://en.wikipedia.org/wiki/Nodal_precession) and [apsidal](https://en.wikipedia.org/wiki/Apsidal_precession) precession: `NodalPrecessionPeriod` and `ApsidalPrecessionPeriod`. The default unit for both is years.

Example syntax (orbit data from 16 Cyg Bb):
```
	EllipticalOrbit
	{
		Epoch 2456937
		Period 2.18863792
		SemiMajorAxis 1.6795
		Eccentricity 0.683
		Inclination 137.41309
		AscendingNode 316.95447
		ArgOfPericenter 232.8624
		MeanAnomaly 1.03341
		NodalPrecessionPeriod 100
		ApsidalPrecessionPeriod 100
	}
```

For the following images, I've set `OrbitPeriodsShown` to 3.0 in celestia.cfg.

**Nodal precession:**

**Apsidal precession:**

One issue is that the orbit paths look incomplete (this is with fading disabled). I wonder if using some parameter other than period (such as eccentric anomaly) for sampling the orbit would address that.

I'm not sure if the equations used are correct either.

### DIFF
--- a/src/celastro/astro.h
+++ b/src/celastro/astro.h
@@ -244,6 +244,8 @@ struct KeplerElements
     double argPericenter{ 0.0 };
     double meanAnomaly{ 0.0 };
     double period{ 0.0 };
+    double nodalPeriod{ 0.0 };
+    double apsidalPeriod{ 0.0 };
 };
 
 
--- a/src/celengine/parseobject.cpp
+++ b/src/celengine/parseobject.cpp
@@ -133,6 +133,11 @@ GetDefaultUnits(bool usePlanetUnits, double& distanceScale)
  *     # One or none of the following:
  *     MeanAnomaly <degrees>     (default: 0.0)
  *     MeanLongitude <degrees>   (default: 0.0)
+ * 
+ *     # For precessing orbits (default unit is Julian years):
+ *     # Default values result in no precession
+ *     NodalPrecessionPeriod <number>   (default: 0.0)
+ *     ApsidalPrecessionPeriod <number> (default: 0.0)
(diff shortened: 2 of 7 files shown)


pedro-fixingstuff:
The latest commit fails to build as it is right now

AstroChara:
Regarding precession causing the no-fading orbit to be incomplete, we might want to instead draw the Keplerian orbit at that instant (a Keplerian orbit with the current longitude of ascending node and argument of periapsis). This is the approach Universe Sandbox uses for its equivalent orbit render mode, and in absence of fading, I think it's about the best option we have to render these orbits.

pedro-fixingstuff:
I feel like this is ready to be taken a look now.

About rendering the path of the osculating orbit when fading is disabled, my opinion is that it's better left for another moment, as that would mean a change from the current paradigm of showing the actual path an object has traversed/will traverse.

For most realistic cases of precession, the difference between either orbital path wouldn't be significant anyway...

AstroChara:
Just to check, is there anything else that needs to be done here or is this ready for approval and merger?

pedro-fixingstuff:
Critical issues... I might have introduced a memory leak after all

PR #2377, opened 2025-07-26
add separate objective-c++ file for miniaudio on Apple platforms

levinli303:


ajtribick:
I guess we need this unless we update to something like CMake 3.20, where the `LANGUAGE` property seems to [actually behave in a useful manner](https://cmake.org/cmake/help/v3.20/prop_sf/LANGUAGE.html#prop_sf:LANGUAGE) and the `OBJCXX` language is available since CMake 3.16.

375gnu (approved):

PR #2379, opened 2025-07-26
Fixes/improvements to array_view

ajtribick:
- Allow creating array_views from compatible types, e.g. const T* from T*
- Remove constructors from rvalue collections - if collection goes out of scope, view is invalid
- Remove const void specialization - array_view is always to const
- Remove usages of array_view<const T> in codebase

Compatibility check is based on the check for C++20 `std::span` described on the [constructors page at cppreference](https://en.cppreference.com/w/cpp/container/span/span.html#Conversion_requirements).

### DIFF
--- a/src/celutil/array_view.h
+++ b/src/celutil/array_view.h
@@ -14,7 +14,7 @@
 #include <array>
 #include <cstddef> // std::size_t
 #include <vector>
-#include <type_traits> // std::remove_cv
+#include <type_traits>
 
 namespace celestia::util
 {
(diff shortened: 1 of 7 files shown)


src/celutil/array_view.h
```diff
@@ -231,38 +235,31 @@ class array_view<void>
...
-        m_ptr(vec.data()),
-        m_size(vec.size() * sizeof(T))
-    {}
-
     /**
      * Copy another view.
      */
-    template<typename T> constexpr array_view(const array_view<T> &ary) :
-        m_ptr(ary.data()),
-        m_size(ary.size())
```
ajtribick, line 245: there appears to have been a missing `* sizeof(T)` here

src/celutil/array_view.h
```diff
@@ -231,38 +235,31 @@ class array_view<void>
...
-    {}
+    constexpr array_view(const array_view&) = default;
 
-    /** @copydoc array_view::array_view(const array_view<T> &ary) */
-    template<typename T> constexpr array_view(array_view<T> &&ary) :
```
ajtribick, line 249: No resources to steal so have left out the move constructors, copying is fine here.

Issue #1131, opened 2021-10-13
[1.7.0] LunarLambert sometimes makes objects too bright


SevenSpheres:
**Describe the bug**
While updating content, I noticed that some objects (specifically, Jupiter and Europa) had become excessively bright. Removing `LunarLambert 0.5` from the .ssc file seems to fix this. I don't know why it affects only some objects and not others.

**To Reproduce**
1. Add `LunarLambert 0.5` to the code for Jupiter/Europa, or just take the files from [this branch](https://github.com/CelestiaProject/CelestiaContent/tree/update-content)
2. Go to Jupiter/Europa in Celestia
3. See error

**Expected behavior**
Jupiter and Europa appear with normal brightness, like they would in 1.6.x with the same code.

**Screenshots**
![juptoobright](https://user-images.githubusercontent.com/56284207/137213136-0b818459-8b3e-4210-bc4a-87750d5907f9.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Frontend: Qt, Win
 - Version: 1.7.0, [latest working build](https://github.com/CelestiaProject/Celestia/actions/runs/1210787439)

**Additional context**
Possibly related to #1114?

ajtribick:
For comparison, here's what I see when using Saturn with LunarLambert 0.5 as rendered by Celestia 2013 (7033ec32 + minimal changes to get it to compile with Qt5), Celestia 1.7.0 (ec232b7f) and Celestia 1.6.2 (eb4e1a64 + minimal changes to get it to compile with Qt5). Ambient light is set to zero.

![out](https://user-images.githubusercontent.com/3002689/139531995-1bacc6c4-4935-4e22-9787-4c2c0e37e832.jpg)

It looks like Celestia 1.7 is consistent with Celestia 2013, but something was changed that made the rendering in 1.6.2 darker.

SevenSpheres:
Enceladus comparison. Using the darker old texture, it *is* brighter in 1.7 - I don't know if this is the correct behavior or not. Using the brighter texture from Project Echoes, the 1.7 render isn't just brighter, it's *shiny* as if it had specular enabled. The latter is really the issue I'm referring to.
![enceladus](https://user-images.githubusercontent.com/56284207/139538881-e81f3f60-a60e-4982-b8c8-6489f98fe33b.png)

ajtribick:
[...]

PR #2382, opened 2025-08-02
Make LunarLambert phase dependent

pedro-fixingstuff:
This PR reworks the implementation of the lunar-Lambert shading model to match more closely the definition from works by Bonnie Buratti et al. (as an example, I cite her [2019 paper on Charon](https://iopscience.iop.org/article/10.3847/2041-8213/ab0bff)). The model is a combination of the Lambert and Lommel-Seeliger (lunar-like) disk functions, partitioned by _A_ (which Celestia's `LunarLambert` parameter is meant to match)

Equation 1 from that paper shows that the Lommel-Seeliger component is dependent on the phase function f(alpha). This means that, for increasing phase angles, the shading converges to the Lambertian model.

The phase function is approximated here as an exponential in order to make it as generic as possible, and it is scaled using a derivation of Equation 3 from that paper.

Since this introduces a normalization factor to keep the overall brightness constant, I've removed the original form of it, which is the cause of #1131.

**I came across some issues, however:**
- Some objects, such as Europa, became invisible from the first commit;
- Starting on the second commit, Earth turns black.


pedro-fixingstuff:
Closed in favor of #2566

PR #2389, opened 2025-09-04
Search for Qt6GuiPrivate with Qt 6.10

heirecka:
Usage of private Qt modules requires a call to

`find_package(Qt6 COMPONENTS FooPrivate)` since 6.10 [1].

Also replace Qt6Gui_PRIVATE_INCLUDE_DIRS with a
target_link_libraries(... Qt6::GuiPrivate ...) call. It has the same effect and the advantage, that it produces an error if not available (contrary to Qt6Gui_PRIVATE_INCLUDE_DIRS just being empty in the 6.10 case for example).

[1] https://doc-snapshots.qt.io/qt6-dev/whatsnew610.html#build-system-changes


375gnu:
Thanks!

PR #2406, opened 2025-11-27
Make starnames parsing line-ending agnostic

ajtribick:
- Add buffered file handler
- Handle both CR and LF in starnames loader
- Allow UTF-8 BOM in starnames.dat

The buffered file reader will be useful for simplifying the tokenizer implementation, but I don't have time to do that right now. Something for the next PR.

### DIFF
--- a/src/celutil/utf8.cpp
+++ b/src/celutil/utf8.cpp
@@ -12,7 +12,9 @@
 
 #include <cassert>
 #include <cwctype>
+#include <istream>
 
+using namespace std::string_view_literals;
 
 namespace
 {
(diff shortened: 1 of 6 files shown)


src/celutil/utf8.cpp
```diff
@@ -464,6 +466,26 @@ bool UTF8StartsWith(std::string_view str, std::string_view prefix, bool ignoreCa
...
+std::istream&
+SkipUTF8BOM(std::istream& in)
+{
+    constexpr std::string_view bom = "\357\273\277"sv;
+
+    auto pos = in.tellg();
+    std::array<char, bom.size()> data;
+    in.read(data.data(), bom.size());
+    if (in.bad())
+        return in;
```
375gnu: sh'ld you restore the pos?

375gnu (approved):

PR #2418, opened 2025-12-14
curveplot: remove static vertex buffer

ajtribick:
Store the vertex buffer in the Renderer to constrain lifetime


ajtribick:
The duplication Sonar is complaining about was already in the file, so not going to fix that here.

375gnu (approved):

PR #2420, opened 2025-12-24
Various path handling fixes

ajtribick:
- [qt] Use the path constructor from UTF-16 when converting from QString
- [linux] Fix usage of getpwuid_r based on man page
- [win] Replace deprecated SHGetFolderPathW with SHGetKnownFolderPath
- [win] Fix handling of _wgetenv_s return values


src/celestia/qt/qtappwin.cpp
```diff
@@ -251,9 +254,9 @@ CelestiaAppWindow::init(const CelestiaCommandLineOptions& options)
     if (!logPath.makeAbsolute())
         QMessageBox::warning(0, "Celestia", _("Error getting path for log filename!"));
 
-    QString celestia_data_dir = options.startDirectory.isEmpty()
-        ? QString::fromLocal8Bit(::getenv("CELESTIA_DATA_DIR"))
-        : options.startDirectory;
+    QString celestia_data_dir = options.startDirectory;
```
375gnu: As you're touching this code, could you rename `celestia_data_dir` to `celestiaDataDir` so it matches the file style?

375gnu (approved):

PR #2422, opened 2025-12-27
Rewrite CMOD loader to avoid std::function

ajtribick:
- Results in a cleaner API with more explicit allocation semantics
- Remove comparator in mesh sort API: always uses opacity comparison

Probably conflicts with #2420 - I suggest merging that one first then updating this.


ajtribick:
Merged in #2431

PR #2428, opened 2025-12-29
De-vendor CSPICE and disable Spice in the CI builds

ajtribick:
Proprietary CSPICE license is likely incompatible with Celestia, so to stay on the safe side, de-vendor the Spice code and do not create Spice-enabled builds in CI. The relevant code remains in the codebase for users who wish to compile their own Spice-enabled versions.


375gnu (approved):
then we must be careful not to break spice integration

PR #2429, opened 2025-12-29
Use fast_float to supply missing charconv functions

ajtribick:
Replaces the custom implementation in celcompat with https://github.com/fastfloat/fast_float


ajtribick:
Another perhaps simpler option would be to just use fast_float regardless of whether the system has float charconv support. Although then we might need to vendor it because it doesn't seem to be in the Ubuntu 22.04 repositories.

375gnu (approved):

ajtribick:
Rebased to fix merge conflict

PR #2430, opened 2025-12-30
Remove unused Watcher class

ajtribick:
None of the front-ends use it, except Windows. Use the RendererWatcher instead, which does the same thing.


ajtribick:
Windows front-end needs to go.

375gnu (approved):

PR #2431, opened 2025-12-31
Refactor texture management

### LINKED ISSUE #649: Sprite objects loaded from DSC files look only for medres textures
**Describe the bug**
Sprite objects loaded from DSC files look only for medres textures, lores and hires are ignored

**To Reproduce**
Use "Disk Test" from https://github.com/LukeCEL/spritestest

ajtribick:
Like #2421 but for textures. Includes changes from #2422

`MultiResTexture` goes away and is replaced by a single integer (actually enum class) `TextureHandle` - resolution selection is now done by `TextureManager`. Changing resolution results in unloading of GPU textures. Paths to all resolutions are resolved at the time of creation of a new `TextureHandle`.

Note that ring shadow textures are loaded at a maximum of medium resolution, so there is some additional handling of that. On the other hand, unlike meshes the image data is not queried once it goes to the GPU (for meshes the CPU-side data is used for testing object selections), so there is only one `TextureManager` class instead of a `TextureManager`/`RenderTextureManager` split.

Fixes #649


375gnu:
It's a big change to go through right now, does it fix #649?

ajtribick:
Just checked, yes it does fix that issue.

Might be easier to do #2422 first, then I can rebase this one to have only the texture manager changes.

ajtribick:
Need to fix some issues in #2422

ajtribick:
Issue mentioned in #2439 is fixed

375gnu (approved):

PR #2455, opened 2026-02-04
Add directory to TexturePaths key

### LINKED ISSUE #2454: Incorrect texture loaded
When two add-ons reference and have their own texture with the same name, one of the textures is used for both of them.

To reproduce:

Use the two add ons from:
[bug.zip](https://github.com/user-attachments/files/25078754/bug.zip)

it has a folder 1, and folder 2. both add-ons (were modified) to reference and have a gliese625b.png. if you go to Gliese 625 b (and turn off the atmosphere), you'll see the gliese625b.png from add-on folder 1 is used but the Gliese 625 b is actually defined in add-on folder 2.

@ajtribick

levinli303:
closes #2454

### DIFF
--- a/src/celengine/texmanager.cpp
+++ b/src/celengine/texmanager.cpp
@@ -107,6 +107,27 @@ TexturePaths::InfoEqual::operator()(const Info& lhs, const Info& rhs) const
            std::tie(rhs.pathSet, rhs.flags, rhs.bumpHeight);
 }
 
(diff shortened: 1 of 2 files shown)


ajtribick:
Simpler fix #2456

ajtribick:
Actual cause of the bug is because the directory cache lookup is using the filename rather than the directory path. Oops.

PR #2469, opened 2026-02-15
Rebuild fonts when DPI changes

levinli303:
1. Use scaling factor from drawable height / window height, instead of `SDL_GetDisplayDPI`
2. Multiply by scaling factor in safe area
3. Reset font on DPI change


ajtribick:
Thanks - didn't have a High DPI monitor to test ImGui with when I was setting this up.

PR #2474, opened 2026-02-27
Migrate draw_rectangle_solid to use VertexObject/Buffer classes

levinli303:


src/celengine/render.h
```diff
@@ -707,6 +707,10 @@ class Renderer
     std::unique_ptr<celestia::gl::Buffer> m_markerBO;
     bool m_markerDataInitialized{ false };
 
+    mutable std::unique_ptr<celestia::gl::VertexObject> m_rectVO;
```
ajtribick: Only `m_rectInitialized` needs to be mutable

ajtribick (changes requested):

PR #2475, opened 2026-03-06
Add an option to render annotations in 3D space

levinli303:
Add a new render flag, when ShowTangentAnnotations is enabled, each annotation is rendered on an individual plane that passes through the annotation's 3D world position and is tangent to the sphere centered on the observer.

this is useful for rendering in VR.

CelestiaScreenshot


levinli303:
there is clipping issue for foreground annotations because of depth partitioning. Closing for now.

PR #2493, opened 2026-04-14
Support sRGB textures

levinli303:
No texture is read in sRGB format yet, so this PR doesn't change behavior.

ES3 + Needs to generate mipmap:
sRGBA is natively supported. Other sRGB formats are all expanded to sRGBA

ES3 + No need to generate mipmap:
sRGBA/sRGB are natively supported. Other sRGB formats are all expanded to sRGBA

ES2 with EXT_sRGB:
sRGBA/sRGB are natively supported. Other sRGB formats are all expanded to sRGBA
There is no mipmap support, `glTexParameteri(GL_TEXTURE_CUBE_MAP, GL_TEXTURE_MIN_FILTER, GL_LINEAR)` is used instead

ES2 with no EXT_sRGB:
sRGB formats are treated as the non-sRGB formats. In this case we should probably not try to create sRGB texture in code

### DIFF
--- a/src/celengine/texture.cpp
+++ b/src/celengine/texture.cpp
@@ -11,6 +11,7 @@
 #include <cassert>
 #include <cstdlib>
 #include <cmath>
+#include <vector>
 
 #include "glsupport.h"
 
@@ -26,6 +27,8 @@ using namespace celestia;
 using celestia::util::GetLogger;
 using celestia::engine::Image;
 using celestia::engine::PixelFormat;
+using celestia::engine::expandLuminanceToRGBA;
+using celestia::engine::expandLuminanceAlphaToRGBA;
 
 namespace
 {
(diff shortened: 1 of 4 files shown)


src/celengine/texture.cpp
```diff
@@ -59,8 +62,57 @@ GetTextureCaps()
     return texCaps;
 }
 
+#ifdef GL_ES
+bool
+needsToRGBAExpansion(PixelFormat format, bool needsMipmap)
```
ajtribick: `needsRGBAExpansion` would be a better name for this.

src/celimage/image.cpp
```diff
@@ -362,6 +363,34 @@ void Image::forceLinear()
     format = getLinearFormat(format);
 }
 
+std::vector<std::uint8_t>
```
ajtribick: Probably could use `std::unique_ptr<std::uint8_t[]>` here

ajtribick (approved):

PR #2507, opened 2026-04-30
Use bolometric luminosity instead of luminosity

AstroChara:
In astronomy, "luminosity" generally refers to bolometric luminosity, so this should be used instead of V-band luminosity.

### DIFF
--- a/src/celengine/stardbbuilder.cpp
+++ b/src/celengine/stardbbuilder.cpp
@@ -528,8 +528,7 @@ applyTemperatureBoloCorrection(const StarDatabaseBuilder::StcHeader& header,
                 // Astronomical Society of Canada, Vol 92. p36.
 
                 double logT = std::log10(static_cast<double>(*temperature)) - 4.0;
-                double bc = -8.499 * std::pow(logT, 4) + 13.421 * std::pow(logT, 3)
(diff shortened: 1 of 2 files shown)


src/celengine/stardbbuilder.cpp
```diff
@@ -528,8 +528,7 @@ applyTemperatureBoloCorrection(const StarDatabaseBuilder::StcHeader& header,
                 // Astronomical Society of Canada, Vol 92. p36.
 
                 double logT = std::log10(static_cast<double>(*temperature)) - 4.0;
-                double bc = -8.499 * std::pow(logT, 4) + 13.421 * std::pow(logT, 3)
-                            - 8.131 * logT * logT - 3.901 * logT - 0.438;
+                double bc = -0.438 + logT * (-3.901 + logT * (-8.131 + logT * (13.421 - logT * 8.499)));
```
375gnu, line 531: could you add a source for the formula?

ajtribick: It's already in the file, just outside the diff window:

https://github.com/CelestiaProject/Celestia/pull/2507/changes#diff-19a592201c2aa6d94e97a34128e9cb5f9ede2feff02dffc28547ba91f92d2222R524-R528

ajtribick (approved):

PR #2508, opened 2026-05-02
Add new celx commands

levinli303:
1. Adding rotation to observer:gotolocation
2. Adding observer:changeorbitdistance
3. Adding upframe to observer:gotodistance
4. Adding celestia:setfaintestxxx
7. Adding syncZoom to observer:setfov to support .cel's setfov behavior which doesn't update zoom
8. Fix celestia_playaudio for detecting the nopause argument

### DIFF
--- a/src/celscript/legacy/command.cpp
+++ b/src/celscript/legacy/command.cpp
@@ -362,34 +362,36 @@ void CommandChangeDistance::process(ExecutionEnvironment& env, double /*unused*/
 // Orbit command: rotate about the selected object
 
 CommandOrbit::CommandOrbit(double _duration, const Eigen::Vector3f& axis, float rate) :
-    TimedCommand(_duration),
-    spin(axis * rate)
+    TimedCommand(_duration)
 {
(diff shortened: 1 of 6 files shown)


levinli303:
Code duplication is expected.

ajtribick:
I'm not entirely sure that replicating 1-to-1 the way the .cel scripts work into .celx is the best way to go. It might be more useful to have an equivalent to JavaScript's requestAnimationFrame, where the user can supply a callback function for the next frame, rather than porting the cel commands.

levinli303:
> I'm not entirely sure that replicating 1-to-1 the way the .cel scripts work into .celx is the best way to go. It might be more useful to have an equivalent to JavaScript's requestAnimationFrame, where the user can supply a callback function for the next frame, rather than porting the cel commands.

while I think it seems like a pattern common enough to be supported directly in Celestia, I've changed these methods to be instantaneous for now. I can achieve the same behavior by registering a "tick" hook in celx script.

ajtribick (approved):

PR #2510, opened 2026-05-03
ShaderManager updates

ajtribick:
* Convert shadow maps to enum class
* Use `unordered_map` instead of `map` for properties-based shaders
* Use enum for static shaders to avoid text lookups
* Bugfix: `setCloudShadowForLight` unset the self-shadow flag rather than the cloud shadow flag.

### DIFF
--- a/src/celrender/galaxyrenderer.cpp
+++ b/src/celrender/galaxyrenderer.cpp
@@ -197,7 +197,7 @@ GalaxyRenderer::bindTextures()
 void
 GalaxyRenderer::renderGL2()
 {
-    CelestiaGLProgram *prog =  m_renderer.getShaderManager().getShader("galaxy");
+    CelestiaGLProgram *prog =  m_renderer.getShaderManager().getShader(StaticShader::Galaxy);
     if (prog == nullptr)
         return;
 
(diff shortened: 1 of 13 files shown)


levinli303 (approved):

src/celrender/galaxyrenderer.cpp
```diff
@@ -466,7 +466,7 @@ GalaxyRenderer::initializeGL3(const CelestiaGLProgram *prog)
         }
         else
         {
-            m_renderData.emplace_back(gl::Buffer(util::NoCreateT{}), gl::VertexObject(util::NoCreateT{}));
+            m_renderData.emplace_back(gl::Buffer{}, gl::VertexObject{});
```
ajtribick: accidental copy/paste from other PR

src/celrender/galaxyrenderer.cpp
```diff
@@ -341,7 +341,7 @@ GalaxyRenderer::initializeGL2(const CelestiaGLProgram *prog)
...
```
ajtribick: accidental copy/paste from other PR

PR #2512, opened 2026-05-04
miniaudio: share one ma_context/ma_engine across every audio channel

levinli303:
I've observed that on iOS (macOS is fine). If I do

1. playaudio on channel 0
2. playaudio on channel 1
3. playaudio on channel 1 with another audio file (which internally stopaudio on channel 1 first)

channel 0 stops playing. This fixes that by using a shared context and engine.


ajtribick (approved):

munix9:
Why not take this opportunity to update to the latest version of miniaudio?
The version currently in use, 0.11.17, is from 2023
The latest version is 0.11.25

PR #2517, opened 2026-05-08
Async CTX virtual texture tile loading

juliendms:
## Summary

Move CTX (virtual texture) tile loading off the render thread. Today, every cache miss in `VirtualTexture::getTile` synchronously reads a tile file, decodes it, and uploads it to GL inside the patch render loop -- so a pan that reveals a dozen new patches stalls the frame for a dozen load+upload roundtrips.

After this change:
- A single shared worker thread (`TileLoader`, in an anonymous namespace inside `virtualtex.cpp`) handles file I/O and image decode for all `VirtualTexture` instances in the process.
- The render thread performs bounded GL uploads (`kMaxUploadsPerFrame = 4` per `VirtualTexture` per frame) from a per-owner result queue at the start of `beginUsage()`.
- `Tile` gains a `State` enum (`NotLoaded` / `Queued` / `Ready` / `Failed`) that replaces the implicit `tex == nullptr` / `loadFailed` pair. State is read and written exclusively on the render thread; the worker treats `Tile*` as an opaque cookie that round-trips through the request/result pair unchanged.
- Cancellation is by id: each `VirtualTexture` registers a `uint64_t` with the loader on construction and deregisters on destruction. Worker results are delivered under the registry mutex, so a destroyed owner can never see a stale callback.

## Design notes

- **No public interface change.** `LODSphereMesh` is untouched. `getTile`, `beginUsage`, `endUsage`, `bind` signatures are unchanged.
- **No GL on the worker thread.** `Image::load` is the only worker-side call; `ImageTexture` construction is exclusively in `beginUsage`.
- **`getTile` becomes non-blocking.** On a miss it enqueues a request (idempotent via the state check) and returns the deepest already-`Ready` ancestor. The first frame after a tile becomes visible may show a coarser ancestor; once the worker delivers and `beginUsage` uploads, subsequent frames show the finer tile.
- **FIFO request queue, unbounded.** No priority, no drop policy in V1.
- **Owner registry under a mutex.** The worker pushes the result while holding the same mutex `unregisterOwner` takes, so result-queue lifetime and registry membership are kept consistent.

## User-visible effect

Smooth panning across high-LOD CTX surfaces. Freshly-revealed patches show a brief lower-LOD blur for a handful of frames instead of a per-pan freeze. Tested with the Moon with up to level 6 virtual texture diffuse and normal.

Before:
https://github.com/user-attachments/assets/1bac0270-fcc2-4f11-9001-f3614c4edb6c

After:
https://github.com/user-attachments/assets/f25577ed-e4f5-4386-92f6-d54ea6ac61b1

> **Disclaimer:** The code in this PR was generated by AI (Anthropic Claude).

### DIFF
--- a/src/celengine/virtualtex.cpp
+++ b/src/celengine/virtualtex.cpp
@@ -104,7 +104,7 @@ public:
     // unreachable in normal use.
     static std::atomic<bool>& shutdownFlag()
     {
-        static std::atomic<bool> flag{ false };
+        static std::atomic flag{ false };
         return flag;
     }
 
@@ -161,7 +161,7 @@ private:
         {
             TileRequest req;
             {
-                std::unique_lock<std::mutex> lock(requestsMutex);
+                std::unique_lock lock(requestsMutex);
                 requestsCv.wait(lock, [this] {
                     return exitRequested || !requests.empty();
                 });
--- a/src/celengine/virtualtex.cpp
+++ b/src/celengine/virtualtex.cpp
@@ -10,12 +10,20 @@
 
 #include "virtualtex.h"
 
+#include <atomic>
 #include <cassert>
+#include <condition_variable>
+#include <cstdint>
 #include <cstdio>
 #include <fstream>
+#include <mutex>
 #include <optional>
+#include <queue>
 #include <system_error>
+#include <thread>
+#include <unordered_map>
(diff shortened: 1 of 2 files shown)


levinli303:
superseded by #2593

PR #2520, opened 2026-05-10
Remove Synchronous Orbit

ajtribick:
As suggested in the code comments, `SynchronousOrbit` should be replaced with frame-based orbits.

The `LongLat` directive now overrides the `OrbitFrame` and uses a `FixedPosition` orbit. In order to support the defaulting of the rotation model, and keep consistency for sub-objects, the `FixedPosition` orbit now has an optional period.

This will cause observable changes in behaviour if the .ssc file uses a `LongLat` orbit in combination with a reference frame that isn't the default. It does fix the issue that `LongLat` does not work for class `surfacefeature`, since this always produces a FixedPosition in the BodyFixed frame.

### DIFF
--- a/src/celengine/parseobject.cpp
+++ b/src/celengine/parseobject.cpp
@@ -1594,25 +1594,6 @@ CreateOrbit(const Selection& centralObject,
         GetLogger()->error("Object has incorrect FixedPosition syntax.\n");
     }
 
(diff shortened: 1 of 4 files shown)


ajtribick:
Not doing the TODO that Sonar is complaining about as part of this.

AstroChara (approved):

PR #2523, opened 2026-05-14
Frame refactor

ajtribick:
Removes the center property from the `ReferenceFrame` object. This property was only used for orbit frames, removing it allows us to more aggressively share the reference frames. E.g. there is only one instance of the `J2000EclipticFrame`.

This enables a new syntax where the `Center` of a reference frame can be specified in the `OrbitFrame` block itself. Non-OrbitFrame uses of ReferenceFrames (`BodyFrame`, `ConstVector` frames in `TwoVector` frames) will ignore the Center and should now allow specifying without the `Center`.

For future optimization, i.e. not for this PR: convert reference frames to constant vector frames at startup.

While fixing this, I ran into the issue that several places in the code are using the name tree hierarchy instead of the frame tree hierarchy to determine object parents.

Annoyingly celx treats all frames as ObserverFrame (which allows for doing stuff like setting the observer frame to an object's orbit frame), too late to fix that now. For the purposes of celx, the center of a body frame is always taken to be the body itself.

- [x] Reimplement the circular references check, ensuring it works for timelines
- [x] Fix remaining places where name hierarchy parents (via `PlanetarySystem`) are used
- [x] More tests to verify the syntax still works (done locally by editing cassini.ssc, automation is somewhat tricky at the moment)
- [x] Earth works
- [x] Pluto-Charon barycenter works
- [x] Gamma Cephei Ab works

### DIFF
--- a/src/celengine/parseobject.cpp
+++ b/src/celengine/parseobject.cpp
@@ -879,63 +879,75 @@ Selection
 getFrameCenter(const Universe& universe, const AssociativeArray* frameData, const Selection& defaultCenter)
 {
     const std::string* centerName = frameData->getString("Center");
-    if (centerName == nullptr)
-    {
-        if (defaultCenter.empty())
-            GetLogger()->warn("No center specified for reference frame.\n");
+    if (!centerName)
         return defaultCenter;
-    }
 
     Selection centerObject = universe.findPath(*centerName, {});
     if (centerObject.empty())
     {
         GetLogger()->error("Center object '{}' of reference frame not found.\n", *centerName);
-        return Selection();
+        return defaultCenter;
     }
 
     // Should verify that center object is a star or planet, and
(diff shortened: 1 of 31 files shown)


src/celengine/parseobject.cpp
```diff
@@ -1101,79 +1119,88 @@ CreateFrameVector(const Universe& universe,
                 observer = center;
 
             if (observer.empty() || target.empty())
-                return nullptr;
+            {
+                GetLogger()->error("Missing Observer and Target in RelativeVelocity.\n");
```
levinli303, line 1123: should we return std::nullopt here

ajtribick: yes. good catch!

levinli303 (approved):

PR #2525, opened 2026-05-15
Allow Qt 6.10+ to use built-in Wayland pointer warp support

### LINKED ISSUE #2393: Rewrite Qt Wayland handling using wp_pointer_warp_v1
Now that Wayland has a [pointer warp protocol](https://wayland.app/protocols/pointer-warp-v1), it would be good to rewrite the Wayland handling in terms of that. This fits better with the built-in Qt mouse events model and how the pointer handling works on other platforms, and I think will get supported out of the box in Qt 6.10.

ajtribick:
Resolves #2393

Useful on Ubuntu 26.04.


munix9:
Does that mean that for distributions with Qt >= 6.10, `USE_WAYLAND_DRAG` should be set to `OFF`, and for those with Qt < 6.10, it should be set to `ON`?

ajtribick:
Generally yes. But it could potentially be useful with Qt 6.10+ if the compositor only supports the zwp_pointer_constraints protocol rather than the slightly newer wp_pointer_warp.

munix9:
> Generally yes. But it could potentially be useful with Qt 6.10+ if the compositor only supports the zwp_pointer_constraints protocol rather than the slightly newer wp_pointer_warp.

I see. That could get pretty tricky when building packages for the various distributions and versions.

ajtribick:
There are better ways to implement this.

Issue #2235, opened 2024-10-08
Use HarfBuzz for text shaping


ajtribick:
Currently we have a hand-written text shaping routine which assumes 1 codepoint -> 1 glyph. This may not always be the case. We should investigate using HarfBuzz to do text shaping instead. (This library is already used in GTK and Qt so should not bloat the dependencies too much)

Also worth investigating whether its worth caching the results of the text shaping to minimise the amount we need to do this per frame.

ajtribick:
Some notes on script segmentation are here: https://github.com/OpenType/opentype-layout/blob/master/docs/script_segmentation.md

PR #2527, opened 2026-05-18
Make default font paths localizable

levinli303:
The previous implementation that uses `LocaleFilename` cannot handle having a ttc index. It was bad for CJK fonts which usually comes in .ttc format to save space.

Move the default Font / TitleFont / LabelFont entries out of celestia.cfg.in and into LoadFontHelper, exposed as gettext keys (DEFAULT_MAIN_FONT, DEFAULT_TITLE_FONT, DEFAULT_LABEL_FONT) so translators can ship locale-appropriate defaults without users having to touch their config.

Also consolidate the "fontpath,index,size" parsing inside LoadTextureFont; index/size become std::optional so callers can either accept the parsed values or override them.

### DIFF
--- a/celestia.cfg.in
+++ b/celestia.cfg.in
@@ -238,9 +238,9 @@ StarTextures
 # TitleFont: Used to display object names, messages, and script text.
 #            Default "DejaVuSans-Bold.ttf,15"
 #------------------------------------------------------------------------
-  Font       "DejaVuSans.ttf,9"
-  LabelFont  "DejaVuSans.ttf,9"
-  TitleFont  "DejaVuSans-Bold.ttf,15"
+#  Font       "DejaVuSans.ttf,9"
+#  LabelFont  "DejaVuSans.ttf,9"
+#  TitleFont  "DejaVuSans-Bold.ttf,15"
 
 
 #------------------------------------------------------------------------
--- a/src/celestia/celestiacore.cpp
+++ b/src/celestia/celestiacore.cpp
@@ -2617,13 +2617,20 @@ LoadFontHelper(const Renderer* renderer,
                const char* defaultKey,
                const std::filesystem::path& defaultPath)
 {
-    std::filesystem::path fontPath = configPath;
-    if (fontPath.empty())
+    if (!configPath.empty())
     {
-        const char* translated = _(defaultKey);
-        fontPath = translated == defaultKey ? defaultPath : translated;
+        if (auto font = LoadTextureFont(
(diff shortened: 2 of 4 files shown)


levinli303:
Would love to get your ideas on this PR @ajtribick @375gnu 

Currently there is no way for "DejaVuSans.ttf,9" to map to a font of a specific index in a .ttc file with `LocaleFilename`

Here is the sample commit that shows after this PR what changes would be needed if we were to switch to Noto fonts to support all supported locales https://github.com/celestiamobile/Celestia/commit/0ef1663cf148366b3f2e88305e8194c86b6fc1ab

ajtribick:
Sorry for delay in response, had taken a look yesterday but wanted some time to think it over. I am wondering if this is a good approach - it seems like it would be awkward to manage if we did decide to change the bundled font.

Maybe it would be better instead to take a list of fonts to use as fallbacks, if glyph retrieval fails for the first font then try the next one, etc. (Possibly with a view to using fontconfig as a longer term solution?)

levinli303:
> Sorry for delay in response, had taken a look yesterday but wanted some time to think it over. I am wondering if this is a good approach - it seems like it would be awkward to manage if we did decide to change the bundled font.
> 
> Maybe it would be better instead to take a list of fonts to use as fallbacks, if glyph retrieval fails for the first font then try the next one, etc. (Possibly with a view to using fontconfig as a longer term solution?)

the font list fallback should be the right way to go ahead, but even with that the fallback font list will need to be locale dependent (for example CJK has glyphs sharing code point but looking different, Japanese variations should be higher on the fallback list than Korean for a Japanese user), so we would still need to store the font fallback list somewhere. And that would be either in .po or as a locale dependent config file Celestia reads using `LocaleFilename`

ajtribick:
Right, I think that then gets to the point where we need to consider a proper font-shaping library (i.e. harfbuzz, issue #2235) rather than the codepoint-based system we currently have. This then brings in the need to do bidi and script segmentation, so likely icu becomes a hard dependency as well.

Since that's a much larger change, I will approve this for now.

ajtribick (approved):

levinli303:
I think in the long term for font discovery + rasterization we might want to implement it per-platform with a fallback where platform solution is not available as I don't see a cross-platform lib handling it.

CoreText for Apple, DirectWrite for Windows, AFont + FreeType for Android (API 29+), and Fontconfig +  FreeType for Linux.

PR #2529, opened 2026-05-20
Update Qt Wayland drag handling

### LINKED ISSUE #2393: Rewrite Qt Wayland handling using wp_pointer_warp_v1

ajtribick:
- Attempt to use warping in Qt 6.10 if wp_pointer_warp is detected
- Provide fallback if Wayland protocol files are not supplied
- Don't attempt to destroy global Wayland objects

Supersedes #2525
Fixes #2393

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,7 +259,7 @@ jobs:
               -DENABLE_SDL=ON                                \
               -DENABLE_QT5=${{matrix.config.qt5}}            \
               -DENABLE_QT6=${{matrix.config.qt6}}            \
-              -DUSE_WAYLAND=${{matrix.config.wayland}}       \
+              -DUSE_WAYLAND_DRAG=${{matrix.config.wayland}}  \
               -DENABLE_FFMPEG=ON                             \
               -DENABLE_LIBAVIF=${{matrix.config.avif}}       \
               -DENABLE_MINIAUDIO=${{matrix.config.minaudio}} \
@@ -343,7 +343,7 @@ jobs:
               -DENABLE_SDL=ON                \
               -DENABLE_QT5=OFF               \
               -DENABLE_QT6=ON                \
-              -DUSE_WAYLAND=OFF              \
+              -DUSE_WAYLAND_DRAG=OFF         \
               -DENABLE_FFMPEG=ON             \
               -DENABLE_LIBAVIF=ON            \
               -DENABLE_MINIAUDIO=ON          \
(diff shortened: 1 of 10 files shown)


munix9:
I'm not entirely sure when `USE_WAYLAND_DRAG` should be used.

In a Linux distribution that offers `Qt >= 6.10.0`, would I use `-DUSE_WAYLAND_DRAG=OFF`?
That would cause all `HAS_WAYLAND_DRAG` checks in the code to be ignored.

But then I don’t understand, for example, lines 22/23 in `src/celestia/qt6/CMakeLists.txt`: If I use `-DUSE_WAYLAND_DRAG=OFF` because my system offers `Qt >= 6.10.0`, why is the version still checked against `Qt >= 6.10.0` in line 23?
https://github.com/CelestiaProject/Celestia/pull/2529/changes#diff-b3f09c31c4dccd2aac85bd7254a2e73fe6e6562f00566f9597c3dc08ed886905R22-R23

In this case, the check in line 23 would never apply, because I would only define `-DUSE_WAYLAND_DRAG=ON` for Qt < 6.10.0.

The situation is similar for lines 108–113 in `src/celestia/qt/qtdraghandler.cpp`
https://github.com/CelestiaProject/Celestia/pull/2529/changes#diff-048c3e1cf817a8ae5db20a83285130fab1488c4563d142d31b9009df77efa580R108-R113

Or should one simply always use `-DUSE_WAYLAND_DRAG=ON` for Linux (in which case, in my opinion, some of the checks are still odd)?

I’m a bit confused ;-)
More documentation on this would certainly be helpful.

ajtribick:
The situation is this:

wp_pointer_warp seems to be the official Wayland protocol to use when implementing the "infinite drag", and Qt 6.10+ supports it for setPos. Unfortunately it seems this is rather new, and many Wayland compositors don't currently support it.

zwp_pointer_constraints/zwp_relative_pointer are older Wayland protocols with wider support than wp_pointer_warp, but are not at all supported by Qt (which doesn't really have a concept of relative mouse motion events that this protocol would use).

If you have Qt 6.10+ and can rely on running on a compositor with wp_pointer_warp support, then the safe option is to use USE_WAYLAND_DRAG=OFF, that way you don't rely on the Wayland code in Celestia doing stuff correctly.

If you need to support compositors that only support the older unofficial protocols, or are using Qt <6.10 then you either need to switch USE_WAYLAND_DRAG=ON, or you will have to put up with drag being constrained by the screen edges on Wayland. On Qt 6.10+ we still prefer using wp_pointer_warp if it is detected, because that option allows us to stick to public Qt APIs and do a minimum of hacking around with libwayland.

If you are a Wayland hater who doesn't want to have any Wayland stuff on your system (presumably you have also compiled Qt with Wayland support being disabled), you should use USE_WAYLAND_DRAG=OFF.

ajtribick:
Regarding line 23, you only get there if you've previously set USE_WAYLAND_DRAG=ON, AND the system libwayland supplies any of the protocol definition files that are relevant for the implementation (which sets HAS_WAYLAND_DRAG). So I'm not sure why you think that check for the Qt version occurs if USE_WAYLAND_DRAG=OFF.

PR #2542, opened 2026-05-27
Add Point Spread Function (PSF) star renderer

### LINKED ISSUE #1948: Physically based star rendering
**What is the problem:**
- Celestia star styles don't model their actual difference in brightness physically accurately.
- This became even more noticeable in [the branch implementing gamma correction](https://github.com/CelestiaProject/Celestia/tree/srgb-test-v1):
screenshot by ajtribick

**What were ways to solve:**
The problem was already noticed and described by Chris Layrel, 2010 forum discussion is [here](https://celestiaproject.space/forum/viewtopic.php?f=10&t=16031). In a roundabout way, it led to [this branch](https://github.com/CelestiaProject/Celestia/tree/new-stars-v2). But, as onetwothree said, "it's buggy and slow". One of the problems was the squares around bright stars (Gaussians don't have a limit on the angular size, but the shader does):
screenshot by onetwothree

**What I suggest:**
I've developed a rendering algorithm that solves this problem. Unfortunately, I've virtually no knowledge of the C/C++, and have limited time. Therefore, I'm creating the issue to document the current developments, perhaps I'll return to this later.

The render below shows three columns of stars renderings of different magnitudes. The first is based on the Chris's formula given on the forum, and in fact consists of two independent Gaussians. The second is an implementation of the photopic point source function (PSF) from [this paper](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.41.1625&rep=rep1&type=pdf). The third column is my attempt to optimize the formula and create a constraint on the angular size (to fix the squares bug). The render is gamma corrected and enlarged three times without interpolation.
celestia_stars_v1

Python functions that implement optimized eye's PSF:
```py
from PIL import Image
import numpy as np

def PSF_fast(theta: int):
    if theta < 0.038:
        return 1 - 62.8*theta + (37.66*theta)**2 - (22.51*theta)**3
    else:
        return 1 / (35950 * theta**2 - 24) - 1.12e-4 # negative values mean zero

def draw_fast(img: Image.Image, br0: float, color0: np.ndarray, center: tuple[int, int], degree_per_px=0.05):
    color0 *= br0 / color0[1] # scaling on brightness and normalizing by green channel
    if np.all(color0 < 1):
        # single pixel render
        color = tuple((255 * gamma_correction(color0)).astype('int'))
        img.putpixel(xy=center, value=color)
    else:
        # glare render on 1°×1° field
        square_half_size = int(0.5 / degree_per_px) # 0.5° is an action radius of PSF_fast
        for x in range(-square_half_size, square_half_size+1):
            for y in range(-square_half_size, square_half_size+1):
                theta = np.sqrt(x*x + y*y) * degree_per_px
                color = color0 * PSF_fast(theta)
                color = (255 * gamma_correction(color)).astype('int')
                coords = tuple(np.array(center) + [x, y])
                color = tuple(img.getpixel(coords) + color)
                img.putpixel(xy=coords, value=color)
```

There are many ways to measure exposure, but it is convenient to measure it as a scale coefficient relative to a case when the entire flow of Vega (zero magnitude) still fits into one pixel. Fitting in a pixel is also what the `if` checks in `draw_fast()`. For example, an exposure of 1 means that Vega occupies a single pixel with no glow, and 5 means that this is true for a star 5 times dimmer than Vega, and Vega itself will have a noticeable glow. A general logic looks like this:

```py
exposure = 5 # variable
br_limit = 1 / (255 * 12.92 * exposure)
# Absolute limit of brightness, that wouldn't be displayed in 8 bit color spaces.
# The limit depends on the gamma correction definition, here sRGB is assumed.

linear_br = 10**(-0.4 * star_mag) # brightness measured in Vegas
if linear_br <= br_limit:
    print('Star is too dim to be displayed')
else:
    scaled_br = linear_br * exposure
    draw_fast(img, scaled_br, [r,g,b], (x,y))
```

**What are known problems:**
- Subpixel render will not work well, star is supposed to be strictly in the center. I assume it will lead to unexpected brightness fluctuations unless some kind of the PSF integration by pixel is implemented, and this is expensive.
- A bug with squares was fixed by adding a hard limit on the glare angular diameter of 1°. This also resulted in very large exposures being unpresentable compared to the original PSF function:
celestia_stars_bright

**Related suggestions:**
- Replace the faintest apparent magnitude with exposure.
- Not to change exposure when changing FOV.

levinli303:
Implements a new physically-based star rendering mode based on the algorithm by Askaniy Anpilogov (CelestiaStarRenderer), as a redesign of the stalled PR #1952 adapted to the current sRGB-aware codebase.

Adds a 4th StarStyle::PointSourceFunction alongside the existing FuzzyPointStars / PointStars / ScaledDiscStars styles. The renderer emits linear radiance with additive blending and combines two passes:

  * Point mode: a volume-conserving linear cone of fixed pixel radius, peakRadiance = exposure * 3 * irradiance / (pi * pointRadius^2)
  * Glow mode: an additive Spencer (1995) photopic eye-PSF approximation enabled when peak radiance exceeds 1.

Two user-tunable parameters drive the look:
  StarPointRadius (1..10, default 2)   - cone radius in points
  StarOptimization (0.1..1.0, default 0.1) - glow falloff

Plumbed end-to-end:
  * shaders/psfstar_{vert,frag}.glsl (GLES-safe: precision-qualified int uniform, guarded pow(), minimum gl_PointSize)
  * src/celengine/psfstarvertexbuffer.{h,cpp} + Renderer integration
  * StaticShader::PsfStar registration; uses canonical in_Intensity attribute binding for per-vertex peak radiance
  * celestia.cfg.in keys with clamping in configfile
  * Qt menu + preferences dialog (radio + two spinboxes, persisted)
  * SDL render dialog (4th radio button)
  * Win32 menu entry


levinli303:
This is mostly AI generated, I'll look into it more before publishing.

levinli303:
@Askaniy I think the appearance has been tuned well enough.
the only big blocker is that when within SolarSystemMaxDistance (default 1 ly), we switched back to the old star render mode, are we OK with an abrupt change, or how should we transition.


here is a video of changing to old star render mode when the old star disc size starts to exceed the psf disc size.

https://github.com/user-attachments/assets/d3fbef2d-442f-4738-91c6-2add2ca2702b

Askaniy:
Thanks! I'm going to compile and test the transition to old star render mode. Also want to check spacecrafts and planets from far distance.

Askaniy:
Compiled (Qt6), testing
1. The new star style is not added to the drop-down list "Stars as..." in the upper toolbar
2. Planets (and reflective light sources) are not connected to the rendering. Is it intentional or is it better to touch that code after replacing FaintestMag system with the exposure system?
3. Suggesting setting the default point radius value to 1.5 pt, it works without expected significant brightness fluctuations

Askaniy:
4. Testing on a 2018 laptop (Intel Core i7 8th gen), the new star rendering drops FPS of a static scene to 25 for the Milky Way plane direction in FOV. It seems to happen because of the number of stars, much higher than for other star styles (confirmed by manipulating peak radiance via point radius and exposure). Probably, at exposure 100 all the Celestia's star database is shown (FPS is recovering only with exposure lower than 50). "Optimization" have no effect on this. I can suggest drop the rendering peak radiance limit from 1/255 to 2/255 or 3/255, it would help a bit (a new setting?..).
5. I think it's better to restore the previous "Optimization" name, "PSF optimization" or "Bloom optimization", or something more clear ("Bloom compactness"?)
6. Tooltips could be added to "point radius" and "optimization", as was done for "max irradiance" and "exposure"

Askaniy:
7. I suggest changing the minimum allowed "optimization" value from 0.1 to 0.05, it still valid
8. It seems like "Max irradiance = 0.", which should remove the upper limit, is fully working only within `SolarSystemMaxDistance`. There's some unknown second upper limit taking place further than 1 ly.

> the only big blocker is that when within SolarSystemMaxDistance (default 1 ly), we switched back to the old star render mode, are we OK with an abrupt change, or how should we transition.

9. I suggest trying to link the bloom size to the object angular size and don't switch back to the old star render mode. Atmospheres should be removed from stars then (like it should be done for limb darkening). If overexposure just at the stellar limb doesn't look good, you can deliberately underestimate the star's angular radius.
Python pseudocode:
```py
peak_radiance = irradiance_to_peak_radiance(soft_clipped(light_source.irradiance()))
if light_source.resolved():
    linked_peak_radiance = ( light_source.angular_radius * (psf_a + 1 / psf_b) ) ** 2.5
    peak_radiance = max(peak_radiance, linked_peak_radiance)
light_source.render(peak_radiance)
```

Alternative suggestion by Pedro J.:

> if the transition between PSF and resolved star can't be dealt with in a satisfactory manner, then we might as well bring back the flare texture and render it instead of the glow when close up

Askaniy:
10. Max irradiance tooltip requires a fix: "Soft-clip on per-star peak radiance." → "Soft-clip on per-star peak irradiance."
11. If "max irradiance" is high, "optimization" somehow *increases* the brightness instead of only compacting the bloom. Possibly, there's something wrong.

Askaniy:
(I made a typo in item 10: not "peak irradiance", just "irradiance"...)

Also, copilot stated in the description

> Only emit the eye-PSF bloom for emissive sources, since the Spencer
> PSF models the eye's response to a bright unresolved star and would
> look wrong as a halo around a planet/moon.

Which is not true, planets/moons/etc should render the same way, I suppose? When a planet is resolved, the star-like rendering should be removed.
However, this check is actually necessary for the mode suggested in item 9, which has a "halo" effect for stars.

levinli303:
@Askaniy  saturn moons look like this when we show bloom for planet and moons

Screenshot 2026-06-02 at 11 28 56 AM

Askaniy:
I see an exposure inconsistency raised between Saturn and its moons, hopefully it could be fixed after proper exposure implementation
For now, user can lower the exposure manually, I think?
It would be great if exposure could be manipulated with `[` and `]` buttons like I have done in the old star rendering branch.

Pedro J.:
> Sweet, very SE-like

src/celestia/qt/preferences.ui
```diff
@@ -1103,6 +1103,10 @@
                  </item>
                  <item>
                   <widget class="QDoubleSpinBox" name="psfRadiusSpinBox">
+                   <property name="toolTip">
+                    <string>Radius of the unresolved star disc in points. Larger values make every star bigger and brighter.
```
Askaniy: > `Larger values make every star bigger and brighter.`

I suggest `Larger values make every star appear bigger while conserving the flux.` because stars do not get `brighter`

levinli303:
@Askaniy  I fixed all except for 9 and 11.

9 just feels like a bigger task that should be better discussed and implemented later. 11 you can check again in the latest.

Askaniy:
Ok, thanks! Compiling now.

Askaniy:
Item 11 was probably related to item 8, which has been fixed. I don't see it happening anymore 👍

But... I've noticed a strange flickering bug that happens at approximately 500 ly for Betelgeuse.
[20260602_234906.webm](https://github.com/user-attachments/assets/fc97eb51-07c5-4227-b608-6be95a3472f1)
I can't reproduce it for other stars. And coming back to Betelgeuse, I can't produce it again. Suggesting ignoring it.

And, as the last thing, what about setting the default exposure to 10 instead of 100? Starfield wouldn't be that spectacular, but it would always deliver around 60 FPS, even while looking at Milky Way and without plugging the laptop into a power source.
screenshot 60 FPS

levinli303:
@Askaniy  it's probably another star right in front of it in the viewing direction.

Askaniy (approved):

PR #2545, opened 2026-05-29
GLES 3.0/GL 3.3 as baseline

### LINKED ISSUE #2543: Require GL 3.3 / GLES 3.0
I think we have cleared the blockers for using a core OpenGL context (except for shaders), and most device nowadays should support higher GL versions now. It seems a good time to require a higher GL version?

I picked GLES 3.0 because iOS doesn't support ES 3.1/ES 3.2. With GL 3.1 we can get UBO in core.

levinli303:
1. Require GLES 3.0/GL 3.3
2. Update shaders and use GLSL layout qualifier
3. Remove a bunch of GLES3 and extension checks
4. Remove GL_LINE_SMOOTH
5. Handle LumAlpha/sLumAlpha/Alpha/Luminance/sLuminance on desktop GL because of removal in core 

- [x] Star, galaxy, spacecraft...
- [x] Orbit, label, hud, console
- [x] Image overlay
- [x] Desktop GL/GL ES (Desktop GL 4.1 on macOS, and GL ES 3.0 via ANGLE on macOS)
- [x] sRGB on/off
- [x] Lua's GL binding

Closes #2543


levinli303:
one perf issue is that switching to 3.3 core profile removes the wider line support, so the lines are almost always triangulated (smooth line/hidpi) and thus slower.

ajtribick (approved):

375gnu (approved):

PR #2549, opened 2026-05-30
Use PixelFormat::Luminance for generated glare/disc textures

ajtribick:
These appear to be using formulae for linear color space rather than sRGB, so do not need to be re-linearized.

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -302,7 +302,7 @@ static std::unique_ptr<Texture>
 BuildGaussianDiscTexture(unsigned int log2size)
 {
     unsigned int size = 1U << log2size;
-    Image img(PixelFormat::sLuminance, size, size, log2size + 1);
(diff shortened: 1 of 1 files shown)


levinli303:
It is much brighter and has a pretty bad banding issue when used as linear

Screenshot 2026-05-30 at 1 04 14 PM

Askaniy:
Artifacts like this happen when gamma correction is applied after the 8-bit 0-255 discretization.
The proper fix would be to render the scene with float values and gamma correct it before discretizing.
However, I'm not familiar enough with the implementation details to say how feasible this is.
(I also don't think brightness is much of an issue here)

levinli303:
> Artifacts like this happen when gamma correction is applied after the 8-bit 0-255 discretization. The proper fix would be to render the scene with float values and gamma correct it before discretizing. However, I'm not familiar enough with the implementation details to say how feasible this is. (I also don't think brightness is much of an issue here)

we are already rendering to floating point frame buffer before performing gamma.

PR #2561, opened 2026-06-04
Fix depth-partition flicker near body surface

levinli303:
Replace the `d > eradius` branch on the convex farZ with the horizon- tangent distance — continuous at `d == eradius` and tighter, so it no longer flips between branches under ULP noise. Drop the `radius/2000` nearZ floor, which used to hide the foreground at low altitudes. With an atmosphere, the shell wraps the camera and extends past the planet horizon, so use the back of the cloud-extended bounding sphere instead.

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -5079,17 +5079,18 @@ Renderer::removeInvisibleItems(const math::InfiniteFrustum &frustum)
         // Test the object's bounding sphere against the view frustum
         if (frustum.testSphere(center, cullRadius) != math::FrustumAspect::Outside)
         {
-            float nearZ = center.norm() - radius;
+            float d = center.norm();
+            float nearZ = d - radius;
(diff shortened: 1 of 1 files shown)


ajtribick (approved):

levinli303:
reversed Z projection might be a better way to solve all these depth issues.

375gnu:
you still need a fallback path as clip control is not necessary supported by all drivers, especially GLES ones.

levinli303:
> you still need a fallback path as clip control is not necessary supported by all drivers, especially GLES ones.

I made reversed Z an option in celestia.cfg in my another PR that also checks the extensions. yeah I tested on iOS and Mac they don't have the extensions. luckily ANGLE has the extension (D3D, Metal, and Vulkan), i'm thinking about moving my iOS/macOS codebases to using ANGLE (on Metal).

levinli303:
this PR is still needed. I'll merge this first.

PR #2564, opened 2026-06-05
Batch large stars into a single draw call

levinli303:
we were doing 1 large star per draw call, this was very inefficient and can be batched


levinli303:
my FPS went from around 30 to 110. @Askaniy 

With this setting Point radius: 1.5, optimization: 0.2, Max irradiance: 10000, Exposure: 10000.

src/celrender/largestarrenderer.cpp
```diff
@@ -1,95 +1,196 @@
...
+                           const Color           &color,
+                           float                  scalar)
+{
+    if (m_nStars < m_capacity)
+    {
+        std::array<unsigned char, 4> packedColor{};
+        color.get(packedColor.data());
+
+        StarVertex *out = &m_vertices[static_cast<std::size_t>(m_nStars) * 6];
+        for (std::size_t i = 0; i < 6; ++i)
```
ajtribick: Possible improvements:

* Use an index buffer instead of sending 6 full vertices
* Use instanced drawing

levinli303: switched to using an index buffer for now, since  instanced drawing is not in vertexobject yet.

Askaniy:
> my FPS went from around 30 to 110.
> 
> With this setting Point radius: 1.5, optimization: 0.2, Max irradiance: 10000, Exposure: 10000.

Testing [on the same laptop as star PSF](https://github.com/CelestiaProject/Celestia/pull/2542#issuecomment-4602024245).
With these settings, FPS is in range 20-50 (depends on the Milky Way direction).
With exposure=100 and compactness=0.1, FPS is in range 24-58.

I think it's great optimization for very bright stars, but it seems the bottleneck for integrated graphics is the number of stars... On any exposure higher than 10-50 the rendering depth of the scene covers a full database slice until the end

Askaniy (approved):

Askaniy:
Could dim stars be batched the same way? Or is there other way to optimize the rendering using the fact that they are all have the save box size and using the same formula?

levinli303:
> Could dim stars be batched the same way? Or is there other way to optimize the rendering using the fact that they are all have the save box size and using the same formula?

they are already batched.

PR #2567, opened 2026-06-06
Draw PSF glow in front of resolved bodies on overflow

### LINKED ISSUE #2560: Seamless transition to close-up stellar glow
The issue was raised in https://github.com/CelestiaProject/Celestia/pull/2542#issuecomment-4598703468
> the only big blocker is that when within SolarSystemMaxDistance (default 1 ly), we switched back to the old star render mode, are we OK with an abrupt change, or how should we transition.

I suggest linking the bloom size to the star's angular size instead.

This requires:
- removing stellar atmospheres ([proper limb darkening](https://github.com/CelestiaProject/Celestia/pull/2047) has the same requirement)
- removing all previous "large star" shaders

Python pseudocode:
```py
peak_radiance = irradiance_to_peak_radiance(soft_clipped(light_source.irradiance()))
if light_source.resolved():
    linked_peak_radiance = ( light_source.angular_radius * (psf_a + 1 / psf_b) ) ** 2.5
    peak_radiance = max(peak_radiance, linked_peak_radiance)
light_source.render(peak_radiance)
```

It may look inconsistent because of overexposure just at the stellar limb. To fix, we can deliberately underestimate the star's angular radius by a some factor (0.5-0.9?).

Alternative suggestion by Pedro J.:

> if the transition between PSF and resolved star can't be dealt with in a satisfactory manner, then we might as well bring back the flare texture and render it instead of the glow when close up

levinli303:
| body         | overflow (glowPeak > linkedGlowPeak) | result                                  |
   | ------------ | ------------------------------------ | --------------------------------------- |
   | emissive     | yes                                  | front of body, peak = `glowPeak`        |
   | emissive     | no (with fade out)                                   | body center, peak = `linkedGlowPeak`    |
   | non-emissive | yes                                  | front of body, peak = `glowPeak`        |
   | non-emissive | no                                   | **skipped**                             |

### DIFF
--- a/shaders/psfstarglow_frag.glsl
+++ b/shaders/psfstarglow_frag.glsl
@@ -19,6 +19,7 @@ uniform float psfB;
 uniform float pointScale;
 
 in vec3  v_color;
+in float v_alpha;
 in float v_peakRadiance;
 in float v_psfRadius;
 in float v_pointSize;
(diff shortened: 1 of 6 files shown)


Askaniy:
Made a small PR to the branch with suggestion. 
Additionally, the PSF glow output max radiance should be clipped by 1, but I'm not sure how implement this properly.

Askaniy (approved):

PR #2570, opened 2026-06-06
Soft-clip dim stars for PSF performance

### LINKED ISSUE #2559: Soft clipping dim stars to improve perfomance
The performance issue of the new star rendering (https://github.com/CelestiaProject/Celestia/pull/2542) can be solved by soft clipping dim stars. I don't have exact numbers right now, but stars have such distribution that clipping of ~10% by lowest brightness could cut the number roughly in half.

I suggest this to be a new user setting (named something like "Dim star clipping"?) with ~10 as a default value, in a range from 1 to 100 (subject to revision based on test results).

Improves [hard clipping I suggested earlier](https://github.com/CelestiaProject/Celestia/pull/2542#issuecomment-4602024245) and that was implemented in https://github.com/CelestiaProject/Celestia/pull/2542/changes/b4d68da9d9ed98bac41a2666a1d91f993dd09ac3

Python pseudocode:
```py
peak_radiance_gate_factor = 10
dimmed_peak_radiance = peak_radiance_gate_factor * sqrt(peak_radiance**2 / peak_radiance_gate_factor**2 - 1)
```
The equation is based on the canonical equation of hyperbola and asymptotically does not change bright stars behavior.

After implementation, default exposure can be set to higher values (to 50 or 100?).

levinli303:
#2559

### DIFF
--- a/celestia.cfg.in
+++ b/celestia.cfg.in
@@ -274,19 +274,26 @@ StarTextures
 # before the Spencer glow radius is computed.  Without it, very bright
 # stars (or high FaintestMagnitude values) produce a runaway bloom that
 # washes out the screen.  Valid range: 1.0 to 1.0e6.
-# Default: 100.0.  Set to 0 to disable the clamp entirely.
+# Default: 500.0.  Set to 0 to disable the clamp entirely.
 #
 # StarExposure (Point Spread Function star style only) is a per-star
 # brightness multiplier applied to the irradiance prior to peak
 # computation:  peak = StarExposure * 3 * irradiance / (pi * r^2)
 # Larger values brighten every star uniformly (and extend the visible
 # magnitude limit via the minPeak cutoff).  Valid range: 1e-3 to 1e6.
-# Default: 10.0.
+# Default: 20.0.
+#
+# StarDimClipFactor (Point Spread Function star style only) sets the
+# threshold below which very dim stars are soft-clipped to zero.  The
+# cutoff is StarDimClipFactor × the 1-code sRGB irradiance level, so
+# larger values suppress more near-invisible stars and reduce overdraw.
+# Valid range: 1.0 to 100.0. Default: 10.0.
 #------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


Askaniy:
Testing with exposure=100:
clipping = 1: 20-25 FPS (MW)
clipping = 10: 54-59 FPS

Visual difference is very small, it's getting significant for higher values like 50 or 100 but still looks about right

What about exposure=50 and clipping=10 as default setting? That's what TheLostProbe suggested on Discord.
Low-performance devices should be fine with it
(And lower the exposure back to 10 again if they are not)

Askaniy:
Milky Way test with `dim star clipping = 10`:
`exposure=10k`, `max irradiance=100k`: 22-32 FPS (plugged-in)
`exposure=100`, `max irradiance=1000`: 60 FPS (plugged-in)
`exposure=100`, `max irradiance=1000`: 48-60 FPS (plugged-out)
`exposure=50`, `max irradiance=1000`: 56-60 FPS (plugged-out)

levinli303:
what about clipping 10, exposure 50, max irradiance 100 as default.

Askaniy:
I recommend something closer to `clipping=10`, `exposure=10-20` and `max radiance=500`:
- `max radiance` could be higher because of optimized bloom, and it's important visually
- my laptop can't really be used for minimal requirements, it has i7 processor and less than 10 years old
- [smartphone test results from Discord showed just 2-3 FPS](https://discord.com/channels/558624707147071489/691413425494032474/1513494432173920327), and the lower the `exposure` the more devices Celestia could run without lagging

levinli303:
updated the default to 10/20/500 respectively. we shouldn't really worry too much about low end device because it is not the default star style for now and user can tweak the parameters. we should use what looks right.

I tested on my end with a iPhone 12 Pro Max (6 years old), it is rendering at 120 fps with exposure 50 + max irradiance 500 + sRGB.

Askaniy (approved):

PR #2571, opened 2026-06-07
Perform premultiply alpha at last

levinli303:
#2567 introduced a bug that non-white stars becomes partially transparent due to because we are using rgb / max(r,g,b) as the color. This PR performs color clamping first before premultiplying alpha for fade out.

Also removed a bunch of max/if checks.


levinli303:
Demo on zooming in on a non-white star.

https://github.com/user-attachments/assets/df448163-940b-4954-ab3a-c5c33c547c53

Askaniy (approved):

PR #2575, opened 2026-06-09
Add glInvalidateFramebuffer calls

levinli303:


src/celengine/framebuffer.cpp
```diff
@@ -218,30 +244,19 @@ FramebufferObject::generateFbo(unsigned int attachments)
...
-                glBindFramebuffer(GL_FRAMEBUFFER, oldFboId);
-                cleanup();
-                return;
-            }
+            glBindFramebuffer(GL_FRAMEBUFFER, oldFboId);
+            cleanup();
+            return;
         }
     }
-    else if (!useRenderbufferMSAA)
```
375gnu, line 241: are you sure, you haven't broken the logic here?

levinli303: the old logic was wrong but harmless. this path is no depth + no MSAA, we don't need to call ` glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, 0, 0)` to detach the depth texture, since we are creating a new FBO.

375gnu (approved):

PR #2576, opened 2026-06-09
Shrink PSF glow billboard to its visible radius

levinli303:
This should greatly improve performance of PSF because the fragment size has been reduced when GPU bound.

with my testing, with 10k max irradiance + 100k exposure. The FPS increased from 33 to 68.

### DIFF
--- a/shaders/psfstarglow_frag.glsl
+++ b/shaders/psfstarglow_frag.glsl
@@ -22,6 +22,7 @@ in vec3  v_color;
 in float v_alpha;
 in float v_peakRadiance;
 in float v_psfRadius;
+in float v_p04;
 
 void main(void)
 {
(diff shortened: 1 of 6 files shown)


Askaniy:
Performance improvement is noticeable 👍
But I still think the bottlenecks are different on high-end and low-and devices. On high-end it's bloom, on low-end is the number of stars (https://github.com/CelestiaProject/Celestia/pull/2570 helped, but maybe there are more clear way to optimize the point mode?)

Milky Way test:
- 27-40 FPS → 30-45 FPS (`exposure=100`, `max irradiance=1000`) — +12%
- 14-23 FPS → 24-31 FPS (`exposure=10k`, `max irradiance=100k`) — +50%

Askaniy (approved):

levinli303:
This one is for GPU bound. The other one is for CPU bound.

PR #2577, opened 2026-06-09
Render rings as a separate transparent pass entry

levinli303:
we used to render rings with depthMask disabled, so anything rendered after it will be rendered in front of the object. the right way to do it is mark rings as not opaque so it gets rendered at the end of depth partition.

### DIFF
--- a/src/celengine/render.cpp
+++ b/src/celengine/render.cpp
@@ -1331,6 +1331,15 @@ void Renderer::renderItem(const RenderListEntry& rle,
                      m);
         break;
 
+    case RenderListEntry::RenderableRingSystem:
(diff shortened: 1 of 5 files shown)


AstroChara (approved):
This PR is for fixing a depth sorting bug where rings placed around a dummy object does not render correctly, as reported on Discord.

[Bug example image](https://cdn.discordapp.com/attachments/1177190673095589939/1513738883106340976/RingDepthSorting.png?ex=6a2cc731&is=6a2b75b1&hm=81d2c2c79a96b68adb74c1ca9bb6dcf7d75880abef845397f1b41140f7bfedba&)

ajtribick (approved):

PR #2578, opened 2026-06-12
Render rings with BJ Jonsson radial profiles

levinli303:


AstroChara:
Context: I asked levin to make this PR so we could see how their ring rendering implementation works.

My current thought is that I would like to explore options for theory-based rendering of ring textures (for example, calculate backlighting from transparency), since this code hard-coded the Saturn rings' optical behaviour (the 139 degrees angle that the forward scattering texture is taken from), and also forward scattering and backlighting data are not available for the vast majority of ring systems and hence this code would probably be mostly useless beyond Saturn. This will probably be done on Discord.

PR #2579, opened 2026-06-12
Update CustomOrbit (and add CustomRotation) for Triton

pedro-fixingstuff:
Orbit is from [Emelyanov and Samorodov (2015)](https://academic.oup.com/mnras/article/454/2/2205/2892708), replacing the previous implementation which was never used for some reason.
Rotation is from the 2015 IAU WGCCRE report, with the original source apparently being [Davies et al. (1991)](https://www.rand.org/content/dam/rand/pubs/notes/2009/N3425.pdf).

**On accuracy:** I'm able to reproduce the Voyager 2 encounter with a closest approach distance of 39751 km, happening at 09:13 UTC of 1989 Aug 25. From [Wikipedia](https://en.wikipedia.org/wiki/Voyager_2#Mission_profile), the actual fly-by happened at 09:23, with a distance of 39800 km.

The CustomOrbit and CustomRotation are consistent with each other, with the only divergence being a small but persistent offset in longitude. One possible cause might be a definition of the prime meridian which doesn't exactly match the sub-Neptune point - I couldn't find anything about it in Davies et al. from a quick glance.

### DIFF
--- a/src/celephem/customorbit.cpp
+++ b/src/celephem/customorbit.cpp
@@ -2708,10 +2708,10 @@ CreateUranianSatelliteOrbit(int n)
                                                    uran_zeta_phi[n]);
 }
 
-/*! Orbit of Triton, from Seidelmann, _Explanatory Supplement to the
- *  Astronomical Almanac_ (1992), p.373-374. The position of Triton
- *  is calculated in Neptunocentric coordinates referred to the
- *  Earth equator/equinox of J2000.0.
+/*! Orbit of Triton, from Emelyanov and Samorodov (2015), MNRAS 454, p.2205-2215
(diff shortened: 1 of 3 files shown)


AstroChara (approved):

src/celephem/customorbit.cpp
```diff
@@ -2720,60 +2720,114 @@ class TritonOrbit : public CachingOrbit
...
     }
 
     double getBoundingRadius() const override
     {
-        return 354800 * BoundingRadiusSlack;
+        return 354700 * BoundingRadiusSlack;
     }
+
+private:
+    const std::array<double, 7> KI
```
ajtribick: These arrays could be `static constexpr`

ajtribick (approved):

PR #2583, opened 2026-06-14
Move InfoURL into a separate manager class

ajtribick:
Store the info URLs in one place. Main advantage is that we no longer need to create a custom StarDetails just to add an info URL.

### DIFF
--- a/src/celengine/location.h
+++ b/src/celengine/location.h
@@ -37,8 +37,6 @@ public:
     float getImportance() const;
     void setImportance(float);
 
-    const std::string& getInfoURL() const;
-
     bool isLabelColorOverridden() const { return overrideLabelColor; }
     void setLabelColorOverridden(bool _override) { overrideLabelColor = _override; }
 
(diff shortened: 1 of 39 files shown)


src/celengine/stardbbuilder.cpp
```diff
@@ -824,6 +826,12 @@ StarDatabaseBuilder::finish()
         UserCategory::addObject(star, category);
     }
 
+    for (auto [catalogNumber, infoUrl] : infoUrls)
```
levinli303: can this be `auto&` too?

src/celengine/location.h
```diff
@@ -37,8 +37,6 @@ class Location
     float getImportance() const;
     void setImportance(float);
 
-    const std::string& getInfoURL() const;
```
levinli303, line 40: the member variable has not been removed

ajtribick: fixed

src/celengine/stardbbuilder.cpp
ajtribick: fixed

levinli303 (approved):

src/celengine/stardbbuilder.h
```diff
@@ -100,9 +101,11 @@ class StarDatabaseBuilder
     // List of stars loaded from binary file, sorted by catalog number
     std::vector<Star*> binFileCatalogNumberIndex;
     // Catalog number -> star mapping for stars loaded from stc files
-    std::map<AstroCatalog::IndexNumber, Star*> stcFileCatalogNumberIndex;
-    std::map<AstroCatalog::IndexNumber, AstroCatalog::IndexNumber> barycenters;
-    std::multimap<AstroCatalog::IndexNumber, UserCategoryId> categories;
+    std::unordered_map<AstroCatalog::IndexNumber, Star*> stcFileCatalogNumberIndex;
```
375gnu: why?

ajtribick: whatever, reverted

Issue #833, opened 2020-08-28
(One more) way to improve albedo


Askaniy:
Albedo-related improvement was first raised by the Celestia Origin team, then on Discord and here (#624).
Last time (on Discord) I wrote that it would be nice to connect `Albedo` to `BondAlbedo` too, but… `BondAlbedo` is limited to 1, while `GeomAlbedo` is unlimited. This can create problems with calculating temperatures for bodies with strongly directional reflectivity. 

So I propose to discuss adding four albedo parameters:

1. `Albedo` (from 0 to ∞)
For backward compatibility only. It isn't possible to cover all use cases with one parameter. 
_If specified, interpreted as `GeomAlbedo` as the most likely option._

2. `GeomAlbedo` (from 0 to ∞)
The geometric albedo is the most common parameter and the most useless for us 😕
If simplified, it is the brightness of the full disk (in zero phase angle). However, just in this case, the reflectivity of many bodies increases significantly (due to the [opposition surge](https://en.wikipedia.org/wiki/Opposition_surge)). For this reason, it will be much more accurate to use `Reflectivity` instead.
_Sets the `BondAlbedo` and `Reflectivity` values if it is <1, otherwise as 1._
Because imprecise values are better than no values.

3. `BondAlbedo` (from 0 to 1)
Nothing changes here. Still used for reliable temperature calculations.

4. `Reflectivity` (from 0 to 1)
This albedo has no generally accepted name (for example, "Normal albedo" is used [here](https://www.researchgate.net/figure/LORRI-albedos-on-Pluto-Panel-A-shows-normal-albedo-across-Pluto-from-LORRI-panchromatic_fig12_298918241)). 
It can probably be called "visual spherical albedo" or "visual Bond albedo". Without going into detail, this is the visible average global reflectivity, just what is needed for points' brightness if phase curves are unknown.

Here is an illustration of the light curves from [this](https://arxiv.org/abs/1511.04415) paper:
![light_curves](https://user-images.githubusercontent.com/44873271/91576226-54591080-e950-11ea-9d8e-85192c7cfbc7.png)
You can see the brightness peak at zero phase angle (corresponding to the geometric albedo). 
Table with parameters from the same paper, except for the Bond albedo:
Celestial body | `GeomAlbedo` |  `BondAlbedo` | `Reflectivity`
-- | -- | -- | --
[...]

PR #2584, opened 2026-06-15
Rename Reflectivity to SphAlbedo; rework albedo substitutions

pedro-fixingstuff:
The addition of the `Reflectivity` parameter was proposed in #833, but the quantity it refers to is more often called "spherical albedo" (see [Robinson, 2025](https://arxiv.org/abs/2507.22258) as an example), being one of the suggested names in that issue. Here a shortened form is used, for consistency with `GeomAlbedo` and `BondAlbedo`.

A couple of additional assumptions for input albedo values are made:
- `SphAlbedo` is used as an stand-in for `BondAlbedo` if the latter isn't provided; the first is an wavelength-specific form of the latter (being frequently named _visual Bond albedo_), which is bolometric.
- `Albedo` (deprecated) is used for both `GeomAlbedo` and `BondAlbedo`, thus covering both possible use cases for it (setting the object brightness and temperature).

In a couple of instances, the `getReflectivity` method is replaced with `getGeomAlbedo`, as it's used to get the _maximum_ rather than average brightness, considering phase functions.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -109,7 +109,7 @@ msgstr ""
 
 #: ../src/celengine/solarsys.cpp:991
 #, c++-format
-msgid "Incorrect Reflectivity value: {}\n"
+msgid "Incorrect SphAlbedo value: {}\n"
 msgstr ""
 
 #: ../src/celengine/solarsys.cpp:999
--- a/po/be.po
+++ b/po/be.po
@@ -115,8 +115,8 @@ msgstr "Няправільнае значэньне GeomAlbedo: {}\n"
 
 #: ../src/celengine/solarsys.cpp:991
 #, c++-format
-msgid "Incorrect Reflectivity value: {}\n"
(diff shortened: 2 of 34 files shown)


ajtribick:
I'm not sure about the abbreviation, think this should either be `SphericalAlbedo` or maybe `SphAlbedo`

AstroChara:
I like both the renaming and the substitutions, and regarding Albedo parameter, this makes me think we should make the logic consistent with TrueColorTools.

> I'm not sure about the abbreviation, think this should either be `SphericalAlbedo` or maybe `SphAlbedo`

While I enjoy the consistent 4-letter abbreviation, if we have to change this, I prefer SphAlbedo over the former, which would be inconsistent with the name of the geometric albedo, GeomAlbedo.

pedro-fixingstuff:
Went with `SphAlbedo`.

ajtribick (approved):

PR #2593, opened 2026-06-20
Async texture, model loading and eviction of stale entries

### LINKED ISSUE #2447: Virtual texture tile unloading
The current virtual texture implementation never unloads tiles. We should implement a strategy for doing this.

levinli303:
Support for loading textures/models (mainly disk reads) on worker threads, and eviction on render thread.

### DIFF
--- a/src/celengine/CMakeLists.txt
+++ b/src/celengine/CMakeLists.txt
@@ -73,7 +73,8 @@ set(CELENGINE_SOURCES
   octree.h
   octreebuilder.h
   opencluster.cpp
-  opencluster.h  orbitsampler.h
+  opencluster.h
+  orbitsampler.h
   overlay.cpp
   overlay.h
   imageoverlay.cpp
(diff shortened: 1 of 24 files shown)


ajtribick:
Sorry for the delay in reviewing this, had been thinking about trying something based on [MoodyCamel::ConcurrentQueue](https://github.com/cameron314/concurrentqueue) (in vcpkg, also Ubuntu/Debian repositories as libconcurrentqueue-dev) but this seems like a good start. Wondering if we actually need the priority queue functionality here?

levinli303:
priority is not used for now yeah we can make it simply a std::queue.

levinli303:
@ajtribick  can I get a review on this? the number of lock operations should be minimum. I've put the code in beta version of mobile builds for a week (roughly 2000 sessions) and haven't seen crashes or user feedback so the code should be safe. if in the future we want to use a lock-free queue, it should be pretty straightforward to migrate.

ajtribick (approved):

ajtribick:
Sorry for the delay reviewing, heatwaves are NOT good for coding! Let's go with this, seems to be working on Windows also.

PR #2595, opened 2026-06-20
Drop Qt5 support

### LINKED ISSUE #2323: Qt5 end-of-life
Qt 5.15 has reached end-of-life and the Qt documentation website has removed the switch to toggle to Qt 5.15 documentation (although it seems that for now it can be accessed by manually modifying the ULRs). It may be worth considering if we want to continue to support the Qt5 build, or to simplify to be Qt6-only.

ajtribick:
Fixes #2323


375gnu (approved):
let's dig it

PR #2596, opened 2026-06-20
Fix some render issues

levinli303:
1. sR8 extension doesn't support mipmapping so we still need expansion in this case
2. revert some logic in #2577 so that only rings are rendered as a separate pass only when we are outside's rings innerRadius


375gnu (approved):

levinli303:
dropped

2. render transparent items (comet tails/rings) at the end of a depth partition (previously batched point star rendering is deferred and happens at last) 

since in 3 we are setting depthMask to false when we are outside the ring region, the ring will not occlude stars.

PR #2603, opened 2026-06-21
[ci] More fixes for external PRs

ajtribick:
Turns out some of the properties on the PR object are not set when coming from an external PR. Recommendation for getting the PR details is with the GitHub CLI, so let's try using that.

(Unfortunately there's no way to see whether this works until we merge it to main, but it seems to work in my own workflow test repository)

Also explicitly set workflow permissions (good practice)


ev3nvy:
Maybe also set `timeout-minutes` while you're at it? I try to add it wherever I can on personal projects, so that a potential rougue job doesn't eat up a significant portion of my usage limits.

ajtribick:
Thanks for the tip! The default 6 hours is quite long, so I think some reductions are reasonable. Going with some fairly generous limits for now, can do some more observations to see if we can reduce them a bit further.

(In particular, I don't have much of an idea of how long the brew cache on MacOS can take if there are no cached package versions)

375gnu (approved):

PR #2606, opened 2026-06-21
Improved Modify handling for textures

ajtribick:
Do not reset a texture if the corresponding key is not present in the catalog file. Instead use an empty string to reset texture paths.

Alternative to #2489


levinli303 (dismissed):

ajtribick:
Thinking about it further, I'm not entirely sure about the approach for BumpHeight. For backwards compatibility, specifying `BumpMap` without `BumpHeight` should probably always reset to the default bump height value.

I am not sure a modify operation specifying just `BumpHeight` is actually useful, probably best to ignore it.

levinli303 (approved):

PR #2613, opened 2026-06-26
Atmosphere phase function fixes

pedro-fixingstuff:
The Mie phase function had an error in that the angle theta fed to it was the phase angle rather than the angle between incoming and outgoing directions (these angles aren't the same, though comlementary to each other). Here it is fixed by inverting a couple of signs in the equation. This means that the MieAsymmetry .ssc parameter will work the opposite way it did before - unfortunate, but it should now match the definition of the phase parameter g (for the Henyey-Greenstein function)/k (Schlick).

The Rayleigh function was previously commented out in favor of isotropic scattering; now it is used. It is multiplied by a factor (taken from the scattersim source code) so that the average value is close to 1, matching the isotropic approach.

### DIFF
--- a/src/celengine/solarsys.cpp
+++ b/src/celengine/solarsys.cpp
@@ -729,8 +729,15 @@ void ReadAtmosphere(Body* body,
         atmosphere->mieCoeff = *mieCoeff;
     if (auto mieScaleHeight = atmosData.getLength<float>("MieScaleHeight"))
         atmosphere->mieScaleHeight = *mieScaleHeight;
-    if (auto miePhaseAsymmetry = atmosData.getNumber<float>("MieAsymmetry"); miePhaseAsymmetry.has_value())
+    if (auto miePhaseAsymmetry = atmosData.getNumber<float>("MieHGAnisotropy"); miePhaseAsymmetry.has_value())
+    {
         atmosphere->miePhaseAsymmetry = *miePhaseAsymmetry;
+    }
+    else if (auto miePhaseAsymmetry = atmosData.getNumber<float>("MieAsymmetry"); miePhaseAsymmetry.has_value())
+    {
+        GetLogger()->warn("Deprecated parameter MieAsymmetry used in {} definition.\nUse MieHGAnisotropy instead.\n", body->getName());
+        atmosphere->miePhaseAsymmetry = -*miePhaseAsymmetry; // invert sign to address established usage
+    }
     if (auto rayleighCoeff = atmosData.getVector3<float>("Rayleigh"); rayleighCoeff.has_value())
         atmosphere->rayleighCoeff = *rayleighCoeff;
     //atmosData->getNumber("RayleighScaleHeight", atmosphere->rayleighScaleHeight);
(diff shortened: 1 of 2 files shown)


AstroChara (approved):
Pedro have tested these changes and reported them in Discord, and they appear to function correctly.

For Rayleigh scattering, I have come across [NASA Planetary Data System page](https://pds-atmospheres.nmsu.edu/education_and_outreach/encyclopedia/rayleigh_phase.htm) which also uses 3/4 as a factor for Rayleigh scattering, so I think including it is a good decision. We can also delete the Rayleigh factor comment as well.

I believe we should try to determine a way to test for Mie scattering. Maybe checking the O'Neal implementation from https://github.com/ebruneton/clear-sky-models (also see its paper: https://arxiv.org/abs/1612.04336) would help?

For now, these should be good.

levinli303:
this breaks add-ons which are not what we can control, i'd rather not change mieK to -mieK to keep compatibility.

levinli303:
we could introduce a new field in ssc like `MieScatterDirection` to influence how `MieAsymmetry` is read from the file. if `MieScatterDirection` is absent, we keep the compatibility by reading it as `-MieAsymmetry`. In this way both correctness and compatibility can be kept.

AstroChara:
After a discussion on Discord, we've reached an agreement to introduce a more precisely-named parameter, `MieHGAnisotropy` ("anisotropy" being a slightly more common word to refer to this specific Mie scattering behaviour than "asymmetry" according to Google Scholar), to refer specifically to the Henyey-Greenstein phase function's anisotropy parameter _g_. `MieAsymmetry` is to be understood as Celestia's deprecated legacy parameter.

Internally, `MieHGAnisotropy` will become the default parameter, and `MieAsymmetry` will be multiplied by -1 for use with the corrected Mie phase function currently present in the PR.

Pedro has stated they will try to add this parameter; I will implement this change in _CelestiaContent_ once they have implemented it here.

src/celengine/solarsys.cpp
```diff
@@ -730,6 +730,8 @@ void ReadAtmosphere(Body* body,
     if (auto mieScaleHeight = atmosData.getLength<float>("MieScaleHeight"))
         atmosphere->mieScaleHeight = *mieScaleHeight;
     if (auto miePhaseAsymmetry = atmosData.getNumber<float>("MieAsymmetry"); miePhaseAsymmetry.has_value())
+        atmosphere->miePhaseAsymmetry = -*miePhaseAsymmetry; // invert sign to address established usage
+    if (auto miePhaseAsymmetry = atmosData.getNumber<float>("MieHGAnisotropy"); miePhaseAsymmetry.has_value())
```
levinli303, line 732: we can move `if (auto miePhaseAsymmetry = atmosData.getNumber<float>("MieAsymmetry"); miePhaseAsymmetry.has_value())` case into an `else if` of checking MieHGAnisotropy.

and add warning log about the deprecation using `GetLogger()->warn()`.

levinli303 (changes requested):

src/celengine/solarsys.cpp
pedro-fixingstuff: done

AstroChara (approved):
t

levinli303 (approved):

PR #2617, opened 2026-06-28
Fix two crashes

levinli303:
Add a non-throwing celestia::util::PathToString() helper and use it everywhere a path was converted to a narrow string, since path::string() throws on Windows for paths not representable in the active code page.

Replace the assert(star != nullptr) in StarDatabaseBuilder::createOrUpdateStar with an stcError and return false, so a star with no position and no existing record logs an error instead of crashing.

### DIFF
--- a/src/celttf/truetypefont.cpp
+++ b/src/celttf/truetypefont.cpp
@@ -18,6 +18,8 @@
 #include <unordered_map>
 #include <vector>
 
+#include <boost/container_hash/hash.hpp>
+
 #include <celcompat/charconv.h>
 #include <celengine/glsupport.h>
 #include <celengine/render.h>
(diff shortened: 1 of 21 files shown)


src/celutil/fsutils.cpp
```diff
@@ -98,6 +98,16 @@ U8FileName(std::string_view source, bool allowWildcardExtension)
     return std::filesystem::u8path(source);
 }
 
+std::string
+PathToUTF8(const std::filesystem::path &path)
+{
+#ifdef _WIN32
+    return WideToUTF8(path.native());
+#else
+    return path.native();
```
375gnu, line 107: This returns utf8 only in utf8 locate. Please rename `PathToUTF8` to `PathToString`

375gnu (changes requested):

src/celutil/fsutils.cpp
levinli303: fixed.

src/celttf/truetypefont.cpp
```diff
@@ -772,7 +773,7 @@ template<> struct std::hash<FontCacheKey>
 {
     std::size_t operator()(const FontCacheKey &k) const
     {
-        return std::hash<std::string>()(k.filename.string()) ^ std::hash<int>()(k.index) ^ std::hash<int>()(k.size);
+        return std::hash<std::string>()(celestia::util::PathToString(k.filename)) ^ std::hash<int>()(k.index) ^ std::hash<int>()(k.size);
```
ajtribick: I think we should use `boost::hash_combine` here

ajtribick (approved):

375gnu (approved):

PR #2626, opened 2026-07-13
Add support for HDR tone mapping

levinli303:
Took the simple shader from https://learnopengl.com/Advanced-Lighting/HDR . Tonemapping is needed for proper atmosphere rendering.

PSF is also changed accordingly

- Removed the per-fragment brightness clamp so star radiance stays uncapped (HDR).
- Reshaped the glow interior into a flat-topped plateau instead of a center peak.
- Faded the interior exponentially and floored it at the limb brightness so the disc collapses to the limb as it resolves (when alpha = 0.5, the two values are equal).

psf_real_fadeexp

### DIFF
--- a/celestia.cfg.in
+++ b/celestia.cfg.in
@@ -533,4 +533,14 @@ StarTextures
 #------------------------------------------------------------------------
 # SRGBRendering true
 
+#------------------------------------------------------------------------
+# Exposure controls the exponential tone-mapping operator
+# (1 - exp(-Exposure * L)) applied to the linear HDR scene during the
+# sRGB output post-process.  It softly rolls off highlights above 1.0
+# instead of hard-clipping them.  Only takes effect when SRGBRendering
+# is enabled.
+# The default value is 1.0.
+#------------------------------------------------------------------------
+# Exposure 1.0
+
 }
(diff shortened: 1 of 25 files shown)


AstroChara:
Should this be turned on by default? It makes sunsets much less blinding, so might be a good addition, as long as it's not too harsh on performance (and I'm not noticing an indication of that) I suppose?

levinli303:
transition of PSF to star mesh is slightly changed. The actual fade (alpha) of PSF glow first decreases exponentially and then hit 0.5 and then starts decrease linearly.
https://github.com/user-attachments/assets/629b6f7c-3b1a-4744-be83-3c2c04c76c3e

levinli303:
> Should this be turned on by default? It makes sunsets much less blinding, so might be a good addition, as long as it's not too harsh on performance (and I'm not noticing an indication of that) I suppose?

Perf difference should be negligible. If it is default on we also need to decide on the default exposure value. from what I observe 1.0 is a good value.

AstroChara:
> Perf difference should be negligible. If it is default on we also need to decide on the default exposure value. from what I observe 1.0 is a good value.

Sounds reasonable. I wonder if this counts as initial implementation of the exposure system Askaniy had intended to implement for star PSF.

AstroChara (approved):
Discussions with Askaniy suggests that the issue with star PSF would require changes to the PSF itself and hence is probably not this PR's problem. If this issue isn't sufficiently problematic to block the merger, then I'm not aware of any other issue and am approving this.

ajtribick (approved):

PR #2629, opened 2026-07-15
Pass PSF glow fade alpha as a float instead of packing it into color

levinli303:
The glow fade alpha was baked into the 8-bit alpha channel of the vertex color, which limits the precision. This lifts it out to be passed separately as float.

### DIFF
--- a/src/celengine/psfstarvertexbuffer.cpp
+++ b/src/celengine/psfstarvertexbuffer.cpp
@@ -154,10 +154,20 @@ PsfStarVertexBuffer::setupVertexArrayObject()
         sizeof(StarVertex),
         offsetof(StarVertex, peakRadiance));
 
+    // continuous distance-derived glow fade; float for a smooth transition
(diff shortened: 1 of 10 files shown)


src/celengine/psfstarvertexbuffer.cpp
```diff
@@ -154,10 +154,20 @@ PsfStarVertexBuffer::setupVertexArrayObject()
         sizeof(StarVertex),
         offsetof(StarVertex, peakRadiance));
 
+    // continuous distance-derived glow fade; float for a smooth transition
+    m_vo->addVertexBuffer(
+        *m_bo,
+        CelestiaGLProgram::TextureCoord1AttributeIndex,
```
375gnu: could you add a new enum for this CelestiaGLProgram::AlphaAttributeIndex?

375gnu (changes requested):

375gnu (approved):

PR #2632, opened 2026-07-17
Fix atmosphere shell rendering

### LINKED ISSUE #2615: Atmo scattering changing in unexpected way with AtmosphereExtinctionThreshold
In #2200, Pedro and I discovered that changing `AtmosphereExtinctionThreshold` ([src/celengine/atmosphere.h](https://github.com/CelestiaProject/Celestia/blob/af775a7de04c0b88b9767fddeb9d60539deb202c/src/celengine/atmosphere.h#L53)) resulted in [Earth's atmosphere becoming visually thinner](https://discord.com/channels/558624707147071489/691413425494032474/1249504464185069662).

This behaviour is unexpected, given that what we did was reducing the AtmosphereExtinctionThreshold, which causes more of the atmosphere to be rendered, and hence the scattering should increase, rather than decrease.

Interestingly, the bug does not seem to significantly affect Titan, the body which the PR was initially tested on.

Additionally, in #2613, Pedro and I noticed that incorporating the 1/4pi factor into the Rayleigh phase function resulted in the atmosphere becoming visually too thin, and I believe this incorrect behaviour may be related to this bug.

levinli303:
1. use the planet radius as the atmosphere density origin
2. transform the view frustum into the scaled shell’s coordinate space
3. use Chapman function (approximation) for atmosphere optical depth
4. integrate atmospheric scattering along view rays

### DIFF
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -748,6 +748,7 @@ AtmosphericEffects(const ShaderProperties& props)
     //source += "    atmSamplePoint = atmEnter * 0.9 + atmLeave * 0.1;\n";
     source += "    h = max(0.0, length(atmSamplePoint) - atmosphereRadius.z);\n";
     source += "    density += exp(-h * mieH);\n";
+    source += "    density *= 0.5;\n";
 #endif
 
     bool hasAbsorption = true;
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -710,6 +710,35 @@ ScatteringPhaseFunctions(const ShaderProperties& /*unused*/)
 
 
 std::string
+ChapmanOpticalDepth()
+{
+    return R"glsl(float chapmanToSpace(vec3 origin, vec3 direction)
+{
+    float r = length(origin);
+    float mu = clamp(dot(origin, direction) / r, -1.0, 1.0);
+    float X = r * mieH;
+    float density = exp(min((atmosphereRadius.z - r) * mieH, 0.0));
+    float grazing = sqrt(1.5707963267948966 * X);
+    if (mu >= 0.0)
+        return (density / mieH) * grazing / ((grazing - 1.0) * mu + 1.0);
+    float sinZenith = sqrt(max(1.0e-8, 1.0 - mu * mu));
+    float tangentDensity = exp(min((atmosphereRadius.z - r * sinZenith) * mieH, 40.0));
+    float upward = (density / mieH) * grazing / ((grazing - 1.0) * (-mu) + 1.0);
+    return max(0.0, (2.0 * tangentDensity / mieH) * sqrt(1.5707963267948966 * X * sinZenith) - upward);
+}
+
+float chapmanOpticalDepth(vec3 origin, vec3 direction, float pathLength)
+{
+    vec3 endpoint = origin + direction * pathLength;
+    if (dot(endpoint, endpoint) < dot(origin, origin))
+        return max(0.0, chapmanToSpace(endpoint, -direction) - chapmanToSpace(origin, -direction));
+    return max(0.0, chapmanToSpace(origin, direction) - chapmanToSpace(endpoint, direction));
+}
+)glsl";
+}
+
+
+std::string
 AtmosphericEffects(const ShaderProperties& props)
 {
     std::string source;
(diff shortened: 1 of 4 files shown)


AstroChara:
Hmm, what's the effect of this PR?

levinli303:
they are correctness fixes.
1. AtmosphereRenderer::render takes in planet radius not planet radius + atmosphere height https://github.com/CelestiaProject/Celestia/blob/9b218d214dee7cf1b53d34864db7aa024501faf9/src/celrender/atmosphererenderer.cpp#L372 atmosphere should look thinner with the PR
2. the 1/3 and 2/3 samples were added but not averaged for calculating density previously. Excessive extinction and scattering should be reduced with the PR.
3. should use atmosphere frustum to do the culling for atmosphere not the planet frustum

AstroChara:
> they are correctness fixes.
> 
> 1. AtmosphereRenderer::render takes in planet radius not planet radius + atmosphere height https://github.com/CelestiaProject/Celestia/blob/9b218d214dee7cf1b53d34864db7aa024501faf9/src/celrender/atmosphererenderer.cpp#L372
>     atmosphere should look thinner with the PR
> 2. the 1/3 and 2/3 samples were added but not averaged for calculating density previously. Excessive extinction and scattering should be reduced with the PR.
> 3. should use atmosphere frustum to do the culling for atmosphere not the planet frustum

I see, thank you for the response.

I am mainly wondering if this PR fixes the bugs with atmosphere rendering that I/we have identified recently. I've gathered them [here](https://docs.google.com/document/d/1fOJAy2CKM7unZbykHdYpuucrmhJzBnwGdlG5sTNxCws/edit?usp=sharing). Mainly I'm concerned about the first one — what happens when the AtmosphereExtinctionThreshold is adjusted with this new PR? Does it behave correctly (becoming thicker as the threshold goes down)?

Though, the atmosphere becoming thinner with this new PR is somewhat odd and concerning, since currently the atmosphere already appears to be too thin, which led us to leave out the 1/4pi factor from the Atmosphere phase function fixes PR (#2613). Hmm...

levinli303:
AtmosphereExtinctionThreshold = 0.05

Screenshot 2026-07-17 at 9 23 38 PM


AtmosphereExtinctionThreshold = 0.1

Screenshot 2026-07-17 at 9 35 27 PM

AstroChara:
Pedro was testing this on Discord and found that reducing atmosphere extinction threshold (should result in a taller atmosphere, and hence more extinction AKA “thicker atmosphere”) still results in less extinction, so it seems that bug hasn’t been resolved.

As I mentioned on Discord, changes to the atmosphere rendering would necessitate changes to atmosphere definitions in CelestiaContent, so I’m hoping that we could fix all these bugs before merging all of them in a single PR, which would save us a lot of trouble..

levinli303:
after using Chapman function:

0.05 threshold
Chapman 0 05

0.00001 threshold
Chapman 0 00001

src/celengine/shadermanager.cpp
```diff
@@ -709,6 +709,38 @@ ScatteringPhaseFunctions(const ShaderProperties& /*unused*/)
 }
 
 
+std::string
+ChapmanOpticalDepth()
+{
+    std::string source;
+
+    source += "float chapmanToSpace(vec3 origin, vec3 direction)\n";
+    source += "{\n";
```
ajtribick: Instead of repeated concatenation, it would be better to either use combined string literals to do most of this at compile time:
```c++
source += "float chapmanToSpace(vec3 origin, vec3 direction)\n"
          "{\n"
          "    float r = length(origin);\n"
          // ...
          "}\n";
```

or use a single raw multiline string literal

ajtribick (approved):

AstroChara:
The current rendering of the **atmosphere in front of the planet disc** seems to be incorrect. The atmosphere's brightness barely, if at all, dropped past the terminator all the way to around the point where the uppermost rendered atmosphere is finally blocked (point "A" in the diagram). Only then did the atmosphere's brightness finally crashes. (See the upper plot)

Given that the planet's shadow blocks the brighter lower atmosphere first, I think the correct behaviour is a rapid drop in brightness near the terminator. (The bottom plot)

Celestia Notes-1

levinli303:
changed to use 6 sample points along the ray instead of just the mid point. perf might be bad

threshold 0.05
Screenshot 2026-07-18 at 11 18 05 AM

threshold 0.00001
Screenshot 2026-07-18 at 11 17 20 AM

threshold 0.05
Screenshot 2026-07-18 at 11 18 14 AM

threshold 0.00001
Screenshot 2026-07-18 at 11 17 31 AM

AstroChara:
Lovely, this solves the problem I mentioned. Can't comment on performance though.

PR #2637, opened 2026-07-21
Composite atmosphere transmission per color channel

levinli303:
Correct atmosphere compositing to preserve RGB transmittance instead of collapsing extinction to scalar alpha.

Uses dual-source blending for single-draw compositing when supported, with a portable two-pass fallback for unsupported devices.


levinli303:
before:

Screenshot 2026-07-21 at 3 27 36 PM

after:

Screenshot 2026-07-21 at 3 29 21 PM

375gnu (approved):

AstroChara (approved):

PR #2655, opened 2026-07-29
Render rings and atmosphere separately from the object and split ring rendering when needed

levinli303:
1. Move non-opaque items (like atmosphere and rings) to be rendered at the end of the depth partition
2. Split ring rendering into far and near rings when the atmosphere is visible (far ring -> atmosphere -> near ring)

### DIFF
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -1911,39 +1911,21 @@ buildRingsFragmentShader(const ShaderProperties& props)
 
     source += "\nvoid main(void)\n{\n";
 
-    // Capture the texture gradients up front, before the ring split branch
-    // below can diverge control flow. Sampling with these gradients keeps the
-    // mip selection identical on both sides of the split so the near and far
-    // halves meet without a seam.
-    if (util::is_set(props.texUsage, TexUsage::DiffuseTexture))
-    {
-        source += "vec2 ringTexDx = dFdx(diffTexCoord.st);\n";
-        source += "vec2 ringTexDy = dFdy(diffTexCoord.st);\n";
-    }
-
-    // When splitting the ring around the atmosphere, classify each fragment by
-    // whether the view ray reaches the atmosphere shell before the fragment:
-    // if so the fragment is behind the atmosphere and belongs to the far half,
-    // otherwise the near half. This follows the curved atmosphere limb exactly,
-    // unlike a flat split through the planet center. ringHalf > 0 keeps the near
-    // half, < 0 the far half, 0 the whole ring. The dropped half is made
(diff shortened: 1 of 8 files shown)


levinli303:
before:
Screenshot 2026-07-29 at 2 10 46 PM

after:
Screenshot 2026-07-29 at 2 12 53 PM

src/celrender/rendererfwd.h
```diff
@@ -15,6 +15,7 @@ class NebulaRenderer;
 class OpenClusterRenderer;
 class PsfGlowLargeRenderer;
 class ReferenceMarkRenderer;
+enum class RingRenderHalf;
```
ajtribick: I don't think this is valid as a forward declaration: you can forward declare an enum but you need to specify the underlying type (and match it where the enum is declared)

Probably can use `int` as the underlying type in this case.

src/celengine/shadermanager.cpp
```diff
@@ -1909,6 +1911,23 @@ buildRingsFragmentShader(const ShaderProperties& props)
...
+    source += "bool dropFragment = false;\n";
+    source += "if (ringHalf != 0.0)\n{\n";
+    source += "    vec3 ringDir = position - eyePosition;\n";
+    source += "    float a = dot(ringDir, ringDir);\n";
+    source += "    float b = dot(eyePosition, ringDir);\n";
+    source += "    float c = dot(eyePosition, eyePosition) - ringAtmosphereRadius * ringAtmosphereRadius;\n";
+    source += "    bool behind = b < 0.0 && a + b > 0.0 && b * b >= a * c;\n";
+    source += "    dropFragment = (ringHalf > 0.0 && behind) || (ringHalf < 0.0 && !behind);\n";
+    source += "}\n";
     source += "vec4 diff = vec4(ambientColor, 1.0);\n";
```
ajtribick:
```suggestion
    source += "bool dropFragment = false;\n"
              "if (ringHalf != 0.0)\n{\n"
              "    vec3 ringDir = position - eyePosition;\n"
              "    float a = dot(ringDir, ringDir);\n"
              "    float b = dot(eyePosition, ringDir);\n"
              "    float c = dot(eyePosition, eyePosition) - ringAtmosphereRadius * ringAtmosphereRadius;\n"
              "    bool behind = b < 0.0 && a + b > 0.0 && b * b >= a * c;\n"
              "    dropFragment = (ringHalf > 0.0 && behind) || (ringHalf < 0.0 && !behind);\n"
              "}\n"
              "vec4 diff = vec4(ambientColor, 1.0);\n";
```

ajtribick (dismissed):

ajtribick (approved):

PR #2660, opened 2026-08-07
Apply eclipse and ring shadows to atmospheres and clouds

levinli303:


AstroChara (approved):
The build correctly displays darkened atmosphere under eclipse and ring shadow.

2026-08-07_Earth_TallAtmo_RingShadow-Eclipse

The atmosphere shadow boundary at the ring shadow edge might be too sharp, but I'm not sure if this is actually an issue or not.

PR #2664, opened 2026-08-08
Use reference counting for GL buffers

ajtribick:
- Vertex object owns its member buffers
- Use intrusive_ptr as weak references not necessary

### DIFF
--- a/src/celrender/gl/buffer.h
+++ b/src/celrender/gl/buffer.h
@@ -11,8 +11,14 @@
 
 #pragma once
 
+#include <atomic>
+#include <cstdint>
+
+#include <boost/smart_ptr/intrusive_ptr.hpp>
+
 #include <celengine/glsupport.h>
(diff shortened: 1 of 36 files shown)


src/celrender/gl/buffer.h
```diff
@@ -56,56 +64,17 @@ class Buffer
...
-     */
-    Buffer(TargetHint             targetHint,
-           util::array_view<void> data,
-           BufferUsage            usage = BufferUsage::StaticDraw);
-
-    //! Copying is prohibited.
-    Buffer(const Buffer &) = delete;
-
     //! Move constructor.
     Buffer(Buffer &&) noexcept;
```
levinli303: can we drop the move constructor and assignment too?

ajtribick: sounds good to me

levinli303 (approved):

PR #2672, opened 2026-08-14
Use instanced rendering in LargeStarRenderer

ajtribick:
Reduce amount of data sent to GPU

### DIFF
--- a/src/celrender/largestarrenderer.cpp
+++ b/src/celrender/largestarrenderer.cpp
@@ -29,21 +29,18 @@ namespace
 {
 struct Corner
 {
-    signed char   x, y;
-    unsigned char u, v;
+    GLbyte  x, y;
+    GLubyte u, v;
 };
 
-// 4-vertex quad in TL, BL, BR, TR order; drawn as two triangles via
-// the static index buffer.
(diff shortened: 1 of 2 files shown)


src/celrender/largestarrenderer.cpp
```diff
@@ -74,10 +71,10 @@ LargeStarRenderer::render()
     makeCurrent();
 
     m_bo->invalidateData().setData(
```
ajtribick, line 73: (not something I've changed in this pull request)
This call to `setData` looks wrong. The buffer has already been orphaned, why do we then reallocate it again with a call to `glBufferData`? Shouldn't this be `setSubData`?

levinli303: the size of the `m_vertices` can change. calling `setSubData` could overflow the buffer. I guess we can call `setSubData` if the allocated size is greater than or equal to the size of `m_vertices`.

ajtribick: Or just allocate the capacity at startup?

ajtribick: anyway will take a look at this in a follow-up PR

levinli303 (approved):

PR #2673, opened 2026-08-15
Support separate atmosphere scale heights

levinli303:
This PR also
1. changes default `AtmosphereSegmentCount` to 3
2. adds UI for changing `AtmosphereSegmentCount`, `CloudSegmentCount`, and `SeparateRayleighMieScaleHeights` in qt

### DIFF
--- a/src/celengine/atmosphere.h
+++ b/src/celengine/atmosphere.h
@@ -10,6 +10,9 @@
 
 #pragma once
 
+#include <algorithm>
+#include <cmath>
+
 #include <Eigen/Core>
 
 #include <celutil/color.h>
@@ -36,4 +39,21 @@ struct Atmosphere
     Eigen::Vector3f absorptionCoeff{ Eigen::Vector3f::Zero() };
 
     float cloudShadowDepth{ 0.0f };
+
+    bool hasValidScaleHeights() const noexcept
+    {
+        return mieScaleHeight > 0.0f && std::isfinite(mieScaleHeight) &&
+               rayleighScaleHeight > 0.0f && std::isfinite(rayleighScaleHeight);
+    }
+
+    float getLegacyScaleHeight() const noexcept
+    {
+        return hasValidScaleHeights() ? std::max(mieScaleHeight, rayleighScaleHeight) : mieScaleHeight;
+    }
+
+    float getLegacyMieCoeff() const noexcept
+    {
+        float legacyScaleHeight = getLegacyScaleHeight();
+        return hasValidScaleHeights() ? mieCoeff * mieScaleHeight / legacyScaleHeight : mieCoeff;
+    }
 };
--- a/src/celengine/atmosphere.h
+++ b/src/celengine/atmosphere.h
@@ -10,9 +10,6 @@
 
 #pragma once
 
-#include <algorithm>
-#include <cmath>
-
 #include <Eigen/Core>
 
 #include <celutil/color.h>
@@ -40,20 +37,7 @@ struct Atmosphere
 
     float cloudShadowDepth{ 0.0f };
 
-    bool hasValidScaleHeights() const noexcept
-    {
-        return mieScaleHeight > 0.0f && std::isfinite(mieScaleHeight) &&
-               rayleighScaleHeight > 0.0f && std::isfinite(rayleighScaleHeight);
-    }
-
-    float getLegacyScaleHeight() const noexcept
-    {
-        return hasValidScaleHeights() ? std::max(mieScaleHeight, rayleighScaleHeight) : mieScaleHeight;
-    }
-
-    float getLegacyMieCoeff() const noexcept
-    {
-        float legacyScaleHeight = getLegacyScaleHeight();
-        return hasValidScaleHeights() ? mieCoeff * mieScaleHeight / legacyScaleHeight : mieCoeff;
-    }
+    bool hasValidScaleHeights() const noexcept;
+    float getLegacyScaleHeight() const noexcept;
+    float getLegacyMieCoeff() const noexcept;
 };
--- a/src/celengine/shadermanager.cpp
+++ b/src/celengine/shadermanager.cpp
@@ -751,50 +751,48 @@ constexpr std::string_view
 RayleighViewOpticalDepth()
 {
     return R"glsl(
-        if (separateScaleHeights)
-        {
-            float rayleighViewStartColumn;
-            if (rayleighViewColumnValid && viewInward == cachedViewInward)
-                rayleighViewStartColumn = cachedRayleighViewColumn;
-            else
-                rayleighViewStartColumn = chapmanToSpaceWithScaleHeight(viewStartRadius, viewInward ? -viewStartMu : viewStartMu, rayleighH);
-            float rayleighViewEndColumn = chapmanToSpaceWithScaleHeight(viewEndRadius, viewInward ? -viewEndMu : viewEndMu, rayleighH);
-            rayleighSegmentDepth = max(0.0, viewInward ? rayleighViewEndColumn - rayleighViewStartColumn : rayleighViewStartColumn - rayleighViewEndColumn);
-            cachedRayleighViewColumn = rayleighViewEndColumn;
-            rayleighViewColumnValid = true;
-        }
+        float rayleighViewStartColumn;
+        if (rayleighViewColumnValid && viewInward == cachedViewInward)
+            rayleighViewStartColumn = cachedRayleighViewColumn;
+        else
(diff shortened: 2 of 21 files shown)


levinli303:
To be compatible with existing atmosphere definitions, two new fields are added `FallbackMie` and `FallbackMieScaleHeight`. The current logic is that (this can be discussed but we should avoid breaking compatibility with add-ons):

| Fallback fields present | `SeparateRayleighMieScaleHeights` | Result |
|---|---:|---|
| Neither | Either | Legacy path: use `Mie` and `MieScaleHeight`; Rayleigh shares `MieScaleHeight`. |
| Either or both | `false` | Legacy fallback path: use `FallbackMie` and/or valid `FallbackMieScaleHeight` where provided; otherwise use the corresponding `Mie` values. Rayleigh shares the selected Mie height. |
| Either or both | `true` | Separated path: use `Mie`, `MieScaleHeight`, and `RayleighScaleHeight` independently. |
| Either or both, but a physical scale height is invalid | `true` | Use the legacy fallback path to avoid invalid atmosphere calculations. |

So to test with Earth, add or change these lines and enable "Separate Rayleigh and Mie scale heights" in settings:

```
# Correct values
RayleighScaleHeight     8.5
Mie                     0.01
MieScaleHeight          1.2

# The old Mie/MieScaleHeight
FallbackMie             0.0014
FallbackMieScaleHeight  8.5
```

pedro-fixingstuff:
Why not use the highest of either MieScaleHeight/RayleighScaleHeight as fallbacks, instead of adding extra parameters?

Perhaps the scattering coefficients could even be scaled accordingly by the ratio of scale heights, as we've been doing on recent content updates.

levinli303:
> Why not use the highest of either MieScaleHeight/RayleighScaleHeight as fallbacks, instead of adding extra parameters?
> 
> Perhaps the scattering coefficients could even be scaled accordingly by the ratio of scale heights, as we've been doing on recent content updates.

if an existing add-on has defined both `MieScaleHeight` and `RayleighScaleHeight`, previously it will just use `MieScaleHeight`. if we do max(MieScaleHeight, RayleighScaleHeight) it will break if RayleighScaleHeight > MieScaleHeight.

AstroChara:
In my opinion, we would probably want to fall back to whichever of the scale height values that has a larger value.

For example, Earth's RayleighScaleHeight is 8.5 and MieScaleHeight is 1.2; if `SeparateRayleighMieScaleHeights` is disabled, the program should use RayleighScaleHeight's 8.5 for both (as is currently the case). We would probably want to use MieScaleHeight instead in the case that its value is larger.

So something like:
- If no scale height is defined, the atmosphere does not render at all.
- If only one scale height is defined, use it for both Rayleigh and Mie scattering.
- If both scale heights are defined and SeparateRayleighMieScaleHeights is disabled, use the larger scale height, and rescale the coefficients associated with the scattering type with a smaller scale height (as Pedro mentioned).
- If both scale heights are defined and SeparateRayleighMieScaleHeights is enabled, use the appropriate scale height for the corresponding scattering.

> Why not use the highest of either MieScaleHeight/RayleighScaleHeight as fallbacks, instead of adding extra parameters?
> 
> Perhaps the scattering coefficients could even be scaled accordingly by the ratio of scale heights, as we've been doing on recent content updates.

I did hear devs explaining before about how Celestia parsing works, but IMO all it means to me is that we need a better way of parsing: one that first stores parsed values in a "buffer" that stays until everything, including extras, is loaded, and only assigns the appropriate values afterwards (IIRC the current system is not like this, leading to issues like trying to modify Radius and SemiAxes).

pedro-fixingstuff:
Admittedly I haven't fully absorbed both replies, but just want to add that since RayleighScaleHeight wasn't a valid parameter so far, we could assume it was never used as the sole scale height for an atmosphere definition, and thus doesn't need to be used as a placeholder for MieScaleHeight, whereas the inverse behavior (Mie H substitutes for Rayleigh H) can be worked by the parser just fine.

AstroChara:
> Admittedly I haven't fully absorbed both replies, but just want to add that since RayleighScaleHeight wasn't a valid parameter so far, we could assume it was never used as the sole scale height for an atmosphere definition, and thus doesn't need to be used as a placeholder for MieScaleHeight, whereas the opposite behavior (Mie H substitutes for Rayleigh H) can be worked by the parser just fine.

Wait I forgot (just woke up) -- actually, is the parser even concerned with the config? I expect it to show up only inside the code, where both Rayleigh scale height ("H") and Mie H should be present (and currently Rayleigh is just set to 0) so wouldn't my scheme work out fine? If anything is not given, default to 0, and if it's 0, substitute it with the other.

Like...if we have a code in...say, shadermanager.cpp or something that checks if either Rayleigh H or Mie H is 0 and replace it with the other:

if (RayleighH = 0) and (MieH != 0):
    RayleighH == MieH
if (MieH = 0) & (RayleighH != 0):
    MieH == RayleighH

then that should be able to perfectly handle my request, no?

levinli303:
ok `RayleighScaleHeight` is not used in solarsys.ssc I guess probably not used in add-ons either.

levinli303:
I'll implement this

| Valid scale heights | `SeparateRayleighMieScaleHeights` | Mie coefficient used | Mie height used | Rayleigh height used |
|---|---:|---|---|---|
| Both `MieScaleHeight` and `RayleighScaleHeight` | `true` | `Mie` | `MieScaleHeight` | `RayleighScaleHeight` |
| Both `MieScaleHeight` and `RayleighScaleHeight` | `false` | `Mie × MieScaleHeight / max(MieScaleHeight, RayleighScaleHeight)` | `max(MieScaleHeight, RayleighScaleHeight)` | Same as Mie |
| Only `MieScaleHeight` | Either | `Mie` | `MieScaleHeight` | Same as Mie |
| `MieScaleHeight` missing or invalid | Either | Legacy behavior | Existing `MieScaleHeight` value | Same as Mie |

A scale height is valid when it is positive and finite.


To test:

```
# Correct values
RayleighScaleHeight     8.5
Mie                     0.01
MieScaleHeight          1.2
```

AstroChara:
I’ll test the case with only RayleighScaleHeight as well later. As Pedro said it’s not an immediate problem, but I believe it would soon crop up, so might as well try it out.

As expected, no atmosphere shows up.

AstroChara:
Also note that currently Absorption still uses MieScaleHeight. For now I think it should use RayleighScaleHeight.

For more details, its possible use cases are:
- Gas absorption (RayleighH)
- Dust absorption (MieH), although in Mars's case this would be imperfect as it cannot replicate blue sunsets.
- Ozone layer (neither fits, but RayleighH is better)

levinli303:
> Also note that currently Absorption still uses MieScaleHeight. For now I think it should use RayleighScaleHeight.
> 
> For more details, its possible use cases are:
> 
> * Gas absorption (RayleighH)
> * Dust absorption (MieH), although in Mars's case this would be imperfect as it cannot replicate blue sunsets.
> * Ozone layer (neither fits, but RayleighH is better)

Changed to use RayleighScaleHeight now.

src/celengine/atmosphere.h
```diff
@@ -36,4 +39,21 @@ struct Atmosphere
...
 
     float cloudShadowDepth{ 0.0f };
+
+    bool hasValidScaleHeights() const noexcept
+    {
+        return mieScaleHeight > 0.0f && std::isfinite(mieScaleHeight) &&
+               rayleighScaleHeight > 0.0f && std::isfinite(rayleighScaleHeight);
+    }
+
+    float getLegacyScaleHeight() const noexcept
```
ajtribick: Maybe worth moving to a .cpp file to avoid bringing in `<algorithm>` in the .h file?

src/celengine/shadermanager.cpp
```diff
@@ -715,31 +715,92 @@ ScatteringPhaseFunctions(const ShaderProperties& /*unused*/)
...
+            rayleighViewColumnValid = true;
+        }
+)glsl";
+}
+
+constexpr std::string_view
+AtmosphereScatteringAccumulation()
+{
+    return R"glsl(
+            if (separateScaleHeights)
```
ajtribick: do we want to do this conditional in the shader code itself, or would it make sense to have the unified and separate scale heights as their own shaders?

src/celengine/atmosphere.h
levinli303: fixed.

src/celengine/shadermanager.cpp
levinli303: that makes sense, added to `ShaderProperties`.

ajtribick (approved):

PR #2675, opened 2026-08-15
Fix some unnecessary reallocations in orphan buffer pattern

ajtribick:
For now, leaving the instances in atmosphererenderer, cometrenderer, linerenderer as the counts in those cases are less clear.

### DIFF
--- a/src/celengine/pointstarvertexbuffer.cpp
+++ b/src/celengine/pointstarvertexbuffer.cpp
@@ -46,23 +46,21 @@ void PointStarVertexBuffer::startBasicPoints()
 
 void PointStarVertexBuffer::render()
 {
-    if (m_nStars != 0)
-    {
-        makeCurrent();
+    if (m_nStars == 0)
+        return;
 
-        if (m_texture != nullptr)
-            m_texture->bind();
+    makeCurrent();
 
-        m_bo->invalidateData().setData(
-            util::array_view(m_vertices.get(), m_nStars),
-            gl::Buffer::BufferUsage::StreamDraw);
+    if (m_texture != nullptr)
+        m_texture->bind();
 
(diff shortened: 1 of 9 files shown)


src/celengine/pointstarvertexbuffer.cpp
```diff
@@ -46,23 +46,21 @@ void PointStarVertexBuffer::startBasicPoints()
...
+    if (m_texture != nullptr)
+        m_texture->bind();
 
-        if (m_pointSizeFromVertex)
-            m_vo1->draw(m_nStars);
-        else
-            m_vo2->draw(m_nStars);
-        m_nStars = 0;
-    }
+    m_bo->invalidateData().setSubData(0, util::array_view(m_vertices.get(), m_nStars));
```
375gnu, line 57: but `invalidateData()` allocates a new buffer.

375gnu:
I think it's worth to check what GL3/GL4 added to make buffer management better.

src/celengine/pointstarvertexbuffer.cpp
ajtribick: My understanding here is that the `invalidateData` call will likely return the existing buffer storage if it isn't currently being used by the driver, since the size and usage parameters remain the same. (Depends on driver optimisation though). So you end up avoiding the implicit synchronisation of `glBufferSubData` because you are guaranteed to have a buffer that isn't in use.

ajtribick: Basing this on https://wikis.khronos.org/opengl/Buffer_Object_Streaming#Buffer_re-specification

375gnu: yes, you're right.

375gnu (approved):

PR #2676, opened 2026-08-16
New splash and logo.png

TheLostProbe:
New splash on dark background:
image

On light background:
image

logo.png on startup:
image
I added a slight drop shadow to make it more readable against the Sun.


Askaniy (approved):

TheLostProbe:
Ah @Askaniy, would you like to be credited for the splash text? I forgot but I think somewhere there is a file which has credits for who created the original splash, I can edit that for this PR.

Askaniy:
Yes, good idea, thanks

PR #2679, opened 2026-08-18
Fix cloud scattering above atmosphere

### LINKED ISSUE #2640: Atmo scattering on top of clouds above atmosphere
**When the cloud shell is above the atmosphere rendering boundary (set by scale height and AtmosphereExtinctionThreshold), they should be expected to simply block the atmosphere from rendering. However, currently (Build 6812 on Mac), they instead begin to render a "ghost" atmosphere on top of themselves.**

In order to reproduce this bug, I made a simple white test cloud texture with four Alpha values: 0 (completely transparent), 0.25, 5, and 1 (completely opaque), in order to observe the dependence of this ghost atmosphere's visibility on the opacity of the clouds. I then varied its altitude and observed the effects.
[TestAtmosphere.zip](https://github.com/user-attachments/files/30298964/TestAtmosphere.zip)

On Earth, with scale height of 8.5 and AtmosphereExtinctionThreshold of 0.000125, the uppermost layer the program renders is at ~76.39 km.

When the clouds' altitude is 100 km, the clouds appear to behave more or less normally, blocking the lower atmosphere as expected. However, by 120 km, the ghost atmosphere begins to render, faintly visible near terminator, and by 150 km, the ghost atmosphere becomes very bright.

150 km
Image

The atmosphere's "halo" at the edge of the ghost atmosphere forms by the time the cloud altitude is set to 170 km, being broader and dimmer than the real atmosphere's. The halo contracts and brightens with increasing cloud altitude, and by 200-300 km it is slightly taller than the real atmosphere, and is positioned just above it.

170 km
Image

200 km (the top right terminator features transition from full opacity to 0.5 opacity, hence the discontinuity)
Image

By 300 km, the ghost atmosphere appears to have multiple terminators. At 500 km and 1000 km, as the "maximum brightness colour" extends taller upwards, more and more terminators appear.

300 km
Image

500 km
Image

1000 km
Image

As long as the extinction threshold is constant, these behaviour scale linearly with scale height. However, they do not appear to scale linearly with the atmosphere render boundary.

seanwessmith:
Fixes #2640.

Clouds above the finite atmosphere boundary can produce an extra atmospheric band when viewed from outside the cloud layer. Disable cloud scattering in that case, using `Renderer::getAtmosphereShellHeight(*atmosphere)` for both the guard and the shader radius. This shares #2673's larger-valid-scale-height calculation and its legacy fallback with the atmosphere pass.

Preserve scattering when the camera is inside the cloud layer: atmosphere can lie in front of the cloud fragments. Reuse the existing inside-cloud decision that selects the cloud sphere's interior faces. Clouds below the atmosphere boundary keep their existing scattering and separate-height shader selection.

Validation on macOS / Apple M4 / OpenGL 4.1:

- Baseline `c4eba10a` and the patched core built successfully; 62/62 CTest tests passed on each.
- Real-renderer framebuffer captures using a minimal Earth-scale scene and #2640's original cloud-alpha texture. Outside matrix: Rayleigh-dominant, Mie-dominant and legacy definitions; separate integration on/off; below/at/1 km above the boundary and 150/300/1000 km clouds.
- All six below-boundary image pairs are pixel-identical. High-cloud ghost bands disappear; at/just-above-boundary differences are at most one 8-bit channel level.
- Additional camera-altitude 20/100 km checks at eight longitudes: all 80 camera-below-cloud comparisons match upstream exactly, as do 32 no-cloud controls. The rejected height-only guard changes 30 of those inside-cloud cases.
- 155/155 local automated framebuffer assertions passed, including that rejected-guard negative control. These are local renderer checks, not newly added CI jobs.
- `git diff --check` passed.

The visual harness uses the actual Celestia renderer, a plain surface and the original test cloud texture; this is not a full-content GUI/platform validation. Existing inside-cloud rendering is preserved, rather than a broader rewrite of scattering integration.


AstroChara:
Considering #2673, which allows Rayleigh and Mie scattering to have different scale heights, I believe the atmosphere height is now defined by whichever value is larger between RayleighH and MieH.

I wonder if it'd make sense to define an "AtmoH" variable, since I recall it is used in a few places.

seanwessmith:
Yes—#2673 already provides the shared helper through `Renderer::getAtmosphereShellHeight(const Atmosphere&)`. It uses the larger Rayleigh/Mie scale height when both are positive and finite, multiplied by `-log(AtmosphereExtinctionThreshold)`, with a Mie fallback for legacy definitions. The shell calculation also applies when separate-height integration is disabled.

Updated this PR to use that helper for the cloud guard and shader radius, retaining the separate-height shader setting. I also limited suppression to viewing a high cloud layer from outside it: the existing inside-cloud decision now preserves atmospheric contributions when the camera is below the clouds.

Verified against current upstream with actual OpenGL captures using #2640's cloud texture. The outside ghost bands disappear, all six below-boundary comparisons are pixel-identical, and all 80 camera-below-cloud comparisons match upstream. The rejected height-only guard changes 30 of those inside-cloud cases. The local framebuffer suite passes 155 assertions, and both core builds pass 62 CTest tests. Updated the description with the scope and platform limitations.

Screenshot 2026-09-12 at 2 25 40 PM
Screenshot 2026-09-12 at 2 25 16 PM

PR #2686, opened 2026-08-28
Experimental C++20

ajtribick:
* Replace `array_view` with `std::span`
* Remove deprecated `u8path`


ajtribick:
~~Go to location seems to be broken just by switching to C++20 mode.~~

Fixed in #2695

PR #2688, opened 2026-08-30
[win32] Update SIMBAD link to https

ajtribick:


pedro-fixingstuff:
Is `simbad.cds.unistra.fr` the new default domain, even if the old one still works?

ajtribick:
Ok let's use that one, then

levinli303 (approved):

PR #2698, opened 2026-09-09
Remove Windows front-end

### LINKED ISSUE #1199: Proposal: remove the Win32 frontend in favour of Qt
I don't think it makes sense given the limited amount of active developers and time to have a separate Win32 UI when we already have a full-featured cross-platform UI using Qt. I therefore propose we remove the Win32 frontend.

If there are (useful) features in the Win32 frontend that are not present in Qt, we should migrate them.

Thoughts? Objections?

ajtribick:
Fixes #1199

In this age of HDR displays and cross-platform support, it does not seem worth maintaining this any more.


INSTALL.md
```diff
@@ -372,7 +376,6 @@ List of supported parameters (passed as `-DPARAMETER=VALUE`):
 | ENABLE_NLS           | bool | ON        | Enable interface translation
 | ENABLE_QT6           | bool | OFF       | Build Qt6 frontend
 | ENABLE_SDL           | bool | OFF       | Build SDL frontend
-| ENABLE_WIN           | bool | \*\*\*OFF | Build Windows native frontend
```
375gnu, line 375: line `\*\*\* Ignored on Unix-like systems.` sh'ld be removed as well

levinli303 (approved):

